PR #2, opened 2016-03-03
Add button to delete torrents

dcposch:
image

### DIFF
--- a/main/index.js
+++ b/main/index.js
@@ -29,7 +29,8 @@ global.WEBTORRENT_ANNOUNCE = createTorrent.announceList
   })
 
 var state = global.state = {
-  server: null,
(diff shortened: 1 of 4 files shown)


main/index.js
```diff
@@ -243,6 +254,16 @@ function openPlayer (torrent) {
   })
 }
 
+function deleteTorrent (torrent) {
+  console.log('Deleting %o', torrent)
```
feross: use `debug()` instead of `console.log()`

main/index.js
```diff
@@ -243,6 +254,16 @@ function openPlayer (torrent) {
...
+  torrent.isDeleting = true
```
feross: i'm guess this isn't used anymore?

feross: removed it

PR #4, opened 2016-03-04
separation of concerns

ngoldman:
Hopefully this is not counterproductive/unhelpful -- I went through everything on the main process side and tried to separate things out into smaller files to reduce mental overhead for myself and future contributors. Organizing helps me understand things :)

Things I did:

- renamed `main` to `renderer` to match electron process names.
- moved everything related to the main process into `main`.
- broke `index.js` into:
  - `config.js` - meant to house shared constants and such. I put `startTime` in there too for now although it's kind of a weird outlier
  - `index.js` - first thing to be required, starts up app, sets menu, creates window, sets ups IPC listeners.
  - `ipc.js` - exports an `init` function that handles all IPC listener binding
  - `menu.js` - exports `appMenu` (main menu built from template) and `onToggleFullScreen` (needed by `windows.js`)
  - `windows.js` - exports `main` (main window instance or null) and `create`
    - manages creation and state of main window, with room for adding other windows

Communicating between the different modules above is not great since a lot of the functions originally needed access to variables outside their own scope, and accessing variables from cached required modules is not a good idea. I'd like to add an event emitter for communication between files in the main process.


ngoldman:
I should mention I tested locally, everything looks to be working exactly as before. The PR is bigger than it looks, mostly just rearranging things.

feross:
This looks really good! I agree that separating this stuff into separate files makes it easier to understand :+1:

Also, I do the same thing when I'm trying to understand code. Refactoring requires you to understand code on a deep level. It's kinda like in college, how reading a practice exam helps you prepare for a test a lot less than actually taking the practice exam, or even just copying out the solutions :)

PR #7, opened 2016-03-05
Fix IPC calls after PR #4

feross:
No biggie, just FYI @ngoldman


ngoldman:
ah woops, my bad, thanks for caching it

PR #9, opened 2016-03-05
try hyperx

ngoldman:
This is a test branch to show what working with https://github.com/substack/hyperx would like.

Also under discussion is using https://github.com/Raynos/main-loop to optimize virtual dom updates.

### DIFF
--- a/renderer/views/torrent-list.js
+++ b/renderer/views/torrent-list.js
@@ -29,15 +29,14 @@ function renderTorrent (state, dispatch, torrent) {
   var elements = [
     renderTorrentMetadata(torrent),
     hx`
-      <i
-        className="icon delete"
+      <i.icon.delete
         onclick=${() => dispatch('deleteTorrent', torrent)}>
         close
       </i>
     `,
     hx`
-      <i
-        className="${!torrent.ready ? 'disabled btn icon play' : 'btn icon play'}"
+      <i.btn.icon.play
+        className="${!torrent.ready ? 'disabled' : ''}"
         onclick=${() => dispatch('openPlayer', torrent)}>
         play_arrow
       </i>
@@ -46,8 +45,8 @@ function renderTorrent (state, dispatch, torrent) {
 
   if (state.view.chromecast) {
     elements.push(hx`
-      <i
-        className="${!torrent.ready ? 'disabled btn icon chromecast' : 'btn icon chromecast'}"
+      <i.btn.icon.chromecast
+        className="${!torrent.ready ? 'disabled' : ''}"
         onclick=${() => dispatch('openChromecast', torrent)}>
         cast
       </i>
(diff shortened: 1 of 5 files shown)


ngoldman:
rebasing...

renderer/views/torrent-list.js
```diff
@@ -19,51 +21,75 @@ function renderTorrent (state, dispatch, torrent) {
...
-      className: !torrent.ready ? 'disabled' : '',
-      onclick: () => dispatch('openAirplay', torrent)
-    }, 'airplay'))
+    elements.push(hx`
+      <i
+        className="${!torrent.ready ? 'disabled btn.icon.airplay' : 'btn.icon.airplay'}"
+        onclick=${() => dispatch('openAirplay', torrent)}>
+        airplay
+      </i>
+    `)
```
ngoldman: Not super happy with how this looks compared to plain `h` but it seemed liked the best way to translate the items in the `elements` array into hyperx.

ngoldman: actually this is wrong, one more rebase on its way

feross: is there no way to include some classes in the element name, like a css selector?

`i.btn.icon.airplay`

ngoldman:
Okay pulled in changes from your last two PRs, tested locally, everything looks good to me. Player changes look nice :+1:

renderer/views/torrent-list.js
ngoldman: @feross It's imitating html like so:

```
<element attr="" attr2="${}" fn=${}></element>
```

As far as I know that's as far as it goes.

ngoldman: Probably the smart thing to do is just pull the ternary out into a function.

feross: Actually, looks like this would work!

```js
hx`<div.hi>`
// [ 'div.hi', {}, [] ]
```

ngoldman: whoa, weird!

renderer/views/torrent-list.js
```diff
@@ -19,51 +21,74 @@ function renderTorrent (state, dispatch, torrent) {
...
   ]
+
   if (state.view.chromecast) {
-    elements.push(h('i.btn.icon.chromecast', {
-      className: !torrent.ready ? 'disabled' : '',
-      onclick: () => dispatch('openChromecast', torrent)
-    }, 'cast'))
+    elements.push(hx`
+      <i.btn.icon.chromecast
+        className="${!torrent.ready ? 'disabled' : ''}"
```
ngoldman: this notation works and is a little less of an eyesore

feross: nice!

feross:
LGTM. Let's merge and use it for a while and see how we like it :)

PR #13, opened 2016-03-05
Lock the electron version

feross:


feross:
For context, this PR:

- locks the version of electron (so we can carefully test and upgrade)
- changes the build script to always use the version of electron specified in `package.json` so we don't need to specify it in multiple places like most Electron based projects do. This is done with the `$npm_package_xxx` environment variable

ngoldman:
Cool, that method might help solve https://github.com/moose-team/friends/issues/154 :+1:

feross:
nice!

PR #14, opened 2016-03-05
Use class instead of className; use standard html

feross:
As of this PR (substack/hyperx#22) to `hyperx`, html attributes are automatically converted to properties for the few cases where they’re different: `class`, `for`, and `http-equiv`.

So, let's change `<div className="">` to `<div class="">`.

Also, I changed all the fancy `<div.my-class>` to `<div class="my-class">` so it's just normal html which is easier to understand, I think. It's less succinct, but it's also clearer and matches the rest of the other attributes that will be in the tag.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "create-torrent": "^3.22.1",
     "debug": "^2.2.0",
     "drag-drop": "^2.3.1",
-    "hyperx": "^1.4.0",
+    "hyperx": "^2.0.0",
     "network-address": "^1.1.0",
     "pretty-bytes": "^3.0.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -16,7 +16,7 @@
     "create-torrent": "^3.22.1",
     "debug": "^2.2.0",
     "drag-drop": "^2.3.1",
-    "hyperx": "^1.4.0",
+    "hyperx": "feross/hyperx",
```
feross: This is just temporary

ngoldman:
LGTM

dcposch:
Sure. No strong opinion.

PR #16, opened 2016-03-05
Cast buttons and fullscreen in player controls

dcposch:
image

### DIFF
--- a/renderer/views/player.js
+++ b/renderer/views/player.js
@@ -53,9 +53,10 @@ function Player (state, dispatch) {
 function renderPlayerControls (state, dispatch) {
   var positionPercent = 100 * state.video.currentTime / state.video.duration
   var playbackCursorStyle = { left: 'calc(' + positionPercent + '% - 4px)' }
+  var torrent = state.view.torrentPlaying._torrent
 
-  return hx`
-    <div class="player-controls">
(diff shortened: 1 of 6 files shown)


renderer/views/player.js
```diff
@@ -63,11 +64,39 @@ function renderPlayerControls (state, dispatch) {
...
         <div class="playback-cursor" style=${playbackCursorStyle}></div>
       </div>
-      <i class="icon play-pause" onclick=${() => dispatch('playPause')}>
-        ${state.video.isPaused ? 'play_arrow' : 'pause'}
+    `
+  ]
+  if (state.view.devices.chromecast) {
+    elements.push(hx`
+      <i.icon.chromecast
+        class="${!torrent.ready ? 'disabled' : ''}"
```
feross: This line shouldn't be necessary anymore, since the user can't get to the player before the torrent is ready anymore.

This was only necessary when the chromecast/airplay buttons were visible in the torrent list.

feross:
LGTM, just one comment inline

renderer/views/player.js
dcposch: word. fixed

ngoldman:
:tada: woo

PR #18, opened 2016-03-05
autoresize window on play

### LINKED ISSUE #17: hyperx PR broke automatic window resize to fit video
Not sure why this happened, but I bisected and found that that's the root cause.

ngoldman:
function references in hyperx need to look like:

```
onloadedmetadata=${onLoadedMetadata}
```

and not:

```
onloadedmetadata="${onLoadedMetadata}"
```

also removed some unnecessary `torrent.ready` checks in the player.

> fixes #17

### DIFF
--- a/renderer/index.js
+++ b/renderer/index.js
@@ -85,7 +85,6 @@ function init () {
   })
 
   document.addEventListener('keydown', function (e) {
-    console.log('keydown ' + e.which)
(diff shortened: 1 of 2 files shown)


ngoldman:
I think this is trivial enough for a merge since no one else is awake :)

feross:
Nice fix! I wonder why there's a distinction.

PR #25, opened 2016-03-05
Start in correct full screen state

feross:
If the app is already fullscreen when it is loaded, set
state.view.isFullScreen to the correct state.


ngoldman:
lgtm

PR #28, opened 2016-03-05
Batch virtual-dom changes (fix #27)

### LINKED ISSUE #27: Batch virtual-dom changes
We don't want to calculate the virtual DOM multiple times in the same animation tick. That's wasted CPU cycles.

That said, it's hard to guarantee that `update()` won't be called multiple times in one tick. So, `update()` needs to get smarter.

feross:
From `main-loop`:

> main-loop is an optimization module for a virtual DOM system.
Normally you would re-create the virtual tree every time your state
changes. This is not optimum, with main-loop you will only update your
virtual tree at most once per request animation frame.

> main-loop basically gives you batching of your virtual DOM changes,
which means if you change your model multiple times it will be rendered
once asynchronously on the next request animation frame.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "debug": "^2.2.0",
     "drag-drop": "^2.3.1",
     "hyperx": "^2.0.0",
+    "main-loop": "^3.2.0",
     "network-address": "^1.1.0",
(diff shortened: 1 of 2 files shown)


ngoldman:
Woo, I was wanting to try that :+1:

PR #30, opened 2016-03-06
Disable view menu items when window is hidden

feross:


feross:
When you minimize or close the main window, the window-related menu items become disabled.

screen shot 2016-03-05 at 4 46 23 pm

feross:
Self-merging. Review still welcome :)

ngoldman:
lgtm :+1:

PR #31, opened 2016-03-06
Add alternate fullscreen shortcut ⌘+Shift+F (fix #26)

### LINKED ISSUE #26: Add alternate fullscreen shortcut ⌘+Shift+F
⌘+Shift+F is an alternate to ⌘+Control+F to fullscreen some apps on Mac. Chrome supports it.

The standard on Mac is ⌘+Control+F. That's already supported and should continue to show up as the shortcut in the View menu.

screen shot 2016-03-05 at 3 20 38 pm

feross:
Uses https://npmjs.com/package/electron-localshortcut to workaround a
bug in Electron (https://github.com/atom/electron/issues/1334).

We can remove `electron-localshortcut` once that bug is fixed.

### DIFF
--- a/main/index.js
+++ b/main/index.js
@@ -4,6 +4,7 @@ var electron = require('electron')
 var ipc = require('./ipc')
 var menu = require('./menu')
 var windows = require('./windows')
+var shortcuts = require('./shortcuts')
 
 var app = electron.app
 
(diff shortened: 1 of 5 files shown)


feross:
Self-merging, review still welcome.

PR #33, opened 2016-03-06
Fullscreen fixes

feross:
This change does the following:

- Prevent the window from resizing when in fullscreen because that just behaves weirdly.
- Show the header on Linux/Windows when fullscreened, but not in the
player. Users might fullscreen the app when they’re not playing a video.
- Always show the header on OS X (even when fullscreened) since that’s
how the user will exit the video. We can work on adding auto-hiding to
it later.


ngoldman:
sounds good to me :+1:

PR #34, opened 2016-03-06
improve progress percent & bytes display

ngoldman:
Tiny adjustment for progress bytes.

Ensures that at 0 or 100%, progress will not show duplicate values. So no `0 B / 0 B` or `129.4 MB / 129.4 MB`. Instead it looks like this:

screen shot 2016-03-05 at 6 43 16 pm

### DIFF
--- a/renderer/views/torrent-list.js
+++ b/renderer/views/torrent-list.js
@@ -45,23 +45,18 @@ function renderTorrent (torrent, dispatch) {
 
 // Renders the torrent name and download progress
 function renderTorrentMetadata (torrent) {
-  var progressPercent = 0
-  var progressBytes = 0
+  var progress = Math.floor(100 * (torrent.progress || 0))
+  var downloaded = prettyBytes(torrent.downloaded || 0)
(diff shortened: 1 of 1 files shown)


renderer/views/torrent-list.js
```diff
@@ -45,23 +45,18 @@ function renderTorrent (torrent, dispatch) {
 
 // Renders the torrent name and download progress
 function renderTorrentMetadata (torrent) {
-  var progressPercent = 0
-  var progressBytes = 0
+  var progress = Math.floor(100 * (torrent.progress || 0))
+  var downloaded = prettyBytes(torrent.downloaded || 0)
+  var total = prettyBytes(torrent.length || 0)
```
feross: `torrent.progress` and `torrent.downloaded` will never be `undefined`, so no need for `|| 0`.
The length depends on the metadata being received from the network, but I'll initialize it to 0 in webtorrent and remove it from here later.

feross:
Very nice work. This was on my list of changes to make :+1:

PR #35, opened 2016-03-06
Hide cursor after 2s in fullscreen

ngoldman:


ngoldman:
Missed commit from `dc/state` branch. Rebasing.

renderer/index.js
```diff
@@ -360,7 +364,10 @@ function setDimensions (dimensions) {
   var width = Math.floor(dimensions.width * scaleFactor)
   var height = Math.floor(dimensions.height * scaleFactor)
 
-  height += HEADER_HEIGHT
+  // Video player header only shows in OSX where it's part of the title bar. See app.js
+  if (process.platform === 'darwin') {
+    height += HEADER_HEIGHT
+  }
```
feross: good catch

renderer/views/player.js
```diff
@@ -23,10 +23,16 @@ function Player (state, dispatch) {
     state.video.duration = videoElement.duration
   }
 
+  // When in fullscreen, hide player controls if the mouse stays still for a while
+  //console.log('STATIONARY FOR ' + new Date().getTime() - state.video.mouseStationarySince)
```
feross: We should hide the player controls if the mouse is still, even when not in fullscreen.

YouTube, Quicktime do the same.

ngoldman:
Looks like after rebase it was already in master :stuck_out_tongue_closed_eyes:

feross:
Heh, I'm confused. So @dcposch made these changes? I left some inline comments, in any case.

ngoldman:
I saw a branch that looked like it had a commit that didn't get merged and it looked like good stuff so I thought I'd try to get it in, not realizing it had already been pushed to master.

¯\_(ツ)_/¯

feross:
Okay, got it. Thanks for explaining :+1:

dcposch:
Yeah I did, late last night. I'll make PRs from now on so you can see what I'm up to.

PR #38, opened 2016-03-06
Use single quotes in hyperx

feross:
This is more consistent with the rest of standard style, and if this was JSX it would be enforced by the tool.


ngoldman:
:+1: s'all good

PR #39, opened 2016-03-06
Hide player controls when mouse is still, even when not fullscreened

### LINKED ISSUE #37: Hide player controls when mouse is still, even when not fullscreened
We should hide the player controls if the mouse is still, even when not in fullscreen.

YouTube and QuickTime do the same.

feross:


feross:
Will send a separate PR for hiding the header bar too.

ngoldman:
beat me to changing fullscreenVideoMouseMoved to videoMouseMoved, I'm working on that locally for hiding header.

feross:
@ngoldman okay, i'll let you do hiding the header bar :)

PR #43, opened 2016-03-06
hide header on video after mouse idle

ngoldman:
For some reason this took me a long time to figure out.

### DIFF
--- a/renderer/views/header.js
+++ b/renderer/views/header.js
@@ -5,21 +5,26 @@ var hyperx = require('hyperx')
 var hx = hyperx(h)
 
 function Header (state, dispatch) {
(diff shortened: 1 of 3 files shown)


renderer/views/header.js
```diff
@@ -41,6 +46,7 @@ function Header (state, dispatch) {
       return hx`
         <i
           class='icon add'
+          title='add torrent'
```
ngoldman: free tooltips

feross: nice, good old-fashioned too. maybe we can also add a `<marquee>`?

feross:
LGTM

PR #45, opened 2016-03-06
header disappear when mouse not over

feross:
This makes the header disappear right away when the mouse leaves the window, instead of waiting for the inactive mouse timeout.

This way the header matches what the controls do.

### DIFF
--- a/renderer/views/app.js
+++ b/renderer/views/app.js
@@ -11,8 +11,13 @@ var TorrentList = require('./torrent-list')
 var isOSX = process.platform === 'darwin'
 
 function App (state, dispatch) {
+  var cls = []
+  if (state.url === '/player') {
+    cls.push('player')
+  }
+
   return hx`
-    <div class='app'>
+    <div class='app ${cls.join(' ')}'>
       ${getHeader()}
       <div class='content'>${getView()}</div>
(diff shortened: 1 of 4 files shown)


feross:
don't merge, still working on this

feross:
okay, this is good to go

feross:
Self merging, but @ngoldman and @dcposch can you please take a look?

renderer/views/app.js
```diff
@@ -11,8 +11,14 @@ var TorrentList = require('./torrent-list')
...
 function App (state, dispatch) {
+  var cls = [
+    process.platform,
+    state.isFullScreen ? 'fullscreen' : 'not-fullscreen',
+    state.url === '/player' ? 'view-player' : ''
+  ]
+
   return hx`
-    <div class='app'>
+    <div class='app ${cls.join(' ')}'>
```
ngoldman: I was thinking in the long run leaning on CSS classes more to show/hide will be a lot easier. :+1:

ngoldman:
lgtm

renderer/views/app.js
feross: Agreed :+1:

PR #46, opened 2016-03-06
add persistent app config

### LINKED ISSUE #24: persistent state
Currently the app loses the torrent list every time the main window closes or is reloaded, and when the app is restarted. Ideally state is stored on disk somewhere so that sessions are just resuming what's already been in progress beforehand.

Looking for good modules (or just vanilla solutions, like a state.json?) to solve this issue. Currently avoiding anything that requires native dependencies or grows the dependency graph in a big way.

- https://github.com/LinusU/node-application-config
    - This module is used by ghauth and has worked well for me in the past. Might be too small for handling things like the torrent list and video playback state once more things need to be tracked, but works for basics and would still be useful for storing application preferences in the future (like default download directory).

ngoldman:
Using `application-config` for now. So far working great. Storing magnet URIs of torrents in config instead of `.torrent` file paths since those are often deleted.

Played around with some different ways to capture and save torrent info until I realized I could just save info from `state.client.map` with the props I wanted on every client `torrent` event. That way no deduping is necessary. The WebTorrent client always tells the truth.

> resolves #24

### DIFF
--- a/renderer/index.js
+++ b/renderer/index.js
@@ -13,6 +13,7 @@ var state = require('./state')
 var throttle = require('throttleit')
 var torrentPoster = require('./lib/torrent-poster')
 var WebTorrent = require('webtorrent')
+var cfg = require('application-config')('WebTorrent')
 
 var createElement = require('virtual-dom/create-element')
 var diff = require('virtual-dom/diff')
@@ -35,10 +36,7 @@ function init () {
   state.client = new WebTorrent()
   state.client.on('warning', onWarning)
   state.client.on('error', onError)
-
-  state.saved.torrents.forEach(function (torrent) {
-    state.client.add(torrent.torrentFile)
-  })
+  state.client.on('torrent', saveTorrents)
 
   // For easy debugging in Developer Tools
   global.state = state
@@ -53,6 +51,8 @@ function init () {
   updateThrottled = throttle(update, 1000)
 
   dragDrop('body', onFiles)
+
+  restoreSession()
 
   chromecasts.on('update', function (player) {
     state.devices.chromecast = player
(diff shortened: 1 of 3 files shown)


ngoldman:
One important thing to note, this does NOT implement deleting from the config.

ngoldman:
Okay now deleting torrents persists.

ngoldman:
Should be noted, `torrent.destroy` doesn't remove files from `/tmp/webtorrent`. That's probably the desired behavior once we get a default & custom download folder into prefs.

renderer/index.js
```diff
@@ -251,6 +251,35 @@ function addTorrentEvents (torrent) {
...
+    return {
+      name: torrent.name,
+      magnetURI: torrent.magnetURI,
+      path: torrent.path
+    }
+  })
+
+  cfg.write({
+    torrents: state.saved.torrents
+  }, function (err) {
```
feross: Why not `cfg.write(state.saved)` to save the whole thing?

renderer/index.js
```diff
@@ -286,7 +315,9 @@ function openPlayer (torrent) {
 }
 
 function deleteTorrent (torrent) {
-  torrent.destroy(update)
+  torrent.destroy(function () {
+    saveTorrents() // updates after writing to config
```
feross: Good call. We can add some indication to the UI that it's being deleted later, so the button feels responsive.

renderer/index.js
```diff
@@ -251,6 +251,35 @@ function addTorrentEvents (torrent) {
   }
 }
 
+function restoreSession () {
+  cfg.read(function (err, data) {
+    if (err) console.error(err)
+    state.saved = data
+    if (!state.saved.torrents) state.saved.torrents = []
```
feross: Why is this necessary?

feross:
THIS IS OFFICIALLY A REAL APP NOW. :rocket:

renderer/index.js
ngoldman: Thought that might lead to overwrites in the future, but that would be fine for now.

renderer/index.js
ngoldman: On initial read of config it's an empty object, so state.saved.torrents.forEach would throw an error.

feross: If we detect it's an empty object, can't we just not do `state.saved = data`, since `state.saved.torrents` is already set to  an initial value in state.js?

ngoldman: Truth.

PR #48, opened 2016-03-06
perfectly vertically center OS menu buttons (OS X)

feross:


feross:
Before:

screen shot 2016-03-06 at 12 42 26 am

After (header is 1px shorter):

screen shot 2016-03-06 at 12 42 53 am

ngoldman:
lgtm but needs rebase after that last merge

ngoldman:
I think I made it `38px` earlier because because the line between header and list looked better as 1px than 2px.

feross:
Rebased and merged :)

PR #49, opened 2016-03-06
Unsquash the UI when window is small

### LINKED ISSUE #44: UI is squashed when window is small
screen shot 2016-03-05 at 9 08 11 pm

I can fix this one in a little bit. Issue so I don't forget.

feross:
Fix #44


feross:
Before:



After:

screen shot 2016-03-06 at 1 02 15 am

ngoldman:
:+1:

PR #56, opened 2016-03-07
Save and load torrents, document index.js

dcposch:
### This PR
* Clean up `index.js`
* Separate `state.saved.torrents` from the list of currently active torrents (`state.client.torrents`)
  This prepares us to support torrents which are not currently downloading -- such as the default videos that will show when Webtorrent.app is first installed. Such torrents are in `state.saved.torrents` but not in the active torrents list.

## Next PR
* Make the Blender Foundation videos show up by default
* Add a small Download button next to the big Play button
  This starts / pauses the torrent download. While downloading the Play button will pulsate to show activity. Once the torrent is fully downloaded it will be solid green, to show that we're done and are now seeding. The button will act as a toggle--if you turn it off, it turns grey to indicate that the torrent is not currently torrenting. Pressing play automatically starts the torrent (if it is not already fully DL'd) and starts playing / streaming.

### DIFF
--- a/renderer/index.js
+++ b/renderer/index.js
@@ -32,15 +32,23 @@ global.WEBTORRENT_ANNOUNCE = createTorrent.announceList
 
 var vdomLoop, updateThrottled
 
+/**
+ * Called once when the application loads. (Not once per window.)
+ * Connects to the torrent networks, sets up the UI and OS integrations like
+ * the dock icon and drag+drop.
+ */
 function init () {
+  // Connect to the WebTorrent and BitTorrent networks
+  // WebTorrent.app is a hybrid client, as explained here: https://webtorrent.io/faq
   state.client = new WebTorrent()
   state.client.on('warning', onWarning)
   state.client.on('error', onError)
-  state.client.on('torrent', saveTorrents)
+  state.client.on('torrent', saveState)
 
-  // For easy debugging in Developer Tools
-  global.state = state
-
+  // The UI is built with virtual-dom, a minimalist library extracted from React
+  // The concepts--one way data flow, a pure function that renders state to a
+  // virtual DOM tree, and a diff that applies changes in the vdom to the real
+  // DOM, are all the same. Learn more: https://facebook.github.io/react/
   vdomLoop = mainLoop(state, render, {
     create: createElement,
     diff: diff,
@@ -48,12 +56,22 @@ function init () {
   })
   document.body.appendChild(vdomLoop.target)
 
+  // Calling update() updates the UI given the current state
+  // For performance, we don't want to do this too often. Hence, events that
+  // affect the UI can be v frequent--like torrent progress updates--call
+  // updateThrottled() instead, which calls update() at most once a second.
   updateThrottled = throttle(update, 1000)
+  setInterval(updateThrottled, 1000)
 
-  dragDrop('body', onFiles)
+  // All state lives in state.js. `state.saved` is read from and written to a
+  // file. All other state is ephemeral. Here we'll load state.saved:
+  loadState()
 
-  restoreSession()
+  // For easy debugging in Developer Tools
+  global.state = state
 
+  // OS integrations:
+  // ...Chromecast and Airplay
   chromecasts.on('update', function (player) {
     state.devices.chromecast = player
     update()
@@ -63,10 +81,15 @@ function init () {
     state.devices.airplay = player
   }).start()
 
+  // ...drag and drop a torrent or video file to play or seed
+  dragDrop('body', onFiles)
+
+  // ...same thing if you paste a torrent
   document.addEventListener('paste', function () {
     electron.ipcRenderer.send('addTorrentFromPaste')
   })
 
+  // ...keyboard shortcuts
   document.addEventListener('keydown', function (e) {
     if (e.which === 27) { /* ESC means either exit fullscreen or go back */
       if (state.isFullScreen) {
@@ -77,6 +100,7 @@ function init () {
     }
   })
 
+  // ...focus and blur. Needed to show correct dock icon text ("badge") in OSX
   window.addEventListener('focus', function () {
     state.isFocused = true
     if (state.dock.badge > 0) electron.ipcRenderer.send('setBadge', '')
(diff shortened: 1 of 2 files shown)


renderer/index.js
```diff
@@ -48,12 +56,22 @@ function init () {
   })
   document.body.appendChild(vdomLoop.target)
 
+  // Calling update() updates the UI given the current state
+  // For performance, we don't want to do this too often. Hence, events that
+  // affect the UI can be v frequent--like torrent progress updates--call
+  // updateThrottled() instead, which calls update() at most once a second.
   updateThrottled = throttle(update, 1000)
+  setInterval(updateThrottled, 1000)
```
ngoldman: aren't actions from dispatch already calling updateThrottled enough? why forced an update every second?

dcposch: Don't know--the call to `setInterval` was already there, but it was further down outside of the init function.

Maybe @feross knows

renderer/index.js
```diff
@@ -192,6 +201,40 @@ electron.ipcRenderer.on('addFakeDevice', function (e, device) {
...
+    console.log("loaded state: " + JSON.stringify(data, null, 4) /* pretty print */
+    state.saved = data
+    if (!state.saved.torrents) state.saved.torrents = []
+    state.saved.torrents.forEach(startTorrenting)
+  })
+}
+
+// Write state.saved to the JSON state file
+function saveState () {
+  cfg.write(state.save, function (err) {
```
ngoldman: typo from last comment is still here. `state.save` should be `state.saved`, right?

ngoldman:
lgtm other than two comments above. thanks for documenting the index!

renderer/index.js
dcposch: Oh--I know. Its so that `update` gets called at least once a second, to keep the cursor in sync while a video is playing

dcposch: This is actually a derp -- limiting `updateThrottled` to *at most once a second*, then calling it exactly once a second, is equivalent to removing `updateThrottled` entirely everywhere it's called and just saying

    setInterval(update, 1000)

@feross does that look right?

ngoldman: Haha yeah that doesn't sound great. I guess `updateThrottled` does ensure it's never called more than once a second since multiple dispatches can come in within a second.

renderer/index.js
```diff
@@ -211,12 +254,25 @@ function isNotTorrentFile (file) {
   return !isTorrentFile(file)
 }
 
-function addTorrent (torrentId) {
-  if (!torrentId) torrentId = 'magnet:?xt=urn:btih:6a9759bffd5c0af65319979fb7832189f4f3c35d&dn=sintel.mp4'
-  var torrent = state.client.add(torrentId)
+// Adds a torrent to the list, starts downloading/seeding
+function addTorrent (torrentOrMagnetURI) {
+  if (!torrentOrMagnetURI) torrentOrMagnetURI = 'magnet:?xt=urn:btih:6a9759bffd5c0af65319979fb7832189f4f3c35d&dn=sintel.mp4'
+  var torrent = startTorrenting(torrentOrMagnetURI)
```
ngoldman: Since WebTorrent supports a bunch of different things we can too!

```
magnet uri (string)
info hash (hex string)
http/https url to a torrent file (string)
filesystem path to a torrent file (string)
```

renderer/index.js
```diff
@@ -342,25 +367,29 @@ function openAirplay (torrent) {
...
+// Set window dimensions to match video dimensions or fill the screen
 function setDimensions (dimensions) {
-  // TODO: eliminate blocking remote call
-  state.mainWindowBounds = electron.remote.getCurrentWindow().getBounds()
+  state.mainWindowBounds = {
+    x: window.screenX,
+    y: window.screenY,
+    width: window.outerWidth,
+    height: window.outerHeight
+  }
```
ngoldman: :+1:

ngoldman:
Merging since everything looks good and I want to keep building on top of what @dcposch's done.

renderer/index.js
```diff
@@ -192,6 +201,40 @@ electron.ipcRenderer.on('addFakeDevice', function (e, device) {
...
 })
 
+// Load state.saved from the JSON state file
+function loadState () {
+  cfg.read(function (err, data) {
+    if (err) console.error(err)
+    console.log('loaded state: ' + JSON.stringify(data, null, 4)) /* pretty print */
+    state.saved = data
+    if (!state.saved.torrents) state.saved.torrents = []
+    state.saved.torrents.forEach(startTorrenting)
```
ngoldman: just noticed this introduces an error since torrents are currently saved as objects, will fix

renderer/index.js
```diff
@@ -211,12 +254,25 @@ function isNotTorrentFile (file) {
...
 
-function addTorrent (torrentId) {
-  if (!torrentId) torrentId = 'magnet:?xt=urn:btih:6a9759bffd5c0af65319979fb7832189f4f3c35d&dn=sintel.mp4'
-  var torrent = state.client.add(torrentId)
+// Adds a torrent to the list, starts downloading/seeding
+function addTorrent (torrentOrMagnetURI) {
+  if (!torrentOrMagnetURI) torrentOrMagnetURI = 'magnet:?xt=urn:btih:6a9759bffd5c0af65319979fb7832189f4f3c35d&dn=sintel.mp4'
+  var torrent = startTorrenting(torrentOrMagnetURI)
+  state.saved.torrents.add({
+    name: torrent.name,
```
ngoldman: I think this should be `.push` not `.add`

dcposch:
thx for review!

DON'T worry about merging my PRs--i'll merge them when ready by rebasing, to keep the history of the `master` branch pretty

my branch was full of fail when merged. making a follow up commit with my fixes since then.

ngoldman:
Sorry @dcposch! Will hold off on merging your PRs in the future. I fixed a couple things in #58 already.

renderer/index.js
feross: Eventually I want to get rid of the `setInterval` since we don't want to do DOM diffing every 1s. That's super expensive, especially when the app is in the background. But yeah this works for now :)

renderer/index.js
feross: The convention that WebTorrent uses throughout the various repos is `torrentId`. There's just a lot of ways to specify a torrent.

feross: Also, if you paste you can already use any of the string types from that list!

Issue #42, opened 2016-03-06
Need better placeholder when torrent screenshot is not ready yet


feross:
When there's not enough data to show a torrent screenshot, the background is black:

![screen shot 2016-03-05 at 8 11 25 pm](https://cloud.githubusercontent.com/assets/121766/13552220/a4081e0c-e30e-11e5-8561-7fc74860e032.png)

We can do better.

giuscri:
Hash the blob obtained so far, map the hash to a value from a list of light/soft/_solarized_ colors and use that as the background. :3

ngoldman:
@giuscri sounds fun but I think that would prove rather expensive. I tried something simple in #57.

feross:
Fixed in #57

PR #57, opened 2016-03-07
better default torrent bg

ngoldman:
Addresses #42.

Playing around with linear gradients from http://uigradients.com. This one looks pretty good to me so far:

screen shot 2016-03-06 at 9 18 38 pm

Could use improvement but it's better than all black :)

### DIFF
--- a/renderer/index.css
+++ b/renderer/index.css
@@ -223,7 +223,7 @@ body.drag::before {
 
 .torrent {
   height: 120px;
-  background: rgba(0, 0, 0, 0.5);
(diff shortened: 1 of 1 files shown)


ngoldman:
If greater variety is desired we could also do something like

```css
.torrent:nth-child(3n+1) {
   background: linear-gradient(ABC);
}

.torrent:nth-child(3n+2) {
   background: linear-gradient(DEF);
}

.torrent:nth-child(3n+3) {
   background: linear-gradient(XYZ);
}
```

feross:
This is much better for now! Nice.

PR #58, opened 2016-03-07
fix bugs from #56

ngoldman:
Two small fixes to bugs introduced in #56. Sorry I didn't catch them during review.


ngoldman:
Self-merging since `state.saved` is otherwise broken.

Issue #53, opened 2016-03-06
downloads folder


ngoldman:
Right now downloads just go to `/tmp/webtorrent` and get deleted on restart. This app needs a persistent downloads folder!

## 1. default
What is standard the downloads folder for each OS? For OS X it's `~/Downloads`. Might be good to set default to whatever the OS default is.

## 2. custom
We need a way to set a custom downloads folder location. Perhaps this is the beginning of a preferences view.

giuscri:
A third possibility would be to store streamed content in a `.webtorrent` directory.

ngoldman:
Though the video streaming is prominent I still think of this primarily as a BitTorrent client. Making a distinction between content to be streamed and any other torrent seems odd.

Either way I vote for giving the user the option to control where files are saved, and I think opting for OS defaults is a good way to respect the local environment the app happens to be in.

yoshuawuyts:
having control would be nice

ngoldman:
I poked around and it seems like `~/Downloads` might actually be standard across Windows/Linux/Mac OS X. Can anyone else confirm?

yoshuawuyts:
[Chrome help page](https://support.google.com/chrome/answer/95574?hl=en) says so:
```
Windows Vista/ Windows 7/ Windows 8: \Users\<username>\Downloads
Mac: /Users/<username>/Downloads
Linux: home/<username>/Downloads
```

ngoldman:
:+1: thanks @yoshuawuyts!

wsxy162:
For linux OS，do you know about xdg-user-dirs?  "home/<username>/Downloads" is only
 a hardcoded directory location, "XDG_DOWNLOAD_DIR" would be better. And I still want a custom directory feature.

feross:
@wsxy162 This will be fixed soon.

PR #59, opened 2016-03-07
load state first

ngoldman:
This PR does a few things:

- loads state before initializing app
- wraps IPC setup in a function called by `init()` to avoid a race condition where `update()` gets called too soon
- moves resuming of saved torrents into a function
- removes an unnecessary `console.log`

Loading state first allows setting defaults and preferences first (setup for #53).

Side thought:

Wondering if it might be faster to load app with less going on upfront (just header and empty list initialized) then send saved state over IPC from the main process. Startup is pretty slow for me at this stage and I feel like there's not enough going on in the UI to warrant that.

### DIFF
--- a/renderer/index.js
+++ b/renderer/index.js
@@ -9,7 +9,6 @@ var EventEmitter = require('events')
 var mainLoop = require('main-loop')
 var networkAddress = require('network-address')
 var path = require('path')
-var state = require('./state')
 var torrentPoster = require('./lib/torrent-poster')
 var WebTorrent = require('webtorrent')
 var cfg = require('application-config')('WebTorrent')
@@ -19,6 +18,9 @@ var diff = require('virtual-dom/diff')
 var patch = require('virtual-dom/patch')
 
 var App = require('./views/app')
+
+// For easy debugging in Developer Tools
+var state = global.state = require('./state')
 
 // Force use of webtorrent trackers on all torrents
 global.WEBTORRENT_ANNOUNCE = createTorrent.announceList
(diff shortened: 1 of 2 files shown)


ngoldman:
Real downloads working!

ngoldman:
This PR is ready and should fix an issue with storing an empty object when infoHash isn't present on initial torrent object creation. Also downloads folder!!!

renderer/index.js
```diff
@@ -257,10 +276,27 @@ function isNotTorrentFile (file) {
...
+  var exists = state.saved.torrents.find((x) => x.infoHash === torrent.infoHash)
+  if (exists) return window.alert('That torrent is already downloading.')
+
+  // save only if infoHash is available
+  if (torrent.infoHash) {
+    state.saved.torrents.push({
+      infoHash: torrent.infoHash
+    })
+  } else {
+    torrent.on('infoHash', () => saveTorrentData(torrent))
```
feross: Good catch. Remote torrents, like those that come from filesystem paths or http urls will not have the `infoHash` property set immediately :)

renderer/index.js
```diff
@@ -257,10 +276,27 @@ function isNotTorrentFile (file) {
...
   }
-  state.saved.torrents.push({
+
+  saveState()
+}
+
+// add torrent metadata to state once it's available
+function saveTorrentData (torrent) {
+  var ix = state.saved.torrents.findIndex((x) => x.infoHash === torrent.infoHash)
```
feross: Nice, didn't know about `findIndex`!

feross:
LGTM!! Nice work!

renderer/index.js
ngoldman: me neither, saw it in @dcposch's PR last night

PR #61, opened 2016-03-07
Remove updateThrottled

dcposch:
`updateThrottled` uses `setTimeout`/`clearTimeout` to ensure `update` is called at most once a second, but then we call `updateThrottle` exactly once a second anyway!

that's equivalent to calling `update` once a second directly


ngoldman:
That's not quite right... please see comment from earlier https://github.com/feross/webtorrent-app/pull/56#discussion_r55164029

PR #63, opened 2016-03-07
Double-click video to enter fullscreen

ChrisMorrisOrg:


renderer/views/player.js
```diff
@@ -116,6 +117,7 @@ function renderPlayerControls (state, dispatch) {
 
   // Handles a click or drag to scrub (jump to another position in the video)
   function handleScrub (e) {
+    dispatch('videoMouseMoved')
```
dcposch: do we need this? if we're in `handleScrub`, we just had a mousemove event anyway, right?

ChrisMorrisOrg: @dcposch That's what I had expected too, but the mousemove event is attached to .player and .letterbox. When you're scrubbing on .scrub-bar, it doesn't register the mousemove event. Try dragging the scrub bar for more than two seconds, it will hide.

Also, in my tests, adding another mousemove event to the div.scrub-bar itself does not execute when the user is dragging. I believe that might be because it's relative to the element itself, so if you're dragging, the cursor is actually going to stay in the same position relative to the scrub-bar.

ngoldman: I think this is meant to address #64 -- haven't tried to repro yet

ChrisMorrisOrg: @ngoldman Yeah, I think I messed up a bit with the PR. There were meant to be two seperate ones. One for the double-click to fullscreen, and the other to fix the scrub-bar from hiding when you're scrubbing.

ngoldman: I was able to reproduce -- I think once you're dragging it may not be causing the mousemove event to fire.

ngoldman:
lgtm for now, might want to investigate a way to more efficiently capture both drag and mouse move.

feross:
LGTM

ChrisMorrisOrg:
Perhaps we should look at creating a showControls() function that returns true or false. There seem to be three cases where it should return true:
- on seek
- on mouse move
- on streaming via AirPlay/Chromecast

ngoldman:
I recommend a `.hide-controls` css class that fades them in & out with a css animation. We can just add & remove the class at the app level as necessary. Much cheaper than JS checks and function calls methinks.

dcposch:
@ngoldman yeah --- we have that (currently called `.hide`), but the logic for when to add & remove that could use a bit more sophistication. i think this PR is a good step!

i'll try:
* never hiding controls while we are moused over the controls (as opposed to elsewhere on the video)
* never hiding controls while the video is paused

ChrisMorrisOrg:
> never hiding controls while the video is paused
@dcposch Yep! Good idea. If you're implementing a more sophisticated hideControls() method, or something similar, this will come in handy for when we get AirPlay/Chromecast working properly too.

PR #65, opened 2016-03-07
Show Blender Foundation videos by default

dcposch:
* Finishes the saved torrents feature!
* Torrents load and save correctly. Poster is autogenerated only once.
* Torrents can be paused and restarted
* Download button indicates state:
 - White means paused
 - Pulsating green means downloading
 - Solid green means finished downloading, only seeding

byzanz


main/ipc.js
```diff
@@ -64,7 +64,7 @@ function setAspectRatio (aspectRatio, extraSize) {
 // Display string in dock badging area (OS X)
 function setBadge (text) {
   debug('setBadge %s', text)
-  electron.app.dock.setBadge(String(text))
+  if (electron.app.dock) electron.app.dock.setBadge(String(text))
```
feross: Is this actually `undefined` on non OS X platforms?

feross: yep, it is

PR #68, opened 2016-03-07
Fix #67

### LINKED ISSUE #67: Application does not start torrenting anymore
Apparently in commit [`093d810`](https://github.com/feross/webtorrent-app/commit/093d810b90dfc4b61a5b8f5b5a05eb9c58f3d6c6), `WebTorrent.add()` is called using the torrent's infoHash while the API need the magnetURI.

giuscri:


ngoldman:
Please see my response in #67, what's being stored in the application config has changed. Info hashes do work, they just weren't being stored before.

PR #69, opened 2016-03-07
add loading state, ui tweaks

ngoldman:
On the off chance that loading is taking a while, this PR adds a nice loading spinner. Also makes the script call async which seemed to speed up time to first paint for me. For fun I also added a quick fade in css animation to the app after loading is done, and to new torrents. Up to collaborators if it's worth keeping or tweaking or ditching.

screen shot 2016-03-07 at 12 51 26 pm

### DIFF
--- a/renderer/index.js
+++ b/renderer/index.js
@@ -1,5 +1,6 @@
 /* global URL, Blob */
 
+console.time('init')
 var airplay = require('airplay-js')
 var chromecasts = require('chromecasts')()
 var createTorrent = require('create-torrent')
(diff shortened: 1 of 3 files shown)


renderer/index.js
```diff
@@ -213,6 +211,16 @@ function setupIpc () {
...
 
+function detectDevices () {
+  chromecasts.on('update', function (player) {
+    state.devices.chromecast = player
+  })
+
+  airplay.createBrowser().on('deviceOn', function (player) {
+    state.devices.airplay = player
+  }).start()
+}
```
ngoldman: also moved device detection into its own function and removed the `update()` inside the chromecasts callback. I figure we're generally covered on updates.

renderer/index.js
```diff
@@ -121,6 +117,8 @@ function init () {
   window.addEventListener('blur', function () {
     state.isFocused = false
   })
+
+  console.timeEnd('init')
```
ngoldman: handy for figuring out what's taking so long.

PR #70, opened 2016-03-07
Add AUTHORS.md

feross:
Recognize contributors for their work!


feross:
The shell script is based on the one that is used by Node.js.

ngoldman:
lgtm :+1:

PR #71, opened 2016-03-07
Cross-platform home directory detection

feross:
os.homedir() handles cross-platform complexities for us.

### DIFF
--- a/renderer/index.js
+++ b/renderer/index.js
@@ -1,24 +1,27 @@
 /* global URL, Blob */
 
 var airplay = require('airplay-js')
+var cfg = require('application-config')('WebTorrent')
 var chromecasts = require('chromecasts')()
 var createTorrent = require('create-torrent')
 var dragDrop = require('drag-drop')
 var electron = require('electron')
 var EventEmitter = require('events')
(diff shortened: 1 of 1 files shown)


renderer/index.js
```diff
@@ -20,6 +21,8 @@ var patch = require('virtual-dom/patch')
 
 var App = require('./views/app')
 
+var HOME = os.homedir()
```
ngoldman: nice, good to know!

ngoldman:
:sparkles: lgtm :sparkles:

renderer/index.js
```diff
@@ -318,7 +320,7 @@ function saveTorrentData (torrent) {
 function startTorrenting (torrentId) {
   var torrent = state.client.add(torrentId, {
     // use downloads folder
-    path: state.saved.downloads
+    path: state.saved.downloadPath
```
ngoldman: #73 might be because of a mismatch between `state.saved.downloadPath` and `torrent.path`

feross: I think I know what that issue is caused by. Will post a comment about it in a sec.

PR #74, opened 2016-03-07
add empty list state

ngoldman:
screen shot 2016-03-07 at 3 27 05 pm

### DIFF
--- a/renderer/index.css
+++ b/renderer/index.css
@@ -253,7 +253,20 @@ body.drag::before {
  */
 
 .torrent-list {
-  margin-top: 37px;
+  padding-top: 37px;
+}
+
+.get-started {
+  opacity: .5;
+  padding: 24px;
+  margin: 6px;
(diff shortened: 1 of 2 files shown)


renderer/index.css
```diff
@@ -253,7 +253,20 @@ body.drag::before {
...
+  opacity: .5;
+  padding: 24px;
+  margin: 6px;
+  text-align: center;
+  border: 5px #444 dashed;
+  flex: 1 1 auto;
+}
+
+.lead {
+  font-size: 150%;
```
feross: We've been using `em` throughout.

ngoldman: ah alright, I'll change that

feross:
YES! I was about to add this.

renderer/index.css
ngoldman: actually I'm not even using it, feel free to delete it if you get to it first.

PR #81, opened 2016-03-08
prettier bytes: drop fractional number if integer side is more than one digit

### LINKED ISSUE #50: Simplify speed/size numbers in the UI
We want to be the simplest torrent client! :smile: 

No one cares if they're getting 850.62 KB/s or 850.91 KB/s as their download speed.

Let's round with `Math.floor()` in `pretty-bytes`, like we already do for the progress percentage.

screen shot 2016-03-06 at 1 16 34 am

Flet:
untitled

Closes #50

Is this too clever?

### DIFF
--- a/renderer/views/torrent-list.js
+++ b/renderer/views/torrent-list.js
@@ -3,8 +3,16 @@ module.exports = TorrentList
 var h = require('virtual-dom/h')
 var hyperx = require('hyperx')
 var hx = hyperx(h)
+var pb = require('pretty-bytes')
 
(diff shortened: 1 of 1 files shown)


ChrisMorrisOrg:
That's an interesting approach, I kind of like it.

feross:
I think it's better than what we have now. :+1: I think we can continue to hone this further.

I described one possible further improvement here: https://github.com/feross/webtorrent-app/issues/50#issuecomment-193629820

PR #82, opened 2016-03-08
UI polish (Windows fullscreen, Linux/Windows player window title)

feross:
- Fix fullscreen behavior on Windows
  - The win.isFullScreen() function takes a second to return the correct value after going into fullscreen, so we should just pass the state manually into onToggleFullScreen().
- Update player window title correctly (to match torrent name) on OSs that use that native OS chrome (Windows, Linux). Before it only worked on OS X.
  - This also moves all the state “diffing” for purposes of updating the app’s window via Electron APIs into one function `updateElectron()`. `state.prev` is used to store the previous values for diffing purposes.
  - This is necessary because there's no `virtual-dom` for the Electron app state like dock progress, window title, etc. We don't want to just keep re-calling these OS system calls every time `update()` is called since that's a lot of work and will add jank.
  - We can probably move this into a module later to make it clearer.

### DIFF
--- a/main/menu.js
+++ b/main/menu.js
@@ -55,7 +55,6 @@ function onWindowHide () {
   getMenuItem('Float on Top').enabled = false
 }
 
-function onToggleFullScreen () {
 function onToggleFullScreen (isFullScreen) {
   isFullScreen = isFullScreen != null ? isFullScreen : windows.main.isFullScreen()
   windows.main.setMenuBarVisibility(!isFullScreen)
--- a/main/menu.js
+++ b/main/menu.js
@@ -5,19 +5,21 @@ var windows = require('./windows')
 
 var app = electron.app
 
-function toggleFullScreen () {
-  debug('toggleFullScreen')
+function toggleFullScreen (flag) {
+  debug('toggleFullScreen %s', flag)
   if (windows.main && windows.main.isVisible()) {
-    windows.main.setFullScreen(!windows.main.isFullScreen())
+    flag = flag != null ? flag : !windows.main.isFullScreen()
+    windows.main.setFullScreen(flag)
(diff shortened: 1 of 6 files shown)


feross:
@dcposch @ngoldman post-merge review, please

main/menu.js
```diff
@@ -150,12 +154,12 @@ function getMenuTemplate () {
             if (process.platform === 'darwin') return 'Ctrl+Command+F'
             else return 'F11'
           })(),
-          click: toggleFullScreen
+          click: () => toggleFullScreen()
```
dcposch: toggleFullScreen doesn't actually need a boolean argument, right?

I think only `onToggleFullScreen` is taking true or false. `toggleFullScreen()` is always called with no args afaik

renderer/index.js
```diff
@@ -254,10 +272,7 @@ function updateDockIcon () {
   if (!activeTorrentsExist || progress === 1) {
     progress = -1
   }
-  if (progress !== state.dock.progress) {
-    state.dock.progress = progress
-    electron.ipcRenderer.send('setProgress', progress)
-  }
+  state.dock.progress = progress
```
dcposch: if (state.dock) ...

dcposch:
LGTM

main/menu.js
feross: That's correct. I will change this to remove the anonymous function.

renderer/index.js
feross: I actually think this is okay because it's just our `state` object so `state.dock` will always exist. The `if (state.dock)` check exists somewhere else.

feross:
thanks for the review!

PR #91, opened 2016-03-08
Torrent list styling

dcposch:
* Download button is a lot easier to read: white down arrow when off, animated and pulsating green down arrow when downloading, solid green up arrow when seeding

* Play button shows a spinner if you click play before a torrent is ready, then an exclamation point if the torrent still isn't ready after 10 seconds

* Drop target shows up always, not just when the torrent list is empty. Lights up when you drag something

* Fixed alignment, the Xs to delete torrents are now aligned with the + to add a new torrent
  
  ezgif2

### DIFF
--- a/main/windows.js
+++ b/main/windows.js
@@ -26,7 +26,7 @@ function createMainWindow (menu) {
     title: config.APP_NAME,
     titleBarStyle: 'hidden-inset', // Hide OS chrome, except traffic light buttons (OS X)
     width: 450,
-    height: 450
+    height: 600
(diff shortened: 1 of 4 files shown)


dcposch:
One bug: if you drag a file over the app but don't drop it, then move the mouse away again, the drop target stays lit up.

I think this is because of https://github.com/feross/drag-drop/issues/21

dcposch:
@feross might want to retroactively review. also try it / LMK if it feels right!

PR #93, opened 2016-03-08
switch to prettier-bytes

Flet:
This switches to use a just-published package `prettier-bytes` which does the same thing.

https://github.com/flet/prettier-bytes

It also cuts the decimals to a single digit per @rom1504's comment in #50

I'll also add @feross @rom1504 and @yoshuawuyts as collaborators/npm publishers in case folks want to tweak it more.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "hyperx": "^2.0.2",
     "main-loop": "^3.2.0",
     "network-address": "^1.1.0",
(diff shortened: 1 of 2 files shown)


feross:
@Flet Can you please also add @dcposch and @ngoldman to the github and npm?

Flet:
Done and done!

PR #96, opened 2016-03-09
Fix up drag-drop

feross:
Essentially comes down to this commit to `drag-drop`: feross/drag-drop@d65c46d


feross:
Essentially comes down to this commit to `drag-drop`: feross/drag-drop@d65c46d

PR #97, opened 2016-03-09
Fix seeding functionality

feross:
Fix for small oversight in @dcposch’s recent changes. Broken when
adding the distinction between active and inactive torrents.

Please code review, @dcposch


dcposch:
LGTM

PR #107, opened 2016-03-09
dispatch a pause event before hiding window on darwin

### LINKED ISSUE #52: video keeps playing on window close
### steps to repro

On OS X:

screen shot 2016-03-06 at 11 31 59 am

Play a video. Hit the little red button. Sound keeps going.

Since I think we're just showing/hiding the window on close the video never gets paused.

Flet:
Fixes #52

### DIFF
--- a/main/windows.js
+++ b/main/windows.js
@@ -48,7 +48,9 @@ function createMainWindow (menu) {
   win.on('close', function (e) {
     if (process.platform === 'darwin' && !isQuitting) {
       e.preventDefault()
-      win.hide()
+      win.webContents.executeJavaScript('dispatch("pause")', function () {
+        win.hide()
+      })
     }
   })
 
--- a/renderer/index.js
+++ b/renderer/index.js
@@ -187,6 +187,10 @@ function dispatch (action, ...args) {
     // TODO
     // window.history.forward()
   }
+  if (action === 'pause') {
+    state.video.isPaused = true
+    update()
+  }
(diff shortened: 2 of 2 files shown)


main/windows.js
```diff
@@ -48,7 +48,9 @@ function createMainWindow (menu) {
   win.on('close', function (e) {
     if (process.platform === 'darwin' && !isQuitting) {
       e.preventDefault()
-      win.hide()
+      win.webContents.executeJavaScript('dispatch("pause")', function () {
+        win.hide()
+      })
```
feross: We should find a cleaner way to dispatch this. I'll merge then fix.

main/windows.js
```diff
@@ -48,7 +48,9 @@ function createMainWindow (menu) {
   win.on('close', function (e) {
     if (process.platform === 'darwin' && !isQuitting) {
       e.preventDefault()
-      win.hide()
+      win.webContents.executeJavaScript('dispatch("pause")', function () {
```
feross: It'd be better to use IPC for this. I'll merge and fix up.

main/windows.js
feross: Done here: https://github.com/feross/webtorrent-app/commit/a6e8a3462ee1241b9fb5e24fb023cc43060b6b95

Flet:
Thanks ipc looks a lot cleaner!

PR #121, opened 2016-03-10
Create WebTorrent app icon

feross:
Also added packager build script. This add lots of missing options for packaging Electron apps correctly
for all platforms.

Here's the icon:

screen shot 2016-03-09 at 11 42 27 pm


feross:
@ngoldman Good idea, done.

ngoldman:
I gave you a :tada:

PR #129, opened 2016-03-11
Shrink Electron bundle by MBs!

feross:
Before: 127.6MB (48 MB zipped)
After: 120.5MB (42.8 MB zipped)

I accomplished this by enabling the ASAR archive format, and the (slightly gross) method of excluding files in `node_modules` that are likely to be unneeded.

Example: `test/`, `doc/`, etc.


bin/package.js
```diff
@@ -109,11 +109,33 @@ var linux = {
...
+  electronPackager(Object.assign({}, all, darwin), done.bind(null, cb))
+}
+
+function buildWin32 (cb) {
+  electronPackager(Object.assign({}, all, win32), done.bind(null, cb))
+}
+
+function buildLinux (cb) {
+  electronPackager(Object.assign({}, all, linux), done.bind(null, cb))
+}
```
ngoldman: d'oh, I was just working on making the packager accept specific target platforms.

feross: Oh, sorry bout that! Let's chat about what we're working on on IRC!

package.json
```diff
@@ -49,7 +49,8 @@
   "scripts": {
     "cleanup": "node ./bin/cleanup.js",
     "debug": "DEBUG=* electron .",
-    "package": "node ./bin/package.js",
+    "package": "npm prune && npm dedupe && node ./bin/package.js",
+    "size": "npm run package -- --darwin && du -ch dist/WebTorrent-darwin-x64 | grep total",
```
ngoldman: is this script just to get a file size?

bin/package.js
ngoldman: np I should have checked in :)

ngoldman:
lgtm :+1:

package.json
feross: Yep, exactly.

feross:
Thanks for reviewing and merging. Nice new avatar, btw :+1:

ngoldman:
Thanks! It's an old favicon I made that's been blown up, I still need to make a bigger version :laughing:

feross:
Cool, the pixelated version looks cool too :)

Issue #77, opened 2016-03-08
Set WebTorrent as the default handler for .torrent files and magnet links


feross:
Let's add a button to the Preferences Pane #62 for enabling this for magnet links. Like with lots of other things, I think Transmission gets this UI and wording right:

<img width="344" alt="screen shot 2016-03-07 at 6 24 27 pm" src="https://cloud.githubusercontent.com/assets/121766/13590080/edcc4cd2-e491-11e5-9dbe-54625de8b03c.png">

For .torrent files, I think we just register with the OS that our application can open .torrent files and leave it up to the OS and user to configure WebTorrent as their default.

## Tasks 

- [x] OS X: Handle .torrent files
- [x] OS X: Handle magnet links
- [x] Windows: Handle .torrent files
- [x] Windows: Handle magnet links
- [x] Linux: Handle .torrent files
- [x] Linux: Handle magnet links

feross:
Ubuntu protocol handler intsructions: https://askubuntu.com/questions/514125/url-protocol-handlers-in-basic-ubuntu-desktop

feross:
Okay, since I just got Windows magnet links working, going to just do .torrent file handling too.

feross:
This is done! Whew, that was a lot of hacky, platform-specific code!

PR #132, opened 2016-03-11
enable os x file drop, add osx torrent file icon

### LINKED ISSUE #131: [Mac OS X] Accept torrent files on dock drop
I want to drop a file onto the WebTorrent dock icon and have WebTorrent open that file. Atom and Playback do it so I know it's possible. PR incoming.

ngoldman:
- fixes #131
- partially addresses #77 for osx (now handles .torrent files)

screen shot 2016-03-10 at 9 48 01 pm

:revolving_hearts: lots of love to go around now :revolving_hearts:


bin/package.js
```diff
@@ -134,6 +134,37 @@ function buildLinux (cb) {
...
+    CFBundleTypeExtensions: [ 'torrent' ],
+    CFBundleTypeName: 'BitTorrent Document',
+    CFBundleTypeRole: 'Editor',
+    CFBundleTypeIconFile: 'WebTorrentFile.icns'
+  }]
+
+  fs.writeFileSync(infoPlistPath, plist.build(infoPlist))
+  cp.execSync(`cp ${webTorrentFileIconPath} ${resourcesPath}`)
+}
+
```
ngoldman: I decided to make everything concerning osx contained within this function since other platforms wouldn't need `plist` or anything else in here. Still a little weird, but I feel like it's cleaner than what Playback or Atom are doing.

bin/package.js
```diff
@@ -134,6 +134,37 @@ function buildLinux (cb) {
   electronPackager(Object.assign({}, all, linux), done.bind(null, cb))
 }
 
+function postDarwinism () {
+  var plist = require('plist')
```
feross: Is this in here because it crashes on non-OSX platforms?

bin/package.js
```diff
@@ -112,7 +112,7 @@ var linux = {
 var platform = process.argv[2]
 
 if (platform === '--darwin') {
-  buildDarwin()
+  buildDarwin(postDarwinism)
```
feross: cute

bin/package.js
feross: This is quite clean!

bin/package.js
ngoldman: Don't know, but figured it should be a conditional require since it's used exclusively for Mac OS X builds.

feross:
LGTM!

PR #133, opened 2016-03-11
ensure video is paused before hiding window

Flet:
`sendSync` is throwing an error currently as it does not exist.

This PR adds logic to signal the video to stop playing and only hide after we're sure the video is paused. 

Feedback appreciated!


feross:
LGTM!

feross:
I still can't help but feel that there's gotta be a cleaner way than this, but at least it works now :)

PR #136, opened 2016-03-11
File details

dcposch:
image

### DIFF
--- a/renderer/views/torrent-list.js
+++ b/renderer/views/torrent-list.js
@@ -24,13 +24,15 @@ function renderTorrent (torrentSummary, state, dispatch) {
   // Get ephemeral data (like progress %) directly from the WebTorrent handle
   var infoHash = torrentSummary.infoHash
   var torrent = state.client.torrents.find((x) => x.infoHash === infoHash)
+  var isSelected = state.selectedInfoHash === infoHash
 
   // Background image: show some nice visuals, like a frame from the movie, if possible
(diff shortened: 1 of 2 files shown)


dcposch:
Coming next:
- Play button next to files that are playable
- Hide the big play button for torrents that contain no playable files
- Open Folder link

renderer/views/torrent-list.js
```diff
@@ -114,3 +118,45 @@ function renderTorrentButtons (torrentSummary, dispatch) {
...
+function renderTorrentDetails (torrent, torrentSummary) {
+  var filesElement
+  if (!torrent || !torrent.files) {
+    // We don't know what files this torrent contains
+    var message = torrent
+      ? 'Downloading torrent data using magnet link...'
+      : 'Failed to download torrent data from magnet link. Click the download button to try again...'
+    filesElement = hx`<div class='files warning'>${message}</div>`
+  } else {
+    // We do know the files. List them and show download stats for each one
```
dcposch: Ideally this should be in the webtorrent library

Currently the webtorrent file object just has a property, `done`, that says whether all pieces are loaded. Maybe we could add `numPieces` and `numPiecesLoaded` to show progress.

This would also be more efficient, since webtorrent can increment `numPieces` every time a new piece comes in, instead of us iterate thru all pieces in the file to see which ones are there

feross:
LGTM!

PR #139, opened 2016-03-12
add platform-specific package instructions

ungoldman:
I definitely don't want to build for all three platforms every time I'm testing :)


feross:
lgtm!

PR #140, opened 2016-03-12
Block "Power Saver" mode while video is playing

### LINKED ISSUE #106: Block "Power Saver" mode while video is playing
> The `powerSaveBlocker` module is used to block the system from entering low-power (sleep) mode and thus allowing the app to keep the system and screen active.

http://electron.atom.io/docs/v0.36.8/api/power-save-blocker/

ungoldman:
Using `video.onplay`, `video.onpause`, `dispatch`, `ipc` messages, and `electron.powerSaveBlocker`. I tested for unexpected errors getting thrown but haven't sat around waiting for the display to go to sleep yet. My belief is that it will Just Work™.

> fix #106

### DIFF
--- a/renderer/index.js
+++ b/renderer/index.js
@@ -206,6 +206,9 @@ function dispatch (action, ...args) {
     state.video.isPaused = true
     update()
   }
+  if (action === 'videoPlaying') {
+    ipcRenderer.send('playing-video')
+  }
   if (action === 'videoPaused') {
     ipcRenderer.send('paused-video')
   }
--- a/main/ipc.js
+++ b/main/ipc.js
@@ -9,6 +9,10 @@ var windows = require('./windows')
 
 var app = electron.app
 var ipcMain = electron.ipcMain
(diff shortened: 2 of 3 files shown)


renderer/index.js
```diff
@@ -206,6 +206,9 @@ function dispatch (action, ...args) {
     state.video.isPaused = true
     update()
   }
+  if (action === 'videoPlaying') {
+    ipcRenderer.send('playing-video')
+  }
   if (action === 'videoPaused') {
     ipcRenderer.send('paused-video')
```
ungoldman: should probably normalize these messages at some point

feross: yep, agreed

ungoldman: message/action names as constants in a file required by both main and renderer would do the trick

feross: Not a fan of using constants in this way. Strings are fine, just need to be consistent.

feross:
Just tested this out and it it's working. Just noticed one bug: https://github.com/feross/webtorrent-app/issues/141

PR #142, opened 2016-03-12
fix #141

### LINKED ISSUE #141: Closing video player without pausing causes power saving mode to remain disabled
cc @ungoldman

ungoldman:
> fix #141

### DIFF
--- a/renderer/index.js
+++ b/renderer/index.js
@@ -79,8 +79,6 @@ function init () {
     updateClientProgress()
   }, 1000)
 
-  // All state lives in state.js. `state.saved` is read from and written to a
-  // file. All other state is ephemeral.
   window.addEventListener('beforeunload', saveState)
 
   // listen for messages from the main process
@@ -201,6 +199,7 @@ function dispatch (action, ...args) {
   if (action === 'back') {
     // TODO
     // window.history.back()
+    ipcRenderer.send('unblockPowerSave')
     closePlayer()
   }
   if (action === 'forward') {
@@ -215,10 +214,11 @@ function dispatch (action, ...args) {
     update()
   }
   if (action === 'videoPlaying') {
-    ipcRenderer.send('playing-video')
+    ipcRenderer.send('blockPowerSave')
   }
   if (action === 'videoPaused') {
     ipcRenderer.send('paused-video')
+    ipcRenderer.send('unblockPowerSave')
   }
   if (action === 'playPause') {
     state.video.isPaused = !state.video.isPaused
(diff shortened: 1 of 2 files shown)


renderer/index.js
```diff
@@ -533,11 +534,25 @@ function closePlayer () {
...
+  powerSaveBlockIDs.push(powerSaveBlocker.start('prevent-display-sleep'))
+}
+
+function unblockPowerSave () {
+  powerSaveBlockIDs.filter((id) => {
+    return powerSaveBlocker.isStarted(id)
+  }).map((id) => { powerSaveBlocker.stop(id) })
+
+  powerSaveBlockIDs = []
+}
```
ungoldman: This is a little weird but feels better to me than reassigning a variable every time it's stopped. If we ever have multiple video players this could expand to deal with that but I think it's sufficient for now.

renderer/index.js
```diff
@@ -533,11 +534,25 @@ function closePlayer () {
   if (state.window.isFullScreen) {
     dispatch('toggleFullScreen', false)
   }
+
+  unblockPowerSave()
```
ungoldman: This is getting called inside `closePlayer()`

renderer/index.js
```diff
@@ -533,11 +534,25 @@ function closePlayer () {
...
   restoreBounds()
   stopServer()
   update()
 }
 
+function blockPowerSave () {
+  powerSaveBlockIDs.push(powerSaveBlocker.start('prevent-display-sleep'))
```
dcposch: why not have just one `powerSaveBlockID`?

Then, in `blockPowerSave`:

```js
if (!powerSaveBlocker.isStarted(powerSaveBlockID)) {
  powerSaveBlockID = powerSaveBlocker.start(...)
}
```

...and similarly in `unblockPowerSave`:

```js
if (!powerSaveBlocker.isStarted(powerSaveBlockID)) {
  powerSaveBlocker.stop(powerSaveBlockID)
}
```

renderer/index.js
dcposch: I'd make this a comment in the code.

Also, I understand that we have to restart power saving mode when the player window exits, not just when the video is paused, but could we just say `ipcRenderer.send('closing-player')` here or smth similar? (Or just catch an event in the main process that fires when the app is about to close?) Either way, I think OS integrations like this are best handled from the main process.

This is a nitpick tho and not a big deal.

dcposch:
LGTM after comments

ungoldman:
Okay, this should resolve #141 more cleanly than the last PR.

dcposch:
:100:

feross:
Very ncie!

PR #144, opened 2016-03-13
Code Review: Basis for preferences window

ChrisMorrisOrg:
Just wanting feedback on what I've written so far. This isn't ready for a merge.

Any review/suggestions would be great. Currently it doesn't seem to update the downloadPath, gotta look into that. Secondly, it doesn't currently display the existing downloadPath because I'm not sure what the best way to handle that is. I've tried to mimic the pattern used in the main app, which involved putting the dialogue box call in menu.js, not sure how I feel about that, but it's a central point that has access to windows.js. I figured there's a reason not to include windows.js within the preferences.js file itself.

### DIFF
--- a/main/menu.js
+++ b/main/menu.js
@@ -4,6 +4,7 @@ module.exports = {
   onWindowHide: onWindowHide,
   onWindowShow: onWindowShow,
   showOpenTorrentFile: showOpenTorrentFile,
+  showSelectDownloadDirectory: showSelectDownloadDirectory,
   toggleFullScreen: toggleFullScreen
 }
 
@@ -43,8 +44,11 @@ function toggleFloatOnTop (flag) {
 
 function toggleDevTools () {
   debug('toggleDevTools')
-  if (windows.main) {
+  /* A quick hack to show dev tools for the preferences window if it's open */
+  if (!windows.preferences) {
     windows.main.toggleDevTools()
+  } else {
+    windows.preferences.toggleDevTools()
   }
 }
 
@@ -110,6 +114,21 @@ function showOpenTorrentFile () {
     filenames.forEach(function (filename) {
       windows.main.send('dispatch', 'addTorrent', filename)
     })
+  })
+}
+
+// Open the preferences window
+function showPreferencesWindow () {
+  windows.createPreferencesWindow()
+}
+
+function showSelectDownloadDirectory() {
+  electron.dialog.showOpenDialog(windows.preferences, {
+    title: 'Select a directory to save in.',
+    defaultPath: '/',
+    properties: [ 'openDirectory' ]
+  }, function (directoryname) {
+    windows.main.send('dispatch', 'setDefaultDownloadDirectory', directoryname)
   })
 }
 
@@ -266,6 +285,14 @@ function getAppMenuTemplate () {
         {
           label: 'About ' + name,
           role: 'about'
+        },
+        {
+          type: 'separator'
+        },
+        {
+          label: 'Preferences',
+          accelerator: 'Cmd+,',
+          click: () => showPreferencesWindow()
         },
         {
           type: 'separator'
--- a/main/menu.js
+++ b/main/menu.js
@@ -122,9 +122,9 @@ function showPreferencesWindow () {
   windows.createPreferencesWindow()
 }
 
-function showSelectDownloadDirectory() {
+function showSelectDownloadDirectory () {
   electron.dialog.showOpenDialog(windows.preferences, {
-    title: 'Select a directory to save in.',
+    title: 'Choose a folder for your downloads...',
     defaultPath: '/',
     properties: [ 'openDirectory' ]
   }, function (directoryname) {
--- a/renderer/preferences.js
+++ b/renderer/preferences.js
@@ -1,0 +1,62 @@
+var electron = require('electron')
+
+var vdom = require('virtual-dom')
+var hyperx = require('hyperx')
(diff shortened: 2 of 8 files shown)


main/menu.js
```diff
@@ -271,6 +290,14 @@ function getAppMenuTemplate () {
           type: 'separator'
         },
         {
+          label: 'Preferences',
+          accelerator: 'Cmd+,',
+          click: () => showPreferencesWindow()
```
dcposch: You can also say

```js
click: showPreferencesWindow
```

renderer/preferences.js
```diff
@@ -0,0 +1,62 @@
...
+      </div>
+    </div>`
+  }
+
+  function onclick () {
+    loop.update({ times: state.times + 1 })
+  }
+}
+
+function getDefaultDownloadDirectory() {
```
dcposch: ?

renderer/preferences.js
```diff
@@ -0,0 +1,62 @@
...
+// Events from the UI never modify state directly. Instead they call dispatch()
+function dispatch (action, ...args) {
+  if (action === 'showSelectDownloadDirectory') {
+    ipcRenderer.send('showSelectDownloadDirectory')
+  }
+}
+
+function showSelectDownloadDirectory() {
+  electron.dialog.showOpenDialog(windows.preferences, {
+    title: 'Select a directory to save in.',
```
dcposch: Maybe "Choose a folder for your downloads..." ?

renderer/preferences.js
```diff
@@ -0,0 +1,62 @@
...
+    defaultPath: '/Users',
```
dcposch: http://stackoverflow.com/questions/9080085/node-js-find-home-directory-in-platform-agnostic-way

main/menu.js
```diff
@@ -113,6 +117,21 @@ function showOpenTorrentFile () {
...
 }
 
+// Open the preferences window
+function showPreferencesWindow () {
+  windows.createPreferencesWindow()
+}
+
+function showSelectDownloadDirectory() {
+  electron.dialog.showOpenDialog(windows.preferences, {
+    title: 'Select a directory to save in.',
```
dcposch: This looks like the right place for `showSelectDownloadDirectory`. I'd delete the other copy of this function down in `renderer/preferences.js`

dcposch:
Looks nice so far! Left some comments. Thanks for doing this.

When you think the prefs dialog looks good, I recommend pasting a screenshot into the PR. Github makes this v smooth. On Ubuntu it's Alt + PrintScr, Copy to Clipboard, then Ctrl V directly into a Github comment.

renderer/preferences.js
ChrisMorrisOrg: So, I just put `${'[downloadPath]'}` because I'm not sure what the best way to actually obtain the current state.downloadPath would be.

getDefaultDownloadDirectory(), I forgot to remove that - I was about to start experimenting with getting the state.downloadPath.

renderer/preferences.js
ChrisMorrisOrg: Yeah, that sounds a lot better!

renderer/preferences.js
ChrisMorrisOrg: I actually only put it there for the sake of experimentation. It needs to be state.downloadPath. Thanks for the link, though! :)

main/menu.js
ungoldman: That's not always the case: see https://github.com/feross/webtorrent-app/pull/130

renderer/preferences.js
ungoldman: Here's chrome download settings for reference:

screen shot 2016-03-13 at 9 17 36 pm

Simple and brief is always better imo :)

ChrisMorrisOrg:
Thanks for the feedback @dcposch, @ungoldman! I've made several changes, fixing the issues mentioned and taking suggestions from both of you :).

The only issue I'm having now is actually retrieving the state.downloadPath. Currently, the main window creates a new App() instance with state passed in. I could take a similar approach and make a Preferences() class, but then we'd be maintaining two state objects, which seems like a bad idea.

I'm not sure what the best way to tackle this is. Any ideas? I feel like this is a nontrivial problem, solutions may require a large code refactor.

screen shot 2016-03-14 at 7 06 30 pm

dcposch:
@ChrisMorrisOrg i think there's two things we could do:

1. Move state to the main process. This is a big refactor and might suck in other ways, but it would ensure that there's exactly one state object.
2. Make preferences their own section of the state, `state.saved.prefs`. (The rest of `state.saved` contains things like the torrent list, which is persisted but is not part of "Preferences".) Then, every time we open the prefs modal, we pass in `state.saved.prefs`. The modal has its own copy of that object, which it modifies as the user interacts with the form inputs. Finally, if the user clicks OK, the we send an IPC to the main process, `setPreferences`, which sets the new value of `state.saved.prefs` .

I like option 2. It's easier to implement, and it means that all prefs changes take effect simultaneously when the user clicks OK, rather than piece by piece as they interact with the prefs modal.

ungoldman:
@dcposch's option 2 sounds pretty good to me as well.

feross:
:+1: to option 2. Also, it's worth noting that on OS X, Preferences windows don't have "Save" buttons. When you change a setting, it's immediately saved. We can add the OK, Cancel buttons for other OSs, since users will expect them,

dcposch:
Superseded by #426, right?

PR #148, opened 2016-03-15
Chromecast video controls

dcposch:
image


feross:
Just took a quick look through this. Looks good so far! What's left to do?

dcposch:
I think Chromecast is good to go. Still have to finish Airplay but I'll make a separate PR for that

PR #157, opened 2016-03-19
Cleanup build script; Run plist update function for all builds

feross:
Before, the plist update code wouldn't run when building for all platforms. Fixed that and refactored the build script a bit.


dcposch:
LGTM

PR #168, opened 2016-03-20
fix hide on hover, always show playback position

ungoldman:
- fixed hide on hover (`state.video.location` should be `state.playing.location`)
- showing scrubber without needing to hover on playback bar
- added 5px of margin right to full screen button to even things out a bit

Now hover looks like this (same):
screen shot 2016-03-19 at 8 51 37 pm

Without hover (playback time is now visible):
screen shot 2016-03-19 at 8 51 33 pm

After 2 seconds (controls hide after delay again as they should):
screen shot 2016-03-19 at 8 54 01 pm

### DIFF
--- a/renderer/index.css
+++ b/renderer/index.css
@@ -633,13 +633,13 @@ body.drag .torrent-placeholder span {
 
 .player-controls .playback-cursor {
   position: absolute;
-  top: -8px;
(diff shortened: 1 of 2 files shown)


feross:
LGTM!

PR #171, opened 2016-03-20
Airplay support. Nicer looking cast screen

dcposch:
image


feross:
This looks really nice!

PR #185, opened 2016-03-21
add volume management

grunjol:
DONE
* add volume management in local player, chromecast and airplay
* controlled by shortcuts (CtrlOrCmd+Up, CtrlOrCmd+Down)

TODO:
* test chromecast and airplay
* airplay: could not find a way to get volume in poll cast status, it writes state value in setter by now (help @dcposch !)
* UI control and status (maybe a HUD like in VLC)


renderer/index.js
```diff
@@ -258,6 +261,22 @@ function jumpToTime (time) {
...
+  setVolume(state.video.volume + delta)
+}
+
+function setVolume (volume) {
+  // check if its in [0.0 - 1.0] range
+  volume = Math.max(0, Math.min(1, volume))
+  if (Cast.isCasting()) {
+    Cast.setVolume(volume)
+  } else {
+    state.video.setVolume = volume
```
dcposch: Probably want to call this `state.video.volume`

Also, it looks like there are two separate volumes now: one for local playback and another for remote (cast) playback. I think that's fine

dcposch: NVM, I understand the difference between `volume` and `setVolume` now

dcposch:
This looks good to me!

We probably want to display the volume in the video controls and maybe also on the cast screen.

grunjol:
yes, UI controls and level feedback (i was thinking in a OSD like in VLC) is a needed.

grunjol:
gonna make it right

PR #193, opened 2016-03-21
Speed up Travis

dcposch:


rom1504:
what's the difference between this and https://github.com/feross/webtorrent-app/blob/master/package.json#L38 ?

feross:
@rom1504 This makes travis *only* run `npm install standard` instead of running `npm install`. This is way faster since Electron doesn't need to be installed.

PR #196, opened 2016-03-22
Use largest image in torrent as the poster

dcposch:
...as a fallback, if no the torrent doesn't contain any video files

image


feross:
Nice approach. File size is probably a good enough proxy for image size.

Issue #12, opened 2016-03-05
Releasing builds for linux/mac/windows


rom1504:
Is it easy to build for several platform and put it in a github release ?

Anyway I think it can be important ,once the app is ready, for people that don't want to install node.js (themselves). Especially on windows(/mac).

ngoldman:
Yep, there's https://github.com/electron-userland/electron-packager or https://www.npmjs.com/package/electron-builder for building, and https://www.npmjs.com/package/publish-release or https://github.com/ngoldman/gh-release for releasing to github. Probably lots of others out there too.

feross:
Yeah, so far we haven't introduced any native dependencies. As long as we're careful never to do that, we won't need to manually compile on each platform in order to produce an executable for each platform.

You can run `npm run build` right now to produce final release builds for all platforms. It's currently using `electron-packager` :)

feross:
OS X apps are usually bundled into a `.dmg` file so the user can see a friendly "Drag this app onto the Applications folder" Finder window. Is there an easy way to get `electron-packager` to build us one of those?

How do we bundle the Windows version? Looks like there's lots of files in the output folder, like .DLLs, that are required by, but not embedded within, `WebTorrent.exe`. No idea how to ship this to users.

Same with the Linux release. Can we make a `.deb` file for Ubuntu/Debian?

cc @maxogden @mafintosh - do you guys have any thoughts?

maxogden:
Check out https://www.npmjs.com/package/electron-builder, though I haven't
used it

feross:
The first WebTorrent release (v0.0.0) is live! You can get it from the GitHub releases page here: 

https://github.com/feross/webtorrent-app/releases/tag/v0.0.0

There's a .dmg file for easy installation on OS X. There is a built-in auto-updater so users will get improvements automatically. New releases (if any exist) are downloaded over `https` on app startup, checked to make sure they're signed correctly, and installed the next time the app is restarted.

I'm going to release v0.0.1 later tonight to test that the auto-updater is working as expected!

feross:
We have releases for all platforms. Windows has an installer now. The linux one is just a folder of files for now. Closing this!

Issue #190, opened 2016-03-21
Auto-updater


feross:
- [x] OS X
- [x] Windows
- [x] ~~Linux~~

feross:
I'm working on this right now.

feross:
Got OS X working – will push tomorrow. Time for sleep.

feross:
Electron does have a built in auto-update solution for Linux, so I'm going to scratch it from the list.

More info: http://electron.atom.io/docs/all/#linux

feross:
For Linux, I think we can just release .deb files for now investigate creating a PPA, or submitting to the Ubuntu Software Center, perhaps.

feross:
For Windows, I've applied for an Authenticode cert. Once that's issued, I'll be able to sign the Windows releases and the auto-updater should work. I'll push the work-in-progress version (without signing support) to a branch.

feross:
https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Build_Instructions/Signing_an_executable_with_Authenticode

c0b41:
@feross any progress for linux?

grunjol:
I am working on a deb package generator (PR coming soon)

Regarding auto-update, i see 2 options

* check version and provide download link (current implementation)
* system update 
 * create a deb/rpm repository in webtorrent infrastructure
 * use an external repository (launchpad, bintray)

I use [bintray](https://bintray.com) in some private projects and its very versatile (offers deb, rpm, maven, docker and others), and its free for open source projects

c0b41:
@grunjol hey launchpad or bintray any progress? now linux update useless.

also travis bintray deployment cool [checkout](https://docs.travis-ci.com/user/deployment/bintray)

gtomas41:
Please publish a RPM

PR #197, opened 2016-03-22
OS X: Add auto-updater, sign app, build .DMG and .ZIP assets

feross:
- Add [Squirrel.Mac](https://github.com/Squirrel/Squirrel.Mac) auto-updater (it's built in to electron)
  - Requires [server support](https://github.com/Squirrel/Squirrel.Mac#server-support) which I [added to `webtorrent-www`](https://github.com/feross/webtorrent-www/pull/46) (https://webtorrent.io)
- Sign app with Apple developer certificate, for two reasons:
  - Required for `Squirrel.Mac` to verify app updates are signed by the same author as the current version.
  - Allows app to run without the user needing to disable [OS X Gatekeeper](https://developer.apple.com/developer-id/).
- Signing OS X apps for distribution outside the App Store requires:
    - Xcode
    - Xcode Command Line Tools (xcode-select --install)
    - Membership in the Apple Developer Program
    - I (Feross) will sign every OS X release and ensure that the certificate is safe.
- Produce a .DMG file, for easy installs by the user (#12)
  - Here's how it looks:
  screen shot 2016-03-21 at 8 28 24 pm
- Produce a .ZIP file, which is downloaded and used by the auto-updater.


ungoldman:
Looks great!

feross:
For #190

feross:
@ungoldman Thanks!

feross:
The first release is available here: https://github.com/feross/webtorrent-app/releases/tag/v0.0.0

PR #202, opened 2016-03-22
Add volume management

grunjol:
## DONE

* add volume management in local player, chromecast and airplay
* controlled by shortcuts (CtrlOrCmd+Up, CtrlOrCmd+Down)

## TODO

* test chromecast and airplay
* airplay: could not find a way to get volume in poll cast status, it writes state value in setter by now 
* UI control and status (maybe a HUD like in VLC)

Cleanup from PR #185

### DIFF
--- a/main/shortcuts.js
+++ b/main/shortcuts.js
@@ -18,4 +18,8 @@ function init () {
   // Electron does not support multiple accelerators for a single menu item, so this
   // is registered separately here.
   localShortcut.register('CmdOrCtrl+Shift+F', menu.toggleFullScreen)
+
+  // Control Volume
+  globalShortcut.register('CmdOrCtrl+Up', () => windows.main.send('dispatch', 'changeVolume', 0.1))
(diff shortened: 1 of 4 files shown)


main/shortcuts.js
```diff
@@ -18,4 +18,8 @@ function init () {
   // Electron does not support multiple accelerators for a single menu item, so this
   // is registered separately here.
   localShortcut.register('CmdOrCtrl+Shift+F', menu.toggleFullScreen)
+
+  // Control Volume
+  globalShortcut.register('CmdOrCtrl+Up', () => windows.main.send('dispatch', 'changeVolume', 0.1))
+  globalShortcut.register('CmdOrCtrl+Down', () => windows.main.send('dispatch', 'changeVolume', -0.1))
```
feross: We should not use `globalShortcut` here. This will register the shortcut at the OS level, even when the app is not focused.

Using `localShortcut` would work, but let's put it in the top menu instead, where all the other shortcuts are.

feross:
Nice!

PR #194, opened 2016-03-21
Implement #116: Add Unity Launcher Shortcuts

grunjol:
This PR solves #116 

To implement this the following changes were made

* parameters are not longer torrentId only, handle that with processArgv function
* expose missing menu event invocation (showCreateTorrent)
* add Actions section in .desktop
* fix .desktop/register handler to be used in development mode

PR #203, opened 2016-03-22
Add unity launcher icons

### LINKED ISSUE #116: Add Unity Launcher Shortcuts (Linux)
In Unity, you can add custom entries to its launcher. Similar to the OS X Dock issue #113, let's add Unity Launcher Shortcuts for Linux:

shortcuts

API Docs: http://electron.atom.io/docs/v0.36.8/tutorial/desktop-environment-integration/#unity-launcher-shortcuts-linux

grunjol:
This PR solves #116 

To implement this the following changes were made

* parameters are not longer torrentId only, handle that with processArgv function
* expose missing menu event invocation (showCreateTorrent)
* add Actions section in .desktop
* fix .desktop/register handler to be used in development mode

Cleanup from PR #194
Thanks @ungoldman for the branch management tip

### DIFF
--- a/main/handlers.js
+++ b/main/handlers.js
@@ -27,8 +27,14 @@ function installDesktopFile () {
   var desktopFile = fs.readFileSync(templatePath, 'utf8')
 
   desktopFile = desktopFile.replace(/\$APP_NAME/g, config.APP_NAME)
-  desktopFile = desktopFile.replace(/\$APP_PATH/g, path.dirname(process.execPath))
-  desktopFile = desktopFile.replace(/\$EXEC_PATH/g, process.execPath)
+
+  var appPath = config.IS_PRODUCTION ? path.dirname(process.execPath) : config.ROOT_PATH
+  desktopFile = desktopFile.replace(/\$APP_PATH/g, appPath)
+
+  var execPath = process.execPath + (config.IS_PRODUCTION ? '' : ' \.')
+  desktopFile = desktopFile.replace(/\$EXEC_PATH/g, execPath)
+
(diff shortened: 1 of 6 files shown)


grunjol:
webtorrent-launcher-icons-menu

@feross update with latest CLI changes

feross:
@grunjol Thanks, taking a look now!

main/handlers.js
```diff
@@ -27,8 +27,14 @@ function installDesktopFile () {
   var desktopFile = fs.readFileSync(templatePath, 'utf8')
 
   desktopFile = desktopFile.replace(/\$APP_NAME/g, config.APP_NAME)
-  desktopFile = desktopFile.replace(/\$APP_PATH/g, path.dirname(process.execPath))
-  desktopFile = desktopFile.replace(/\$EXEC_PATH/g, process.execPath)
+
+  var appPath = config.IS_PRODUCTION ? path.dirname(process.execPath) : config.ROOT_PATH
+  desktopFile = desktopFile.replace(/\$APP_PATH/g, appPath)
+
+  var execPath = process.execPath + (config.IS_PRODUCTION ? '' : ' \.')
```
feross: Can you explain what this is doing?

feross: I see, you need to pass the app path into the electron binary for development. Fair enough.

But I don't think this needs to be escaped. It's just a plain string, so `.` should be sufficient. I will test and if this is the case, I will change it.

feross:
Looks good to me!

ungoldman:
:100:

main/handlers.js
feross: Yep, `'` works.

PR #212, opened 2016-03-24
fix invisible header bug

### LINKED ISSUE #211: [ui bug] header not visible when player active
Looks like some recent CSS changes have made the header invisible when the player is active. Also means user is stuck in player mode and has to quit to get back to index.

screen shot 2016-03-23 at 11 09 51 pm

PR incoming.

ungoldman:
fixes #211 

screen shot 2016-03-23 at 11 19 02 pm

works as expected in other states (fullscreen, idle, index). easy one line fix. :tophat: :rabbit2:

### DIFF
--- a/renderer/index.css
+++ b/renderer/index.css
@@ -184,6 +184,7 @@ i:not(.disabled):hover {
   transition: opacity 0.15s ease-out;
   font-size: 14px;
   line-height: 1.5em;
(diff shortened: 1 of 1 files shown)


feross:
Thanks. @dcposch Looks like the z-index was needed afterall.

Issue #224, opened 2016-03-26
Ubuntu first-run crash


feross:
<img width="490" alt="screen shot 2016-03-25 at 6 36 48 pm" src="https://cloud.githubusercontent.com/assets/121766/14057311/180a1b84-f2b9-11e5-8066-4e1fe409d334.png">

grunjol:
@feross you are right, if you run it with a clear OS installation the icons directory is not present.
PR sent

feross:
Thanks for the PR, but I already fixed this in a local branch :)

PR #225, opened 2016-03-26
fix ubuntu first-run crash #224

grunjol:
Check the existence of $HOME/.local/share/icon directory and create it if it doesn't exist.


feross:
Nice solution, but we already depend on `mkdirp` so we'll just use that. I already used it in another commit.

PR #229, opened 2016-03-26
Use relative paths for default torrents

dcposch:
This keeps them working if a user opens the app from DMG, then installs it to a different path and opens it again

### DIFF
--- a/renderer/state.js
+++ b/renderer/state.js
@@ -64,8 +64,8 @@ module.exports = {
         status: 'paused',
         infoHash: '88594aaacbde40ef3e2510c47374ec0aa396c08e',
         displayName: 'Big Buck Bunny',
-        posterURL: path.join(config.ROOT_PATH, 'static', 'bigBuckBunny.jpg'),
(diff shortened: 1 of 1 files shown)


ungoldman:
@dcposch you need to use `path.join` or those paths won't work on windows afaik.

dcposch:
@ungoldman i think it's fine -- i'll test it to make sure, but due to this Chrome bug workaround all paths are string replaced to use forward slash anyway

https://github.com/feross/webtorrent-app/blob/fc425e422154f3f79876e44f6c128b8a086d3c0f/renderer/views/torrent-list.js#L36

ungoldman:
Weird, maybe that's different between main and renderer processes? Or is just not a thing that matters in electron apps? Or anywhere anymore? I was pretty sure this was a major thing for working cross-platform from CLI programs I wrote over the last few years. Now I'm just full of doubts.

feross:
Even though it might not matter on Windows, let's keep using `path.join` consistently throughout the codebase. It provides protection against putting two slashes in the path which prevents bugs.

Also, `standard` enforces it [when it can determine](http://eslint.org/docs/rules/no-path-concat) you're using string concatenation with path. :+1:

PR #233, opened 2016-03-27
add context menu with share/save actions

### LINKED ISSUE #215: Add torrent "share" links
When you add *any* torrent in webtorrent-app, it's instantly becomes available in instant.io.

It might be interesting to add a share link to instant.io for that torrent.

edit: it's particularly obvious that it works if you open that torrent in instant.io in the same host since it's seeded in localhost ! (of course localhost is not super useful, but for example sharing the link with someone on your local network could be nice)

---

TODO: (added by @feross)

- [ ] Add "Save Torrent File As..."
- [ ] Add "Copy Magnet Link to Clipboard"
- [ ] Add "Copy Instant.io Link to Clipboard"

Flet:
Closes #215

### DIFF
--- a/renderer/index.js
+++ b/renderer/index.js
@@ -30,6 +30,7 @@ var torrentPoster = require('./lib/torrent-poster')
 // and this IPC channel receives from and sends messages to the main process
 var ipcRenderer = electron.ipcRenderer
 var clipboard = electron.clipboard
+var dialog = remote.require('dialog')
 
 // For easy debugging in Developer Tools
 var state = global.state = require('./state')
@@ -208,6 +209,9 @@ function dispatch (action, ...args) {
   }
   if (action === 'toggleSelectTorrent') {
     toggleSelectTorrent(args[0] /* infoHash */)
+  }
+  if (action === 'openTorrentContextMenu') {
+    openTorrentContextMenu(args[0] /* torrentSummary */)
   }
   if (action === 'openChromecast') {
     Cast.openChromecast()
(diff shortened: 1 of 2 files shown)


renderer/index.js
```diff
@@ -795,6 +799,44 @@ function toggleSelectTorrent (infoHash) {
...
+    filters: [{ name: 'Torrents', extensions: ['torrent'] }]
+  }
+  dialog.showSaveDialog(remote.getCurrentWindow(), opts, (savePath) => {
+    var torrentFile = fs.createReadStream(torrentSummary.torrentPath)
+    var savedTorrentFile = fs.createWriteStream(savePath)
+    torrentFile.on('error', (err) => console.error('Error reading torrent file', err))
+    savedTorrentFile.on('error', (err) => console.error('Error saving torrent file', err))
+    savedTorrentFile.on('close', () => console.log('Torrent saved', savePath))
+
+    torrentFile.pipe(savedTorrentFile)
```
feross: I would just do `readFile` and `writeFile`. No need for streams.

renderer/index.js
```diff
@@ -795,6 +799,44 @@ function toggleSelectTorrent (infoHash) {
...
+  }))
+
+  menu.append(new remote.MenuItem({
+    label: 'Copy Instant.io Link to Clipboard',
+    click: () => clipboard.writeText(`https://instant.io/#${torrentSummary.infoHash}`)
+  }))
+
+  menu.append(new remote.MenuItem({
+    label: 'Copy Magnet Link to Clipboard',
+    click: () => clipboard.writeText(torrentSummary.magnetURI)
```
feross: `torrentSummary.magnetURI` is not currently a thing, but we can make it so.

feross:
Nice!! This is great. :+1: 

@Flet, added you as a contributor.

renderer/index.js
feross: Nevermind, you're right. It works, just not for the default torrents. Will fix that in a PR.

renderer/index.js
```diff
@@ -795,6 +799,44 @@ function toggleSelectTorrent (infoHash) {
...
+
+  menu.popup(remote.getCurrentWindow())
+}
+
+function saveTorrentFileAs (torrentSummary) {
+  var newFileName = `${path.parse(torrentSummary.name).name}.torrent`
+  var opts = {
+    title: 'Save Torrent File',
+    defaultPath: path.join(state.saved.downloadPath, newFileName),
+    filters: [{ name: 'Torrents', extensions: ['torrent'] }]
```
feross: Not sure how much sense a filter makes here, since they're saving a file, not selecting one. It does nothing on OS X.

feross: Actually, Windows programs seem to default to the same file type as the program. Looks good to me, then.

Will add an All Files option though, since programs seem to have those.

rom1504:
Nice :+1:

renderer/index.js
Flet: I had to add the filter in order for osx to save the file with the .torrent extension (with enforcement by the os).

This was not well documented in electron's docs :/

ungoldman:
:100:

renderer/index.js
feross: Got it. Thanks for explaining.

PR #234, opened 2016-03-27
Speed up init() by >= 2x

dcposch:
Lazy load the WebTorrent, Chromecast, and Airplay modules

image


feross:
One tweak: https://github.com/feross/webtorrent-app/pull/236

PR #237, opened 2016-03-27
Fixes for PR #233

feross:
Some tweaks to @Flet's PR.


Flet:
Thanks! Is there a reason for avoiding streams?

feross:
They're just trickier to get right because there's more events to listen to, and we're not accessing the data in a streaming fashion (i.e. using it as we read it). So, just doing it with standard callbacks is cleaner, imo.

PR #238, opened 2016-03-27
Memoize button click handlers

dcposch:
Stop virtualdom from swapping out every event handler on every update

### DIFF
--- a/renderer/index.js
+++ b/renderer/index.js
@@ -23,6 +23,8 @@ var config = require('../config')
 var TorrentPlayer = require('./lib/torrent-player')
 var torrentPoster = require('./lib/torrent-poster')
 var util = require('./util')
+var {setDispatch} = require('./lib/dispatcher')
+setDispatch(dispatch)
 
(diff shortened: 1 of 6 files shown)


renderer/index.js
```diff
@@ -264,6 +257,13 @@ function dispatch (action, ...args) {
     playPause()
   }
   if (action === 'play') {
+    state.location.go({
+      url: 'player',
+      onbeforeload: function (cb) {
+        openPlayer(args[0] /* infoHash */, args[1] /* index */, cb)
+      },
+      onbeforeunload: closePlayer
+    })
```
feross: Good catch

dcposch:
**Measuring DOM updates**
- Add a `console.log` to `node_modules/virtual-dom/vdom/patch-op.js`
- Delete your config.json and run Webtorrent

- **Before: 30 updates / second**
- **After: 0 updates / seconds**

That's because update() gets called once a second, and even if nothing has changed, `virtual-dom` won't be able to tell that two anonymous functions are equivalent, so it'll replace all the event handlers.

PR #248, opened 2016-03-27
Review and testing required!: refactor cast module

grunjol:
@dcposch @feross: In preparation for DLNA support, the 3 way if/else location===device was a little messy, so I gave it a chance and tried to refactor the cast.js file resulting in this PR.

I don't have a chromecast device and my airplay emulator software is not very good, so i need to verify everything is in order.

I think this can be good base to extend to other external devices (an external player?) and protocols (Miracast? WiDi?)

I tried to keep the interface intact i only changed the openChromecast and openAirplay with an open(<device>), only affected the UI invocation

Also, a code review is always welcome.

### DIFF
--- a/renderer/lib/cast.js
+++ b/renderer/lib/cast.js
@@ -153,7 +153,6 @@ function airplayPlayer (player) {
 function init (callback) {
   update = callback
 
-
   // Start polling Chromecast or Airplay, whenever we're connected
   setInterval(() => pollCastStatus(state), 1000)
 
--- a/renderer/lib/cast.js
+++ b/renderer/lib/cast.js
@@ -4,13 +4,14 @@ var airplay = require('airplay-js')
 var config = require('../../config')
 var state = require('../state')
 
+var debug = require('debug')('webtorrent-app:cast')
+
 // The Cast module talks to Airplay and Chromecast
 // * Modifies state when things change
 // * Starts and stops casting, provides remote video controls
 module.exports = {
   init,
-  openChromecast,
-  openAirplay,
+  open,
   stopCasting,
   playPause,
   seek,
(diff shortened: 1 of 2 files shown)


renderer/lib/cast.js
```diff
@@ -21,50 +22,102 @@ module.exports = {
...
-  // Start polling Chromecast or Airplay, whenever we're connected
-  setInterval(() => pollCastStatus(state), 1000)
-
-  // Listen for devices: Chromecast and Airplay
-  chromecasts.on('update', function (player) {
-    state.devices.chromecast = player
-    addChromecastEvents()
-  })
+// chromecast player implementation
+function chromecastPlayer (player) {
```
dcposch: Maybe `ChromecastPlayer`, and then we can say

```js
this.player = player
this.open = open
...
```

...and create one with `new ChromecastPlayer`

We might even make a `Player` and have them prototype inherit from that

dcposch:
:100:  this is a very nice clean up

dcposch:
@feross thoughts / should we do prototype inheritance here? if so should we use the ES6 `class` and `extends`?

dcposch:
@grunjol i have access to a Chromecast and an Airplay so i'll test this soon

grunjol:
My first impulse was to implement an interface and create different implementations for each device, but then i realize i was coding in javascript and opted for a more functional approach :)

renderer/lib/cast.js
grunjol: looking closely my own code there is no need (as far as i see) to expose the device's internal player in the returning function, just keep the invocation methods.

dcposch:
Haven't forgotten about this -- I'll try to merge this tomorrow

grunjol:
No problem, I am finalizing DLNA support (ETA tomorrow) based on this branch and its working fine, so the 'module interface' is working

grunjol:
@dcposch: updated the PR with the `debug` dependency removal @feross made in the project

dcposch:
Looks great.

My bad not testing this earlier. I'm in Tahoe now with no AppleTV or Chromecast, and I'll be back tomorrow.

I think my landlord's TV has Samsung AllShare so I'll try DNLA as well!

dcposch:
Just tried it out. Chromecast and Airplay both work.

:+1:

dcposch:
I *haven't* tried `setVolume` yet, but that feature isn't exposed, I think: the cast screen doesn't have volume controls.

We can test that if/when we add that feature.

dcposch:
I still think `new ChromecastPlayer()` would be a bit cleaner than having `chromecastPlayer()` return an object with function properties, but it's fine. This is definitely cleaner than what we had before and it works.

dcposch:
@grunjol thanks for doing this!

PR #250, opened 2016-03-28
add binary, fix #247

### LINKED ISSUE #247: Npm release ?
I think it would be convenient to be able to do npm install -g webtorrent-app and then just `webtorrent-app` to run it.
It could be useful on linux anyway.

rom1504:
cmd.js is mostly taken from node_modules/electron-prebuilt/cli.js (what the node_modules/.bin/electron symlink points to)

I'm not 100% sure what it will do to the package.js though. It might be needed to add some exclude for electron-prebuilt in there.

### DIFF
--- a/bin/cmd.js
+++ b/bin/cmd.js
@@ -1,0 +1,10 @@
+#!/usr/bin/env node
+
+var electron = require('electron-prebuilt')
+var proc = require('child_process')
+var path = require('path')
+
+var child = proc.spawn(electron, [path.join(__dirname, '..')], {stdio: 'inherit'})
+child.on('close', function (code) {
+  process.exit(code)
+})
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "debug": "^2.2.0",
     "drag-drop": "^2.11.0",
     "electron-localshortcut": "^0.6.0",
+    "electron-prebuilt": "0.37.2",
     "hyperx": "^2.0.2",
     "main-loop": "^3.2.0",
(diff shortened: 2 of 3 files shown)


rom1504:
this can be tested by doing `npm install rom1504/webtorrent-app#add_binary` and then `node_modules/.bin/webtorrent-app`

rom1504:
Ah electron-prebuilt is ignored by default by electron-manager so that should be ok https://github.com/electron-userland/electron-packager/issues/13#issuecomment-193634938

bin/cmd.js
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env node
+
+var electron = require('electron-prebuilt')
+var proc = require('child_process')
```
feross: We use the convention `cp` as shorthand for `child_process`.

package.json
```diff
@@ -19,6 +19,7 @@
     "debug": "^2.2.0",
     "drag-drop": "^2.11.0",
     "electron-localshortcut": "^0.6.0",
+    "electron-prebuilt": "0.37.2",
```
feross: I will double check that making this change does not cause the packaged builds to start including `electron-prebuilt` in the `node_modules` folder that we ship to users.

feross:
Ideally this would actually package the app, not just launch Electron. Because arguments are handled differently in dev vs. production mode. And on OS X, the app will show up as "Electron" in the menu bar, not "WebTorrent" because there's no Info.plist.

Will open an issue for this.

Issue #122, opened 2016-03-10
Package for Debian


nodiscc:
Hi, a Debian package for webtorrent-app would be nice. I've tried creating one with [fpm](https://github.com/jordansissel/fpm); running `fpm -s npm -t deb` inside the directory does not create a working Debian package (no binary to run from /usr/local/bin).

ngoldman:
Packaging into app form is done with the `npm run package` script. There hasn't been a formal release yet so it might be a little early to release a debian package too :)

feross:
@nodiscc We're going to publish binaries soon, including one that should work on Debian. In the meantime, you can build the app with `npm run package`.

PR #264, opened 2016-03-29
Linux: Create .deb package

grunjol:
Added support for .deb package creation in a cross platform way. 
No native dependencies nor shell commands needed.

You can select package type or (default) package in zip and deb formats

This PR implements #122


dcposch:
:+1:

dcposch:
Sweet. Sometime soon people will be able to just `apt-get install webtorrent`

grunjol:
I think package name should be webtorrent-desktop so something like
 `apt-get install webtorrent-desktop` 
that makes room to install other webtorrent components like `webtorrent-cli` or the library itself in the future

dcposch:
I read a bit about Ubuntu auto updates. Here's what I found:

* If we ever wanted everyone with Ubuntu to be able to type `apt-get install webtorrent-desktop`, we would have to get accepted by the Debian Multimedia package maintainers group.
* However, that sucks because they freeze package versions, often for *years* at a time, to cut "stable" releases: https://www.jwz.org/blog/2016/04/i-would-like-debian-to-stop-shipping-xscreensaver/
* Google does it differently. When you install Chrome on Linux, it gives you a .deb file. **The deb adds its own source to your apt sources.** See `/etc/apt/sources.list.d/google-chrome.list`. For more details try printing out ` /var/lib/apt/lists/dl.google.com_linux_chrome_deb_dists_stable_Release` and check out the rest of the stuff under `/var/lib/apt/lists/dl.google.com*`

That lets them keep their Linux users up to date on a scale of days, not years.

feross:
@dcposch Yes, this approach sounds way better than trying to get accepted by the Debian maintainers, and they'll probably add WebTorrent Desktop on their own anyway once the app is more popular and users start to request it. Let's delay that inevitable day by **offering our own apt source**.

Power users could install the apt source directly, but most users will just want to double-click the .deb file and have it do the right thing.

static/linux/postrm
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+set -e
+rm /usr/bin/webtorrent-desktop
```
feross: Should this also remove `/opt/webtorrent-desktop`? Or is that done automatically by the OS?

package.json
```diff
@@ -37,6 +37,7 @@
     "electron-packager": "^6.0.0",
     "electron-winstaller": "^2.0.5",
     "gh-release": "^2.0.3",
+    "nobin-debian-installer": "^0.0.5",
```
feross: You're on a roll with these npm packages :)

feross: Btw, don't worry about the `async` dep in this package. devDependencies aren't included in the packaged application, so it matters a lot less.

feross:
It would be nice to automatically lock WebTorrent Desktop to the launcher, after it is installed via the `.dmg` file. The VLC installer does this.

static/linux/postrm
grunjol: everything under `/opt/webtorrent-desktop` its handled by the package manager ( directory and files creation/removal)

`/usr/bin/webtorrent-desktop` is a symbolic link to `/opt/webtorrent-desktop/WebTorrent` created post-installation (check linux/postinst file) so it should be removed post-package removal (linux/postrm)

feross: Okay, that makes sense to me!

feross:
I think this is ready to be merged. Is that correct @grunjol?

grunjol:
The launcher is an unity thing, not debian in general.

grunjol:
merge it

dcposch:
@grunjol @feross i think nearly all our Linux users will be on Ubuntu, so automatically adding to the launcher on unity is probably worth it

PR #268, opened 2016-03-29
UX polish: highlight drag-drop even when placeholder isn't visible

dcposch:
Before, if you scrolled so that the bottom placeholder wasn't visible, there was no indication that the app is still a drag target.

### DIFF
--- a/renderer/index.css
+++ b/renderer/index.css
@@ -142,7 +142,6 @@ i:not(.disabled):hover {
  */
 
 .header {
-  background: rgb(40, 40, 40);
   border-bottom: 1px solid rgb(20, 20, 20);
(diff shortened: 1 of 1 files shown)


dcposch:
ezgif-2804445029

renderer/index.css
```diff
@@ -476,6 +471,17 @@ body.drag .torrent-placeholder span {
   color: #def;
 }
 
+body.drag .app::after {
+  content: ' ';
```
feross: This can be empty string, I think

feross:
LGTM!

dcposch:
ezgif-3134741536

PR #274, opened 2016-03-31
Fixes

### LINKED ISSUE #208: UI responsiveness
- [x] Disabling download of a torrent doesn't update UI immediately
- [x] Dropping a large file to seed takes *seconds* to show up in the UI.
  - We should show it right away, and not wait for the `'torrent'` event which only happens *after* the whole file has been hashed.
- [ ] Mouseover events take up to 1 second to kick in because they're not handled right away but only on `update()`. The right way to handle them is immediately, but [throttled](https://npmjs.com/package/throttleit).

dcposch:
image

* Show files immediately when seeding (fixes #208)
* Toggle show/hide in tray icon
* Save state when the app exits... for real this time
* Fix a bug where torrents would save in the "playState=requested" state, so you reopen the app and the spinner is still spinning

### DIFF
--- a/main/tray.js
+++ b/main/tray.js
@@ -6,17 +6,33 @@ var path = require('path')
 var electron = require('electron')
 var windows = require('./windows')
 
+var trayIcon
+
 function init () {
   // No tray icon on OSX
   if (process.platform === 'darwin') return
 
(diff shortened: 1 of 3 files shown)


dcposch:
@feross might want to do a retroactive code review. the state saving function in particular is a bit jank, fixes bugs for now but let me know if you want to do something better there

Issue #118, opened 2016-03-09
Add additional video player keyboard shortcuts


feross:
These are the shortcuts that I personally use a lot in VLC:

- [x] Skip forward 10 seconds (`CMD` `ALT` `right`)
- [x] Skip back 10 seconds (`CMD` `ALT` `left`)
- [x] Increase video speed (`CMD` `+`)
- [x] Decrease video speed (`CMD` `-`)

giuscri:
`CMD` or `Ctrl`? Should that depend of the platform?, `darwin` or fallback?

[Apparently](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent#Properties) the event fired when a key is down has nice properties, such as `e.ctrlKey` and `e.altKey` -- not sure how to catch the `CMD` key.

This, combined with a change in the `playbackRate` property of `<video>` should do the last two tricks.

Pretty the same for the first two. This time, via 
```
dispatch('playbackJump', state.video.currentTime + /* or -*/ 10)
```

ngoldman:
@giuscri in electron you can do `CmdOrCtrl` (which we're doing all over the place for current shortcuts). I would assume command is equivalent to control on non-mac platforms for discussion purposes.

feross:
@giuscri Any interest in sending a PR to fix this issue?

feross:
Fixed by @Lurk in #275!

PR #275, opened 2016-03-31
Add additional video player keyboard shortcuts

Lurk:
As described at #118

Skip forward 10 seconds ((CMD OR CTRL) ALT right) 
Skip back 10 seconds ((CMD OR CTRL) ALT left)
Increase video speed ((CMD OR CTRL) +)
Decrease video speed ((CMD OR CTRL) -)

Don\`t tested with airplay and chromecast
Video speed should be working with airplay but not with chromecast (https://github.com/mafintosh/chromecasts don`t support it)


renderer/index.js
```diff
@@ -99,6 +99,14 @@ function init () {
       }
     } else if (e.which === 32) { /* spacebar pauses or plays the video */
       dispatch('playPause')
+    } else if ((e.ctrlKey || e.metaKey) && e.altKey && e.which === 39) {
+      dispatch('playbackJump', state.playing.currentTime + 10)
+    } else if ((e.ctrlKey || e.metaKey) && e.altKey && e.which === 37) {
+      dispatch('playbackJump', state.playing.currentTime - 10)
+    } else if ((e.ctrlKey || e.metaKey) && (e.which === 107 || e.which === 171)) { /* CMD || CTRL + DOM_VK_ADD || DOM_VK_PLUS  */
+      dispatch('setPlaybackRate', state.playing.playbackRate + 10)
```
grunjol: **playbackRate** is the relative speed to normal so:

* 1.0 is normal speed
* 2.0 is double speed
* 0.5 is half speed (slow motion :D)

and the negative ones are for playing backwards so:
* -1.0 is backwards, normal speed
* -0.5 is backwards, half speed
* -2.0 is backwards, double speed

grunjol: The 'steps' should be implemented in base2, standard players use 1x, 2x, 4x, 8x, 16x, so the function to modify the rate should be 
* increase:
`dispatch('setPlaybackRate', state.playing.playbackRate * 2)` 
* decrease:
`dispatch('setPlaybackRate', state.playing.playbackRate / 2)`

renderer/views/app.js
```diff
@@ -22,7 +22,8 @@ function App (state, dispatch) {
     state.playing.mouseStationarySince !== 0 &&
     new Date().getTime() - state.playing.mouseStationarySince > 2000 &&
     !state.playing.isPaused &&
-    state.playing.location === 'local'
+    state.playing.location === 'local' &&
+      state.playing.playbackRate === 0
```
grunjol: `playbackRate === 0`  means playback is paused, if you `&& !state.playing.isPaused` always gives you false?

renderer/index.js
Lurk: Done, but html5 video can`t play backward.

renderer/views/app.js
Lurk: now it looks like state.playing.playbackRate === 1
When playback rate is normal we can hide player controls

feross:
Sorry for the delay in reviewing this. We'll get to this soon, I promise.

.editorconfig
```diff
@@ -0,0 +1,7 @@
+root = true
```
dcposch: Don't know if we want to add this file. There are many different editors out there, and they each have their own config file. WebTorrent Desktop is built from many npm packages, each with their own git repo.

So I don't think we want to add vim/emacs/intellij/etc configs for every editor to every repo. I recommend putting those in your home folder

.gitignore
```diff
@@ -1,2 +1,3 @@
 node_modules
 dist
+.idea
```
dcposch: Same here

renderer/index.js
```diff
@@ -96,6 +96,14 @@ function init () {
       }
     } else if (e.which === 32) { /* spacebar pauses or plays the video */
       dispatch('playPause')
+    } else if ((e.ctrlKey || e.metaKey) && e.altKey && e.which === 39) {
```
dcposch: I would make these menu items---that way users will be able to discover what the shortcuts are. See `menu.js`

renderer/index.css
```diff
@@ -674,6 +674,13 @@ body.drag .app::after {
   padding: 5px;
   margin: 0 auto;
 }
+.player-controls .rate {
```
dcposch: Looks good, but I recommend hiding the rate whenever it's 1 (ie, 99% of the time) and showing it only when it's `!== 1`

renderer/views/app.js
```diff
@@ -23,7 +23,8 @@ function App (state, dispatch) {
     state.playing.mouseStationarySince !== 0 &&
     new Date().getTime() - state.playing.mouseStationarySince > 2000 &&
     !state.playing.isPaused &&
-    state.playing.location === 'local'
+    state.playing.location === 'local' &&
+      state.playing.playbackRate === 1
```
dcposch: Indentation.

This line makes it so that the player controls *always* show when the playback rate is !== 1

In addition, I think we should hide the playback rate display when the playback rate is === 1

renderer/views/player.js
```diff
@@ -361,7 +364,12 @@ function renderPlayerControls (state) {
         />
     </div>
   `)
-
+  // render playback rate
+  if (state.playing.playbackRate !== 1) {
```
dcposch: NVM, looks good

renderer/index.js
```diff
@@ -341,7 +352,20 @@ function jumpToTime (time) {
...
-
+function setPlaybackRate (direction) {
+  var rate = state.playing.playbackRate
+  if ((direction > 0 && rate >= 1 && rate < 16) || (direction < 0 && rate > -16 && rate <= -1)) {
+    rate *= 2
+  } else if ((direction < 0 && rate > 1 && rate <= 16) || (direction > 0 && rate >= -16 && rate < -1)) {
+    rate /= 2
+  } else if (rate === -1 || rate === 1) {
+    rate *= -1
+  }
```
dcposch: Might want to make this a bit more granular.

The only times I've ever watched videos at a speed other than 1x, it was lecture videos, and I watched them at 1.25x or 1.5x.

dcposch:
Thanks for sending this! Left some comments

.editorconfig
Lurk: .editorconfig supported by many editors (http://editorconfig.org/) by default or with plugin. So maybe we should have it in repo?

.gitignore
Lurk: Ok )

Lurk:
@dcposch Agree with all your comments but one (about .editorconfig) 
I think we should discuss about it more.  Or move that discussion in new issue.
My point is that .editorconfig supported by many editors/IDE, so we can define code style one time in that file  for all contributors.

More about .editorconfig you can read here http://editorconfig.org/

grunjol:
I tend not to push IDE config files. You can always have the local excludes using 
`.git/info/exclude` 
file or globally using 
`git config --global core.excludesfile ~/.gitignore_global`

Lurk:
@grunjol .editorconfig not exactly IDE config file. It more like tab vs spaces thing config. 

And no mater what editor you use vim/emacs/intellij/atom/sublime/etc (full list: http://editorconfig.org/#download) it works.

grunjol:
is it http://standardjs.com/ compatible?

Lurk:
All done.
I remove .editorconfig from pull request
@grunjol answer to your question - yes it is :)

main/menu.js
```diff
@@ -92,6 +92,30 @@ function openSubtitles () {
...
+
+function skipBack () {
+  if (windows.main) {
+    windows.main.send('dispatch', 'skip', -1)
+  }
+}
+
+function increasePlaybackRate () {
+  if (windows.main) {
+    windows.main.send('dispatch', 'setPlaybackRate', 1)
```
dcposch: Would call this `changePlaybackRate`, otherwise it looks like we're setting the playback rate to 1 or -1

.gitignore
```diff
@@ -1,2 +1,2 @@
 node_modules
-dist
+dist
```
dcposch: `git checkout HEAD^ -- .gitignore`

renderer/lib/cast.js
```diff
@@ -344,6 +345,17 @@ function pause () {
   }
 }
 
+function setRate (rate) {
```
dcposch: I'd make this function return `true` or `false` depending on whether this cast target supports setting the playback rate.

renderer/index.js
```diff
@@ -409,7 +415,28 @@ function jumpToTime (time) {
...
+  } else if ((direction > 0 && rate >= 1 && rate < 16) || (direction < 0 && rate > -16 && rate <= -1)) {
+    rate *= 2
+  } else if ((direction < 0 && rate > 1 && rate <= 16) || (direction > 0 && rate >= -16 && rate < -1)) {
+    rate /= 2
+  } else if (direction < 0 && rate === 1) {
+    rate = -1
+  }
+  state.playing.playbackRate = rate
+  if (lazyLoadCast().isCasting()) {
+    Cast.setRate(rate)
```
dcposch: See below --- when `Cast.setRate` returns false, let's set `rate = 1`.

That way, the increase/decrease playback rate won't do anything on Chromecast (instead of showing a label that says "2X" or "0.5X" while the actual playback speed doesn't change)

renderer/index.js
```diff
@@ -409,7 +415,28 @@ function jumpToTime (time) {
...
+  } else if (direction < 0 && rate === 0.25) { /* when we set playback rate at 0 in html 5, playback hangs ;( */
+    rate = -1
+  } else if (direction > 0 && rate === -1) {
+    rate = 0.25
+  } else if ((direction > 0 && rate >= 1 && rate < 16) || (direction < 0 && rate > -16 && rate <= -1)) {
+    rate *= 2
+  } else if ((direction < 0 && rate > 1 && rate <= 16) || (direction > 0 && rate >= -16 && rate < -1)) {
+    rate /= 2
+  } else if (direction < 0 && rate === 1) {
+    rate = -1
```
dcposch: What does `rate=-1` mean?

dcposch:
LGTM otherwise. I might merge this later today, then do a few fixes. I have access to a Chromecast, Airplay, and DLNA so I can test all the cast types.

main/menu.js
Lurk: done

.gitignore
Lurk: I did it, but did not notice any effect

renderer/lib/cast.js
Lurk: done

renderer/index.js
Lurk: done

renderer/index.js
Lurk: You right. Fixed.

dcposch:
LGTM

Issue #254, opened 2016-03-28
Seeding a file should not seed from /tmp/webtorrent – should use existing location


feross:
Discussed with @dcposch.

Fix this line webtorrent/index.js:

```js
if (typeof input === 'string') opts.path = path.dirname(input)
```

PR #278, opened 2016-04-01
Seed in place, don't copy to /tmp

dcposch:
Fixes https://github.com/feross/webtorrent-desktop/issues/254

image

### DIFF
--- a/main/menu.js
+++ b/main/menu.js
@@ -117,12 +117,14 @@ function getMenuItem (label) {
 
 // Prompts the user for a file or folder, then makes a torrent out of the data
 function showCreateTorrent () {
+  // Allow only a single selection
+  // To create a multi-file torrent, the user must select a folder
   electron.dialog.showOpenDialog({
     title: 'Select a file or folder for the torrent file.',
-    properties: [ 'openFile', 'openDirectory', 'multiSelections' ]
+    properties: [ 'openFile', 'openDirectory' ]
   }, function (filenames) {
     if (!Array.isArray(filenames)) return
-    windows.main.send('dispatch', 'seed', filenames)
+    windows.main.send('dispatch', 'seed', filenames[0])
   })
 }
 
--- a/renderer/index.js
+++ b/renderer/index.js
@@ -1,7 +1,7 @@
 console.time('init')
 
 var cfg = require('application-config')('WebTorrent')
-var createTorrent = require('create-torrent')
+var defaultAnnounceList = require('create-torrent').announceList
 var dragDrop = require('drag-drop')
 var electron = require('electron')
 var EventEmitter = require('events')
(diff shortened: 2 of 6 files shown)


dcposch:
FYI @feross the way we're making it "responsive" now by changing the font size everywhere makes the modal look pretty jank when you resize the window.

I think it might look good to just keep the font sizes and list item heights constant. In other words, removing all the media queries from the CSS. Thoughts?


jank

ungoldman:
Might make sense to just take over the window rather than having a modal partially covering things up anyway.

dcposch:
IMO the cleanest look would be a new window that pops over the existing one. If the current window is big (say it covers half my screen) then replacing the whole thing with a Create Torrent modal might look pretty empty.

main/menu.js
```diff
@@ -117,12 +117,14 @@ function getMenuItem (label) {
 
 // Prompts the user for a file or folder, then makes a torrent out of the data
 function showCreateTorrent () {
+  // Allow only a single selection
+  // To create a multi-file torrent, the user must select a folder
```
feross: Why?

renderer/index.js
```diff
@@ -1,7 +1,7 @@
 console.time('init')
 
 var cfg = require('application-config')('WebTorrent')
-var createTorrent = require('create-torrent')
+var defaultAnnounceList = require('create-torrent').announceList
```
feross: This is a good change - thanks.

renderer/index.js
```diff
@@ -552,10 +563,18 @@ function stopTorrenting (infoHash) {
...
-  var torrent = lazyLoadClient().seed(files)
+  state.modal = {
+    id: 'create-torrent-modal',
+    files: files
+  }
+}
+
+// Creates a new torrent and start seeeding
+function createTorrent (options) {
+  var torrent = state.client.seed(options.files, options)
```
feross: Why doesn't this use `lazyLoadClient()` anymore?

feross:
How does this solve issue #254? To solve that, a change to `webtorrent` is required.

Also, why do we want to show the user a modal dialog anyway here? Why not just create the torrent immediately? What do we get from this?

main/menu.js
dcposch: I think it's cleaner if we want to seed in place.

renderer/index.js
dcposch: fixed

dcposch:
> How does this solve issue #254? To solve that, a change to webtorrent is required.


@feross not necessarily. Passing the right options to `client.add` makes WebTorrent seed in place without any changes in that module. You can check out this branch and try it -- you'll see it's seeding files in place now rather than copying them to `/tmp`

I added the modal because I thought users might want more control when creating a torrent. For example, they may want to set:
* The name
* Extra trackers
* Whether it's a private torrent
* etc

(Though this PR doesn't add all those options yet, the modal just shows you what torrent you're about to create. Your only input for now is OK / Cancel)

dcposch:
@feross should I get rid of the modal for now, so that it's consistent between using the Create New Torrent menu item & dragging a file or folder onto the app?

PR #279, opened 2016-04-02
Don't re-verify unchanged files

dcposch:
Keep track of the most recent file modtimes at which we verified that the torrent piece hashes all match.

Requires https://github.com/feross/webtorrent/pull/715

### DIFF
--- a/renderer/index.js
+++ b/renderer/index.js
@@ -40,6 +40,7 @@ var dialog = remote.require('dialog')
 
 // For easy debugging in Developer Tools
 var state = global.state = require('./state')
+global.debugLog = require('debug')
 
 // Force use of webtorrent trackers on all torrents
 global.WEBTORRENT_ANNOUNCE = createTorrent.announceList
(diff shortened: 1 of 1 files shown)


renderer/index.js
```diff
@@ -40,6 +40,7 @@ var dialog = remote.require('dialog')
 
 // For easy debugging in Developer Tools
 var state = global.state = require('./state')
+global.debugLog = require('debug')
```
feross: We don't depend on `debug` anymore in webtorrent-desktop, and this doesn't look like it's used in this PR. Will remove.

feross: Or, were you using this to somehow enable webtorrent's debug logs in the Electron environment? I still haven't figured out how to do that.

Will remove for now, since it's not in `package.json`, but feel free to add it back.

renderer/index.js
```diff
@@ -595,6 +597,10 @@ function addTorrentEvents (torrent) {
     var torrentSummary = getTorrentSummary(torrent.infoHash)
     torrentSummary.status = 'seeding'
     updateTorrentProgress()
+    torrent.getFileModtimes(function (err, fileModtimes) {
+      if (err) return onError(err)
+      torrentSummary.fileModtimes = fileModtimes
```
feross: We need a call to `saveState()` here

dcposch: I made sure `saveState` works on app exit.

(If we were to save `config.json` every time `state.saved` changes, we might save it a bunch of times in succession.)

feross: I couldn't find where `saveState` is called on exit. Link?

dcposch: https://github.com/feross/webtorrent-desktop/blob/78b90296c0adad35b53b83c7aa960b0e2dfcd674/main/tray.js#L49

On Windows and Linux, the only way to quit the app is from the tray icon.

This is broken on Mac, where quitting works differently. I'll add an event handler to fix it.

dcposch: Can't test on Mac, LMK if this works for you: https://github.com/feross/webtorrent-desktop/pull/278/commits/682e9275c8176ba57e09909a366a51015eae34e5

feross: Okay, fixed now.

Issue #186, opened 2016-03-21
Add DLNA/UPnP support


grunjol:
* Discovery 
 - https://www.npmjs.com/package/ssdp-js
 - https://www.npmjs.com/package/node-upnp-client
* Renderer client 
 - https://www.npmjs.com/package/upnp-mediarenderer-client

feross:
What are DLNA and UPnP actually used for? I've never seen any device that supports these.

grunjol:
UPnP is a set of protocols that allow different devices to discover each other and use the services that they can offer.

DLNA is derived from UPnP specifically for the purpose of media interoperability that includes media formats (transcoding) and system management.

I think the main difference between them is DLNA is an open specification, AirPlay is proprietary (the unofficial spec was reverse engineered) and limited to Apple ecosystem.

The model is also different, in Airplay/Chromecast you *push* the content from the source to the renderer, in DLNA case you have  Media Server | Controller | Renderer. In our case, the controller and media server are in the program and the renderer is the TV, but you can have a big storage (like a NAS) a DLNA ready  TV (most of today's SmartTVs) and a controller (ie a tablet/phone), with the controller you can browse the content of the available sources and instruct the renderer to *pull* the files from it.

About the support, I think all the big TV brands with 'smart' capabilities are DLNA certified and I am talking about [Philips/Samsung/Sony/Panasonic](http://www.dlna.org/products/)
Also Android and linux based media players do DLNA

grunjol:
I am working on a `dlnacasts` project to support DLNA with a mafintosh/chromecats compatible API.

grunjol:
@feross, @dcposch : I already implemented [dlnacasts](https://www.npmjs.com/package/dlnacasts) a mafintosh/chromecasts API compatible module for DLNA.

Just waiting PR #248 to be merged to make another PR to fix this issue. Already tested with 3 DLNA devices (2 TVs and one android media player).

feross:
Fixed in #287

PR #287, opened 2016-04-04
Add DLNA support

grunjol:
This PR adds DLNA support, fixing #186

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "application-config-path": "^0.1.0",
     "chromecasts": "^1.8.0",
     "create-torrent": "^3.22.1",
+    "dlnacasts": "^0.0.1",
     "drag-drop": "^2.11.0",
     "electron-localshortcut": "^0.6.0",
     "electron-prebuilt": "0.37.3",
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "application-config-path": "^0.1.0",
     "chromecasts": "^1.8.0",
     "create-torrent": "^3.22.1",
-    "dlnacasts": "^0.0.1",
+    "dlnacasts": "^0.0.2",
     "drag-drop": "^2.11.0",
     "electron-localshortcut": "^0.6.0",
     "electron-prebuilt": "0.37.3",
--- a/renderer/index.css
+++ b/renderer/index.css
@@ -668,8 +668,7 @@ body.drag .app::after {
   margin: 0 auto;
 }
 
-.player-controls .chromecast,
-.player-controls .airplay,
+.player-controls .device,
 .player-controls .fullscreen,
 .player-controls .back {
   display: block;
@@ -682,8 +681,7 @@ body.drag .app::after {
   float: left;
 }
 
-.player-controls .chromecast,
-.player-controls .airplay,
+.player-controls .device,
 .player-controls .fullscreen {
   float: right;
 }
@@ -692,14 +690,12 @@ body.drag .app::after {
   margin-right: 15px;
 }
 
-.player-controls .chromecast,
-.player-controls .airplay {
+.player-controls .device {
   font-size: 18px; /* make the cast icons less huge */
   margin-top: 8px !important;
 }
 
-.player-controls .chromecast.active,
-.player-controls .airplay.active {
+.player-controls .device.active {
   color: #9af;
 }
 
--- a/renderer/index.js
+++ b/renderer/index.js
@@ -246,7 +246,7 @@ function dispatch (action, ...args) {
   if (action === 'openTorrentContextMenu') {
     openTorrentContextMenu(args[0] /* infoHash */)
   }
-  if (action === 'open') {
+  if (action === 'openDevice') {
     lazyLoadCast().open(args[0] /* deviceType */)
   }
   if (action === 'stopCasting') {
--- a/renderer/index.js
+++ b/renderer/index.js
@@ -246,11 +246,8 @@ function dispatch (action, ...args) {
   if (action === 'openTorrentContextMenu') {
     openTorrentContextMenu(args[0] /* infoHash */)
   }
-  if (action === 'openChromecast') {
-    lazyLoadCast().open('chromecast')
-  }
-  if (action === 'openAirplay') {
-    lazyLoadCast().open('airplay')
+  if (action === 'open') {
+    lazyLoadCast().open(args[0] /* deviceType */)
   }
   if (action === 'stopCasting') {
     lazyLoadCast().stopCasting()
--- a/renderer/lib/cast.js
+++ b/renderer/lib/cast.js
@@ -1,5 +1,6 @@
 var chromecasts = require('chromecasts')()
 var airplay = require('airplay-js')
+var dlnacasts = require('dlnacasts')()
 
 var config = require('../../config')
 var state = require('../state')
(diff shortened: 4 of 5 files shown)


renderer/lib/cast.js
```diff
@@ -37,7 +38,7 @@ function chromecastPlayer (player) {
     var torrentSummary = state.saved.torrents.find((x) => x.infoHash === state.playing.infoHash)
     player.play(state.server.networkURL, {
       type: 'video/mp4',
-      title: config.APP_NAME + ' — ' + torrentSummary.name
+      title: config.APP_NAME + ' - ' + torrentSummary.name
```
feross: Could DLNA not handle the wide em dash character?

renderer/index.js
```diff
@@ -241,11 +241,8 @@ function dispatch (action, ...args) {
   if (action === 'openTorrentContextMenu') {
     openTorrentContextMenu(args[0] /* infoHash */)
   }
-  if (action === 'openChromecast') {
-    lazyLoadCast().open('chromecast')
-  }
-  if (action === 'openAirplay') {
-    lazyLoadCast().open('airplay')
+  if (action === 'open') {
```
feross: This name is too generic. How about `openDevice`?

renderer/views/player.js
```diff
@@ -275,13 +290,22 @@ function renderPlayerControls (state) {
...
         airplay
       </i>
     `)
   }
+  if (state.devices.dlna || isOnDlna) {
+    elements.push(hx`
+      <i.icon.device
+        class=${dlnaClass}
+        onclick=${dlnaHandler}>
+        cast
```
feross: Let's not use the same icon as Chromecast. Can you change this to `tv`?

renderer/views/player.js
```diff
@@ -181,21 +181,22 @@ function renderLoadingSpinner (state) {
...
-  if (!isChromecast && !isAirplay) throw new Error('Unimplemented cast type')
+  if (!isChromecast && !isAirplay && !isDlna) throw new Error('Unimplemented cast type')
 
   // Show a nice title image, if possible
   var style = {
     backgroundImage: cssBackgroundImagePoster(state)
   }
 
-  // Show whether we're connected to Chromecast / Airplay
+  // Show whether we're connected to Chromecast / Airplay /DLNA
```
feross: Space after the slash for consistent spacing

renderer/index.css
```diff
@@ -668,8 +668,7 @@ body.drag .app::after {
   margin: 0 auto;
 }
 
-.player-controls .chromecast,
-.player-controls .airplay,
+.player-controls .device,
```
feross: Good call, this is better than continuing to add classes for each device type.

package.json
```diff
@@ -16,6 +16,7 @@
     "application-config-path": "^0.1.0",
     "chromecasts": "^1.8.0",
     "create-torrent": "^3.22.1",
+    "dlnacasts": "^0.0.1",
```
feross: Nice work creating this!

feross:
What is a cheap DLNA device that I can buy to test this out?

feross:
Also, before we merge this, can you get rid of the `async` dependency in `dlnacasts`?

`async` is bloated and includes 50+ useless functions, plus all of `lodash`. I've never needed more than a couple async functions, just `parallel` and `series` are adequate 99% of the time. Can you swap out your `async.parallel` call with [`run-parallel`](https://www.npmjs.com/package/run-parallel) please?

Cheers!

@dcposch Can you review this too?

renderer/lib/cast.js
grunjol: its funny, for some weird reason the long dash breaks the DLNA service, i think I copied it back here to create consistency with the other call (maybe reuse it in the future?)

renderer/index.js
grunjol: good

package.json
grunjol: I didn't find any complete solution with a good API, so i made it based on chromecasts

grunjol:
Cheap DLNA device? any cheap android TV stick can handle DLNA these days. 
Even more, if to have a Samsung or Phillips 'smart tv' ia m pretty sure it supports it.
You can even install a DLNA renderer in linux (`apt-get install gmediarender`) in virtualbox

grunjol:
Will swap async with run-parallel and made all the changes you suggested

package.json
feross: Nice :)

renderer/lib/cast.js
feross: Okay, consistency is a good reason.

feross:
Cool, will test out on a Samsung Smart TV later tonight.

Will merge after you've made the changes.

grunjol:
changes updated

feross:
Okay, I just had a chance to test this out with an Xbox.

It worked for an `.mp4` file, but not a `.webm` file. Nothing we can do about that if the Xbox doesn't support .webm, I suppose.

But we need better error handling. We need to let the user know that the streaming failed and give a reason, if we can.

I'll merge this now, but let's try to improve it over the coming days and weeks.

grunjol:
I didn't catch it because in my tests with android tv sticks (2) and TVs (2), webm was supported (all of them are >2014). 

@feross: completely agree with you, we need to let user know something went wrong.
In fact, its not DLNA specific. If chromcast fail to connect, it silently fails and revert back to internal player. In airplay case, its added to state.errors
```
        state.errors.push({
          time: new Date().getTime(),
          message: 'Couldn\'t connect to Airplay'
        })
```
but not notified to the user neither.

I can catch the unsupported media error and notify the user, leading me to the question. Is there a notification system in webtorrent-desktop? I was thinking something like Android's [Toast](http://developer.android.com/guide/topics/ui/notifiers/toasts.html), or more complex, use some kind of native notification, but i think this short, program related notifications should be handled locally.

grunjol:
today i just need to read the code and not asking questions, status.error was the native solution.
Updating PR.

feross:
> I didn't catch it because in my tests with android tv sticks (2) and TVs (2), webm was supported (all of them are >2014).

It's great that most devices support `.webm` these days.

> If chromecast fail to connect, it silently fails and revert back to internal player. 

I'll fix that in a PR shortly.

The rest of your changes look good.

feross:
@dcposch On second thought, I think the changes in this PR are fine to merge before the release :)

feross:
@grunjol Thanks for this, and all your other, great PRs! I just added you as a GitHub collaborator. Congrats!

Please read the [contributing guidelines](https://github.com/feross/webtorrent-desktop/blob/master/CONTRIBUTING.md) so you know what responsibilities collaborators have. Cheers! :beers:

renderer/views/player.js
```diff
@@ -181,21 +181,22 @@ function renderLoadingSpinner (state) {
...
     backgroundImage: cssBackgroundImagePoster(state)
   }
 
-  // Show whether we're connected to Chromecast / Airplay
+  // Show whether we're connected to Chromecast / Airplay / DLNA
   var castStatus = isStarting ? 'Connecting...' : 'Connected'
   return hx`
     <div class='letterbox' style=${style}>
       <div class='cast-screen'>
         <i class='icon'>${isAirplay ? 'airplay' : 'cast'}</i>
```
dcposch: This means you can have two identical looking cast icons, one for Chromecast and one for Airplay, right?

dcposch:
@feross this broke all casting as well as seeking for me. every time `lazyLoadCasts()` runs, it throws this error:

```
module.js:338 Uncaught Error: Cannot find module 'mime'
```

feross:
No chooser yet, but I think that's fine.

Looking at the mime error.

feross:
Oh, I see what happened. `dlnacasts` doesn't have `mime` in the `package.json`, but `mime` got hoisted to the top level by `npm dedupe` for me, which is why I didn't notice it.

feross:
@dcposch Fixed on master.

grunjol:
fixed in dlnacasts 0.0.3

PR #290, opened 2016-04-04
Name the .deb file consistently

feross:
Instead of "webtorrent-desktop_0.2.0-1_amd64.deb", name it "WebTorrent-v0.2.0.deb" like the rest of the built files.

### DIFF
--- a/bin/package.js
+++ b/bin/package.js
@@ -207,8 +207,6 @@ function buildDarwin (cb) {
         verbose: true
       }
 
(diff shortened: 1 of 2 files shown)


feross:
cc @grunjol, FYI

grunjol:
Package convention
https://www.debian.org/doc/manuals/debian-faq/ch-pkg_basics.en.html

package name MUST be lowercase 

https://lintian.debian.org/tags/package-not-lowercase.html

PR #297, opened 2016-04-05
Uninstall handlers on Windows uninstall/npm run clean

feross:
Now we can call `handlers.uninstall()` to remove the .torrent file and magnet link associations. This gets called in the Windows uninstaller.

On Windows, the registry entries are removed ONLY if they still refer to WebTorrent Desktop.
On Linux, the .desktop and icon files are removed.
On OS X, there is no way to remove the association explicitly. Removing the program is sufficient.

### DIFF
--- a/bin/clean.js
+++ b/bin/clean.js
@@ -5,13 +5,18 @@
  * Useful for developers.
  */
 
-var config = require('../config')
 var os = require('os')
 var path = require('path')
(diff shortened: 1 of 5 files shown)


feross:
- [ ] test windows installer/uninstaller
- [ ] test clean script on linux

PR #298, opened 2016-04-05
Add crash reporting

feross:
This helped me to debug why Electron 0.37.4 was crashing. Turns out that 0.37.4 has a broken Buffer implementation and it seem is (accidentally) using a pre-release version of Node 6. Whoops!

In any case, crash reporting is useful. Going to leave this in.

Server-side components to the reporting are in this PR: https://github.com/feross/webtorrent-www/pull/52


dcposch:
Wow electron really has everything

https://github.com/electron/electron/blob/master/docs/api/crash-reporter.md

PR #299, opened 2016-04-05
Move WebTorrent to its own process

dcposch:


grunjol:
Just curious, why create a new electron process and not use a webworker if you are not gonna use the UI part?

grunjol:
note to self: next time, google it first
http://stackoverflow.com/questions/22774076/using-web-workers-in-stand-alone-v8-javascript-engine

dcposch:
web workers also can't do WebRTC yet

https://bugzilla.mozilla.org/show_bug.cgi?id=922363

dcposch:
webrtc is also the reason why i'm using second, hidden window instead of running webtorrent in the main process. the main process is node, while the renderer processes (windows) are chromium -- so only the renderer processes have a good solid webrtc implementation available.

grunjol:
didn't know that, really a blocking issue!

dcposch:
@feross ready for review now

main/menu.js
```diff
@@ -124,7 +124,10 @@ function showCreateTorrent () {
     properties: [ 'openFile', 'openDirectory' ]
   }, function (filenames) {
     if (!Array.isArray(filenames)) return
-    windows.main.send('dispatch', 'seed', filenames[0])
+    var options = {
+      files: filenames[0]
+    }
+    windows.main.send('dispatch', 'createTorrent', options)
```
feross: This fixes the inconsistency with the modal vs. no modal, right? Nice!

feross:
I'm looking at this now. I will merge it when I'm done reviewing.

main/windows.js
```diff
@@ -56,7 +76,7 @@ function createMainWindow () {
     icon: config.APP_ICON + '.png',
     minWidth: 375,
     minHeight: 38 + (120 * 2), // header height + 2 torrents
-    show: false, // Hide window until DOM finishes loading
+    show: true, // Hide window until DOM finishes loading
```
feross: This causes a white flash on OS X

renderer/index.js
```diff
@@ -140,41 +120,25 @@ function lazyLoadCast () {
...
-  }, 1000)
+  setInterval(update, 1000)
 }
 
 // This is the (mostly) pure function from state -> UI. Returns a virtual DOM
 // tree. Any events, such as button clicks, will turn into calls to dispatch()
 function render (state) {
-  return App(state)
+  try {
+    return App(state)
```
feross: Does this prevent main-loop from complaining when there's an exception in the view code?

renderer/index.js
```diff
@@ -432,24 +398,46 @@ function resumeTorrents () {
...
+    })
 
   cfg.write(copy, function (err) {
     if (err) console.error(err)
     ipcRenderer.send('savedState')
-    update()
   })
+
+  // Update right away, don't wait for the state to save
+  update()
```
feross: Why does this function even `update()` at all?

renderer/index.js
```diff
@@ -457,11 +445,11 @@ function onOpen (files) {
 
   // .torrent file = start downloading the torrent
   files.filter(isTorrent).forEach(function (torrentFile) {
-    addTorrent(torrentFile)
+    addTorrent(torrentFile.path)
```
feross: This will break on OS X where `onOpen` gets called from these event handlers in `main/index.js`: 

```js
app.on('open-file', onOpen)
app.on('open-url', onOpen)
```

`onOpen` also gets called with a string when launching a duplicate copy of the app.

renderer/index.js
```diff
@@ -457,11 +445,11 @@ function onOpen (files) {
...
   })
 
   // everything else = seed these files
-  showCreateTorrentModal(files.filter(isNotTorrent))
+  createTorrentFromFileObjects(files)
```
feross: Why was `filter` removed? Now even `.torrent` files will get passed in here.

feross: Oh, I see it below in `createTorrentFromFileObjects`. Moving it back here to avoid confusion.

renderer/index.js
dcposch: Yes.

Now when there's an exception in view code, it blanks out the window and prints the *actual* stack trace which made debugging a lot easier

renderer/index.js
dcposch: Ya we can probably remove that, but then we have to add `update()` calls after many of our calls to `saveState()`.

PR #302, opened 2016-04-05
Remove media queries

feross:
They make the app feel too much like a webpage. I don't like the UI
jumping around as I resize the window.


dcposch:
:+1:

PR #307, opened 2016-04-06
Show video controls immediately on mouse move

### LINKED ISSUE #208: UI responsiveness

dcposch:
Fixes #208 

@feross LMK if you think there's a cleaner way to do this

Also avoids loading all of the cast modules (chromecast/airplay/dnla) the first time you play/pause/seek a video when you're not casting.

### DIFF
--- a/renderer/index.js
+++ b/renderer/index.js
@@ -144,6 +144,7 @@ function render (state) {
 
 // Calls render() to go from state -> UI, then applies to vdom to the real DOM.
 function update () {
+  showOrHidePlayerControls()
   vdomLoop.update(state)
   updateElectron()
 }
@@ -285,7 +286,7 @@ function dispatch (action, ...args) {
   }
 
   // Update the virtual-dom, unless it's just a mouse move event
-  if (action !== 'mediaMouseMoved') {
+  if (action !== 'mediaMouseMoved' || showOrHidePlayerControls()) {
     update()
   }
 }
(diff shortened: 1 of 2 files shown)


renderer/index.js
```diff
@@ -285,7 +286,7 @@ function dispatch (action, ...args) {
   }
 
   // Update the virtual-dom, unless it's just a mouse move event
-  if (action !== 'mediaMouseMoved') {
+  if (action !== 'mediaMouseMoved' || showOrHidePlayerControls()) {
```
feross: What is this for?

renderer/index.js
```diff
@@ -324,16 +325,26 @@ function changeVolume (delta) {
   setVolume(state.playing.volume + delta)
 }
 
+// TODO: never called. Either remove or make a volume control that calls it
```
grunjol: Started working on it today, ETA tomorrow night

feross: cool!

feross:
This is fine for now, but I think we can clean this up better by handling whether to show/hide the control mostly in CSS. I'll do that later.

PR #312, opened 2016-04-06
Consistent error handling on all cast devices

feross:


feross:
Now an error is shown when the chromecast fails to connect.

PR #315, opened 2016-04-06
Make WebTorrent process easier to debug

dcposch:
Add `client` (the WebTorrent client) and `server` (the WebTorrent-to-HTTP streaming server) to the window, so it's available in Dev Tools.

### DIFF
--- a/renderer/webtorrent.js
+++ b/renderer/webtorrent.js
@@ -24,13 +24,13 @@ global.WEBTORRENT_ANNOUNCE = defaultAnnounceList
 
 // Connect to the WebTorrent and BitTorrent networks. WebTorrent Desktop is a hybrid
 // client, as explained here: https://webtorrent.io/faq
-var client = new WebTorrent()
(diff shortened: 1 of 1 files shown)


dcposch:
PS if merging plz use the squash option to it doesn't make so many merge commits

renderer/webtorrent.js
```diff
@@ -244,7 +244,6 @@ function getTorrentProgress () {
 }
 
 function startServer (infoHash, index) {
-  if (server) return
```
feross: Won't this leak a server that isn't destroyed yet?

dcposch: It was duplicated with the same check in `startServerFromReadyTorrent`

dcposch: (The way it was before could actually leak servers. If you get to `startServer` twice on torrents that aren't ready, then both will pass the `if (server)` check, wait for their respective torrents to become ready, then start two servers, leaking one of them.)

Issue #303, opened 2016-04-05
Linux: Disable Minimize to Tray on flavors that don't support it


feross:
Currently, these users can't quit WebTorrent at all short of using `kill` or rebooting (because clicking `X` only hides the window).

Is there some way we can detect support for the component that is required for the tray icon and only show it then?

grunjol:
@c0b41 could you open dev console (Shift+Ctrl+I) after app start and tell us if you see any errors?

dcposch:
* Added a workaround
* There's no way to tell whether the tray icon was actually added, as far as I can tell.
* If anyone has a Linux distribution where you have this problem and you'd like to try a better fix, I recommend adding a `try ... catch` and some `console.log`s in `tray.js`, and seeing what happens. You can also just open Dev Tools and see if there are any exceptions in the console.

PR #321, opened 2016-04-06
Add File > Quit for Linux users with broken system trays

dcposch:
Works around #303
 
image


dcposch:
This at least makes it possible for users of... unusual Linux distributions to quit the app without having to use `kill`

PR #324, opened 2016-04-06
Two more bugfixes

dcposch:
* Remove broken Developer > Reload
* Wait for the hidden webtorrent window to boot before we send it messages


dcposch:
merged

PR #330, opened 2016-04-06
Add volume controls

grunjol:
i am not a great designer but i think its not too ugly... and it works
volume-control

### DIFF
--- a/renderer/index.css
+++ b/renderer/index.css
@@ -677,6 +677,7 @@ body.drag .app::after {
 
 .player-controls .device,
 .player-controls .fullscreen,
+.player-controls .volume-icon,
 .player-controls .back {
   display: block;
   width: 20px;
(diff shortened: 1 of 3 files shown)


grunjol:
feedback is always welcome

grunjol:
Change to circle to be consistent with the position handler

image

gdelavald:
@grunjol Awesome work!
A few suggestions:
- Dragging the volume control actually drags the whole window, this feels counter intuitive and might hurt UX.
- Also it only changes when you click exactly on the volume bar, I think it would be better to increase the 'hitbox'.

grunjol:
* That's weird, I am not getting any windows dragging in any moment on volume slider, bar, or icon... is it happening to anyone else?
* You are right, the hit area is narrow, i agree with you, i think its related to the use of input[type=range], a pure graphical slider should fix that.

Also, you can control volume dragging the circle, clicking in the bar (narrow as you said) or using the wheel.

PR #331, opened 2016-04-06
No Background Mode

c0b41:
#321 after, maybe this pull request usefull or, waiting that pull request #144

### DIFF
--- a/main/index.js
+++ b/main/index.js
@@ -121,6 +121,8 @@ function processArgv (argv) {
       windows.main.send('dispatch', 'showOpenTorrentFile')
     } else if (arg === '-u') {
       windows.main.send('showOpenTorrentAddress')
+    } else if (arg === '-no-background-mode') {
+      app.noBackgroundMode = true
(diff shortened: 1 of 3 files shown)


grunjol:
@c0b41 it seems a problem with elementary OS. Check this https://github.com/electron/electron/issues/5050

c0b41:
@grunjol yeah, look like elementary os problem [GtkStatusIcon](http://elementaryos.stackexchange.com/questions/4226/how-can-i-get-applications-to-display-a-system-tray-icon), but problem still #303

feross:
I would rather fix this by detecting Tray support (when https://github.com/electron/electron/issues/5065 is fixed in Electron). Let's give that issue a week, in case it's an easy fix for Electron to make. Then we can avoiding adding a preference option for this.

But adding a Linux-specific command line flag / launcher option does not seem clean. Plus, then the user has to remember to launch the app in the correct way each time.

@c0b41 In any case, thanks for attempting a fix for this :)

PR #348, opened 2016-04-07
Exit media when user closes window

dcposch:
Before, if you were in the middle of playing a video and closed the window, the app would actually just pause the video and minimize to tray ---- so if you reopen the window a day later, the video is still there!

This fixes that.

### DIFF
--- a/renderer/index.js
+++ b/renderer/index.js
@@ -216,6 +216,16 @@ function dispatch (action, ...args) {
   if (action === 'setDimensions') {
     setDimensions(args[0] /* dimensions */)
   }
+  if (action === 'backToList') {
+    while (state.location.hasBack()) state.location.back()
+
+    // Work around virtual-dom issue: it doesn't expose its redraw function,
+    // and only redraws on requestAnimationFrame(). That means when the user
+    // closes the window (hide window / minimize to tray) and we want to pause
+    // the video, we update the vdom but it keeps playing until you reopen!
+    var mediaTag = document.querySelector('video,audio')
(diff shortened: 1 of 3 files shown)


renderer/lib/location-history.js
```diff
@@ -34,6 +34,9 @@ LocationHistory.prototype.back = function () {
   var page = this._history.pop()
 
   if (page.onbeforeunload) {
+    // TODO: this is buggy. If the user clicks back twice, then those pages
+    // may end up in _forward in the wrong order depending on which onbeforeunload
+    // call finishes first.
```
feross: Good point

renderer/index.js
```diff
@@ -216,6 +216,16 @@ function dispatch (action, ...args) {
...
   }
+  if (action === 'backToList') {
+    while (state.location.hasBack()) state.location.back()
+
+    // Work around virtual-dom issue: it doesn't expose its redraw function,
+    // and only redraws on requestAnimationFrame(). That means when the user
+    // closes the window (hide window / minimize to tray) and we want to pause
+    // the video, we update the vdom but it keeps playing until you reopen!
+    var mediaTag = document.querySelector('video,audio')
+    if (mediaTag) mediaTag.pause()
```
feross: I wonder if we still need to pause if we're going back? Won't the video tag be unloaded? Or will that still not happen because it won't happen until `requestAnimationFrame`?

dcposch: Yeah, we still need it. The render function isn't called until `requestAnimationFrame`

PR #352, opened 2016-04-07
Create error on zero-byte poster

### LINKED ISSUE #341: Bad torrent poster image selected
File from archive.org: https://archive.org/download/AphexTwinBOACAphex/AphexTwinBOACAphex_archive.torrent

The torrent poster appears to be a zero-byte .jpg image. Let's not use a zero-byte image.

screen shot 2016-04-06 at 21 40 48

AgentEpsilon:
Fixes #341 (Thanks @gdelavald!)

When a 0-byte image is retrieved from a video file to be used as a buffer, the function now creates an error. This prevents the application from producing the unexpected behavior of displaying a *black* background, as opposed to the desired blue gradient.

This happens when the video file used is composed of just audio. When the audio waveform poster generation for sound files is implemented, we may want to trigger that instead, but for now this works fine.

### DIFF
--- a/renderer/lib/torrent-poster.js
+++ b/renderer/lib/torrent-poster.js
@@ -64,7 +64,7 @@ function torrentPosterFromVideo (file, torrent, cb) {
 
       server.destroy()
 
-      if (buf.length === 0) cb(new Error('Generated poster contains no data'))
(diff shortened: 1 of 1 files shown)


feross:
Looks good to me!

Issue #335, opened 2016-04-07
Create 32-bit Linux build


dominictarr:
I get this error...
```
~/Downloads/WebTorrent-linux-x64$ ./WebTorrent 
bash: ./WebTorrent: cannot execute binary file: Exec format error
```

maybe you also need a 32 bit linux?

grunjol:
You are trying to running a 64 bit program in a 32 bit OS.

feross:
@dominictarr Does your processor not support 64-bit? Or did you just pick the 32-bit distribution?

We can make a 32-bit build if necessary.

dominictarr:
@feross I can't remember.

following this: https://www.linuxquestions.org/questions/linux-general-1/how-to-check-linux-kernel-is-32-bit-or-64-bit-612352/

```
> uname -m
i686
> uname -ar
Linux x201 4.2.0-22-generic #27-Ubuntu SMP Thu Dec 17 22:57:22 UTC 2015 i686 i686 i686 GNU/Linux
> getconf LONG_BIT
32
```
looks like I'm 32 bit

grunjol:
In 64bit platform you will see
```
uname -m
x86_64

```

feross:
@dominictarr Can you try running `npm install webtorrent-desktop -g` and then `webtorrent-desktop`?

dominictarr:
@feross yup, npm install worked.

dominictarr:
npm install is only the easy way to install if you are a node developer

because os package managers often don't keep their node versions very up to date,
so you probably need to use nvm or something, and that is a pain.

ungoldman:
It would not be too difficult to reconfigure the package script to produce both 64 and 32 bit distributions. We did it for friends. If it gets the application to a wider audience and doesn't complicate anything other than packaging I vote we Just Do It

yciabaud:
Another user on chat was stuck on 32bit OS, he did not have node installed and I think he gave up trying webtorrent desktop because of this.
A 32bit package makes sense.

feross:
@yciabaud Was he a linux user too?

oseido:
Same issue on a ubuntu 14.04 32 bit version. Can you build a 32 bit version of the WebTorrent Desktop ?

Shell Output:
-
Desktop WebTorrent: 1: WebTorrent: ELF: not found
WebTorrent: 1: WebTorrent: : not found
[...]

PR #379, opened 2016-04-11
Package 32bits and 64bits linux versions in zip and deb formats

grunjol:
This PR complements #335 and finishes #369


feross:
LGTM

feross:
What is libc6 needed for?

grunjol:
libc6 is the standard C library. electron packages all the libraries but assumes you have libc6 in your system (debian depends on that library, so its always dependant)

feross:
Sounds good.

PR #381, opened 2016-04-11
Linux build: Fix incorrect log output

feross:
Now we use a function closure to capture the `destArch` variable so the
for loop can't change it.


feross:
@grunjol - please review this :)

grunjol:
great! tested and working fine

feross:
@grunjol Thanks!

PR #394, opened 2016-04-12
Copy infoHash on right click Torrent

Lunik:
Add new menu item to directly copy the infoHash of the torrent on right click.
Requested here: https://github.com/feross/webtorrent-desktop/issues/391

### DIFF
--- a/renderer/index.js
+++ b/renderer/index.js
@@ -955,7 +955,7 @@ function openTorrentContextMenu (infoHash) {
   }))
 
   menu.append(new MenuItem({
-    label: 'Copy Torrent infoHash',
+    label: 'Copy Torrent infoHash to clipboard',
     click: () => clipboard.writeText(torrentSummary.infoHash)
   }))
 
--- a/renderer/index.js
+++ b/renderer/index.js
@@ -954,6 +954,11 @@ function openTorrentContextMenu (infoHash) {
     click: () => clipboard.writeText(torrentSummary.magnetURI)
   }))
 
+  menu.append(new MenuItem({
+    label: 'Copy Torrent infoHash',
+    click: () => clipboard.writeText(torrentSummary.infoHash)
+  }))
+
   menu.popup(remote.getCurrentWindow())
 }
 


karloluis:
That was a quick hack. Good to see you got around to it.

dcposch:
Thanks!

Are you sure we need this, though?

Most people won't know what an infohash is, and for power users, we already have a button to copy the magnet link. The magnet link contains the infohash.

Lunik:
I use only the infoHash in my webApp. It's easier for me to copy the hash.

However, this fonctionnality is maybe not at here place. How about the menu bar instead ?! (I don't know if this is possible)

But I sure this feature could be helpfull for some of us.

karloluis:
I do understand that this might not be a widely used feature. But the ability to share a Hash with close friends to share something sounds more appealing than having to create a .torrent file and sharing that instead. Keep in mind I don't know exactly how this works but it seems that it might come in handy in some edge cases. It is up to the community to say if they want to consider those.

dcposch:
@karloluis you can already send the magnet link with your friends, for example via a chat app --- you don't have to send the whole torrent file

the magnet link basically *is* the infohash. 

for example, here's an infohash for Sintel: `6a9759bffd5c0af65319979fb7832189f4f3c35d`

...and here's a valid magnet link: `magnet:?xt=urn:btih:6a9759bffd5c0af65319979fb7832189f4f3c35d`

Lunik:
Yes you'r right ! But the magnet provided by the app is way much longer because off the tracker list. 
It's not only the hash link you take as exemple.

grunjol:
@Lunik what about modify your webApp to handle magnets? i mean, you clearly know how to code and extracting the infoHash from the magnet is something you can do using a regex 
`/\burn:btih:([A-F\d]+)\b/i`
or using [parse-torrent](https://github.com/feross/parse-torrent)

feross:
I know this is a really trivial thing, and it might seem silly not to merge it. But I think we shouldn't encourage using just the info hash because it's not guaranteed to work across apps. The tracker to use is an important part of the information you need to find peers in the browser where there's no DHT. It's partially my fault for building support for that into the `webtorrent` package and instant.io long ago.

I hope you guys understand.

PR #395, opened 2016-04-12
Fix two tray icon bugs

dcposch:
* Stop video or audio when you click Hide from the tray menu
* Don't minimize to tray on Linux systems that don't support it

### DIFF
--- a/main/tray.js
+++ b/main/tray.js
@@ -1,7 +1,9 @@
 module.exports = {
-  init
+  init,
+  hasTray
 }
 
+var cp = require('child_process')
 var path = require('path')
 var electron = require('electron')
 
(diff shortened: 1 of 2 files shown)


main/tray.js
```diff
@@ -29,6 +47,13 @@ function init () {
   windows.main.on('hide', updateTrayMenu)
 }
 
+function checkLinuxTraySupport (cb) {
+  cp.exec('dpkg --get-selections libappindicator1', function (err, stdout) {
```
grunjol: dpkg is debian (and derivates) specific, not linux.This won't in other distros (red hat, fedora, etc). If you want to ensure libappindicator1 is installed you need to add it to dependencies on package creation

dcposch: Yeah I know --- if `dpkg` isn't present, then `err` will be set in the callback & it'll assume that tray support isn't there.

Adding `libappindicator1` as a dependency in the deb file sounds good to me. I added this check for the people who install from the zip file.

PR #410, opened 2016-04-14
Simplify play/pause handling

feross:
I found it awkward to listen to the video tags 'playing' and 'paused'
events, when we're controlling the state that defines what state it's
in in the first place.

This commit removes those listeners, in favor of just setting things to
the right state immediately when play(), pause(), or playPause() is
called.

Added play(), pause() methods for clarity.

### DIFF
--- a/renderer/index.js
+++ b/renderer/index.js
@@ -254,10 +254,7 @@ function dispatch (action, ...args) {
       },
       onbeforeunload: closePlayer
     })
-    playPause(false)
-  }
-  if (action === 'pause') {
-    playPause(true)
+    play()
   }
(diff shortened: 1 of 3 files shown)


feross:
@dcposch fyi, going to switch back to merge commits, because it works better with my workflow:

- make new_branch
- commit, commit, ...
- send PR
- make new_branch2, from new_branch
- commit, commit, ...
- merge new_branch PR
- send new_branch2 PR

By rebasing, the commit hash changes and the second PR contains all the commits from the first

feross:
Nevermind, I suppose I can use `git rebase -i`

PR #412, opened 2016-04-14
Only show CC icon for video

feross:


grunjol:
I know is not a common case, but consider a person learning a new language, having both audio and text  description (technically is a caption not a subtitle).

feross:
@grunjol I've never seen a caption file with an audio tag. Is that actually possible?

If we want to support this, I think the right way is to have an "Add Subtitle File..." menu item like VLC has that the user can use when the CC icon is missing.

PR #417, opened 2016-04-16
Windows Portable App

feross:
Fixes https://github.com/feross/webtorrent-desktop/issues/358

### Background

- [What is a portable app?](http://portableapps.com/about/what_is_a_portable_app)
- [Portable Audacity](http://wiki.audacityteam.org/wiki/Portable_Audacity)

### Changes

- A new `*-win.zip` output file is now produced, which just contains the electron-packaged windows app, with an added empty "Portable Settings" folder.
- At startup, when a "Portable Settings" folder is found next to the WebTorrent.exe file, then the app is in "Portable Mode" which entails:
  - Electron "userData" folder is set to the "Portable Settings" folder
  - The `config.json` file is saved in the "Portable Settings" folder
  - Torrent posters and torrent files are saved in the "Portable Settings" folder
  - The default download folder is set to the "Portable Settings" folder

Am I forgetting anything?

### PortableApps.com Platform

I [started a discussion](http://portableapps.com/node/54146) to get WebTorrent Desktop included in the PortableApps.com bundle.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,8 +57,16 @@ Where `[platform]` is `darwin`, `linux`, `win32`, or `all` (default).
 
 The following optional arguments are available:
 
-- `--package=[type]` - Package only one output file. `type` is `deb`, `dmg`, `zip`, or `all` (default)
 - `--sign` - Sign the application (OS X, Windows)
+- `--package=[type]` - Package single output type.
+   - `deb` - Debian package
+   - `zip` - Linux zip file
+   - `dmg` - OS X disk image
+   - `exe` - Windows installer
+   - `portable` - Windows portable app
+   - `all` - All platforms (default)
+
+Note: Even with the `--package` option, the auto-update files (.nupkg for Windows, *-darwin.zip for OS X) will always be produced.
 
 #### Windows build notes
 
--- a/bin/package.js
+++ b/bin/package.js
@@ -10,6 +10,7 @@ var fs = require('fs')
 var minimist = require('minimist')
 var path = require('path')
 var rimraf = require('rimraf')
+var series = require('run-series')
 
 var config = require('../config')
 var pkg = require('../package.json')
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -57,8 +57,16 @@ Where `[platform]` is `darwin`, `linux`, `win32`, or `all` (default).
 
 The following optional arguments are available:
 
-- `--package=[type]` - Package only one output file. `type` is `deb`, `dmg`, `zip`, or `all` (default)
 - `--sign` - Sign the application (OS X, Windows)
+- `--package=[type]` - Package single output type.
```
dcposch: Looks good!

bin/package.js
```diff
@@ -348,65 +346,90 @@ function buildWin32 (cb) {
...
+        console.log('Windows: Created installer.')
+        cb(null)
+      }).catch(cb)
+    }
+
+    function packagePortable (cb) {
+      // Create Windows portable app
+      console.log('Windows: Creating portable app...')
+
+      var portablePath = path.join(buildPath[0], 'Portable Settings')
```
dcposch: So this just adds an empty Portable Settings folder to the build path and then zips it. Simpler than above -- we're not making a self extracting exe. Seems elegant

bin/package.js
```diff
@@ -348,65 +346,90 @@ function buildWin32 (cb) {
...
 }
 
 function buildLinux (cb) {
-  var distPath = path.join(config.ROOT_PATH, 'dist')
-
   console.log('Linux: Packaging electron...')
   electronPackager(Object.assign({}, all, linux), function (err, buildPath) {
     if (err) return cb(err)
-    console.log('Linux: Packaged electron.')
+    console.log('Linux: Packaged electron. ' + buildPath[0])
```
dcposch: What was wrong with the Linux build? Does `packageDeb` need to finish before `packageZip` can start?

package.json
```diff
@@ -15,8 +15,7 @@
   },
   "dependencies": {
     "airplay-js": "guerrerocarlos/node-airplay-js",
-    "application-config": "^0.2.0",
-    "application-config-path": "^0.1.0",
+    "application-config": "feross/node-application-config",
```
dcposch: +1

The original was janky.

If you've already made your own module, you can also use `JSON.stringify(obj, null '\t')` to produce neatly printed configuration files. (Our current config.jsons are all on one line.)

dcposch:
LGTM

package.json
feross: Good idea.

bin/package.js
feross: The change is so that `cb` is always called. It was sometimes not called before.

bin/package.js
feross: Yep!

PR #418, opened 2016-04-16
Keep all torrent files and poster images in app config folder

### LINKED ISSUE #402: First startup: Need to copy torrents and posters to app config directory
Right now, it's impossible to remove an existing default .torrent file or poster image from the app bundle without breaking users with a config that came from an old version, because the posterURL and torrentPath literally reference a file inside the app bundle. Let's fix this.

dcposch:
Fixes #402

Also:
* Make `init()` shorter
* Delete `util.js` -- those kinds of files get messy fast
* Create `torrent-summary.js`, remove a bit of duplicated code dealing with torrentSummary objects


dcposch:
@feross doe this look reasonable?

Issue #281, opened 2016-04-02
Subtitles / Closed Captions support


grunjol:
Basic implementation

- [x] adding subtitle files manually
  - [x] via drag-drop
 - [x] via [CC] icon (Ctrl+Click)
  - [x] via menu (since drag-drop is not very discoverable) (#521)
- [x] detect subtitle files in the same folder (in our case, that's in the torrent) (#520)
- [x] allow selecting among multiple subtitle files
- [x] allow disabling subtitles
- [x] multiple charset encoding

grunjol:
Support for

* Internal: subtitles files inside torrent
* External: locally and remote subtitles files

MoOx:
If that can help, peerflix handle this.

feross:
So does Playback.

MoOx:
I can't find a way in Playback to add external subtitles, peerflix supports this.

feross:
@MoOx I think you have to drop them on the video player while it's playing.

luismanuelmansilla:
Other think is the "track synchronization" of the subtitles.. sometime is very helpfull when the subtitles don't match with the audio.

andreapaiola:
If we want to support only basic (no 3D or special effects support) .srt format the implementation is straightforward...
but there are also other formats for subtitles.

Can we declare the goal here, please?

We can load and handle zipped(?) .srt(s) local files, if you want also remote files the interface will be different with a dialog for the URL or something like that.

feross:
I think `.srt` and `.vtt` support is fine for now. These should be straightforward to support, with the help of the html5 video tag.

We don't need to support zipped files or a remote interface. I think the goal here should be to support the basic set of features that VLC supports:

- [ ] adding subtitle files manually
  - [ ] via drag-drop
  - [ ] via menu (since drag-drop is not very discoverable)
- [ ] detect subtitle files in the same folder (in our case, that's in the torrent)
- [ ] allow selecting among multiple subtitle files
- [ ] allow disabling subtitles

That's probably enough.

feross:
PR sent for basic subtitle support via drag-n-drop: https://github.com/feross/webtorrent-desktop/pull/361

danielhf:
Using only webtorrent in browser. Is there a way to add subtitles to the video without download anything? I mean, only using <script src="webtorrent.min.js"></script>

rom1504:
@danielhf try https://andreapaiola.name/niagara/

danielhf:
[...]

PR #419, opened 2016-04-16
Multiple srt charset encoding support

grunjol:
I was working with srt-to-vtt and to-utf-8 projects to implement proper non-UTF-8 support for srt files.

This handles another item in #281


feross:
These types of commits are not necessary because of the way that semver works. ^1.0.3 will install the newest 1.x.x release.

grunjol:
i realized it after the fact. Thanks for the advise anyway.

PR #423, opened 2016-04-18
OS X: Don't stop music when tabbing to another program

feross:


ungoldman:
https://www.youtube.com/watch?v=SPlQpGeTbIE

Issue #62, opened 2016-03-07
Preferences Window


ChrisMorrisOrg:
After these issues have been opened: #5, #50, #53. Maybe we should consider putting these features into preferences pane?

feross:
Agree that we'll need preferences at some point, but we want to be judicious about what we add and not spew too much complexity too quickly.

I think a preference for #53 make the most sense to do first.

ChrisMorrisOrg:
100% agree. Maybe we should discuss here what the primary customisation points will be, when the project matures enough.

feross:
@ChrisMorrisOrg Want to work on an initial preferences window?

I don't think there should be any UI in the window, but just a preferences menu item like most apps have.

ngoldman:
@feross you mean like a `File > Preferences >` type thing?

feross:
Like this:

<img width="229" alt="screen shot 2016-03-07 at 2 30 30 pm" src="https://cloud.githubusercontent.com/assets/121766/13585548/2ea29214-e471-11e5-8774-fb257ab90830.png">

ChrisMorrisOrg:
Sure thing! I'll work on it tonight unless someone else has already started work on it.

ngoldman:
Not me! I say go for it @ChrisMorrisOrg

ChrisMorrisOrg:
If nobody has started working on this. I will claim it now (sorry for the delay).

karloluis:
Can we get a list of things we know for sure should be on the settings page. All the suggestions seem to be scattered out different tickets.

grunjol:
Now the subtitle part is almost done I think this should be number 1 priority to do.
I will start with the basic structure based on #144 and will go from there to get a modular but simple solution

PS: assigned the issue to myself, will create a branch and periodically will ask for suggestions/reviews

grunjol:
Questions to @feross @dcposch all:
* Should this preference be inside its own window or a 'dialog like' window?
* hardcoded preference items (big file with code and definition items) or files based( modular/pluggable config?) defined items?

feross:
@grunjol I want to move away from the modal approach we have been using. We should use the navigation buttons (back/forward) for this kind of thing. So, I would implement this as a view in the renderer. This actually simplifies things a lot because we can avoid IPC through the main process -- the `state` object already lives in the renderer. We can also avoid managing yet another window in `windows.js`.

When `state.url` is set to `'preferences'`, that view should show up. To start, let's just have one option:

[...]

PR #426, opened 2016-04-19
Add Preferences Support

grunjol:
I took what @ChrisMorrisOrg started, get some feedback from #62 and this is what you get.

Its not as clean as i want, but is what i get.

Feel free to comment.

The whole idea is to implement a preference section with reusable components (file selector, text, checkbox) and pass parameters to them.

The only active right now is Download Path (help text needs rewording)

image

Also, commented is a checkbox preference (Tray disable) which is not working, but you can comment out and see how the pref is saved.

image

You can also see how a section will be displayed

@dcposch @feross feel free to use the branch to make the change you want.

### DIFF
--- a/main/menu.js
+++ b/main/menu.js
@@ -116,6 +116,11 @@ function decreasePlaybackRate () {
   }
 }
 
+// Open the preferences window
+function showPreferences () {
+  windows.main.send('dispatch', 'preferences')
+}
+
 function onWindowShow () {
   log('onWindowShow')
   getMenuItem('Full Screen').enabled = true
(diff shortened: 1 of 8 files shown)


dcposch:
k, i'm merging this today and we can fix it on master

@grunjol @ChrisMorrisOrg thanks for the work!

ChrisMorrisOrg:
Nice!!

grunjol:
Nice to see this merged. 
As you said, we have a lot of work to do now. First step, define and homogenize styles.

PR #427, opened 2016-04-20
Center video on current monitor

### LINKED ISSUE #404: Pressing play on downloaded torrent resets Window position.
0.3.3
Windows 10 latest.
**What did you do?**
Press play on downloaded video
Plays video in window, with the window not moving
**What actually happened?**
Plays video correctly, but resets window screen position.

dcposch:
Fixes #404

### DIFF
--- a/main/ipc.js
+++ b/main/ipc.js
@@ -140,6 +140,14 @@ function setBounds (bounds, maximize) {
   // Assuming we're not maximized or maximizing, set the window size
   if (!willBeMaximized) {
     log('setBounds: setting bounds to ' + JSON.stringify(bounds))
+    if (bounds.x === null && bounds.y === null) {
+      // X and Y not specified? By default, center on current screen
+      var screen = require('screen')
+      var scr = screen.getDisplayMatching(windows.main.getBounds())
+      bounds.x = Math.round(scr.bounds.x + scr.bounds.width / 2 - bounds.width / 2)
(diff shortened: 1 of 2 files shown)


dcposch:
@feross want to check that this actually works on multiple monitors? Otherwise I'll check later. Either way we shouldn't merge this until we're sure

main/ipc.js
```diff
@@ -140,6 +140,14 @@ function setBounds (bounds, maximize) {
   // Assuming we're not maximized or maximizing, set the window size
   if (!willBeMaximized) {
     log('setBounds: setting bounds to ' + JSON.stringify(bounds))
+    if (bounds.x === null && bounds.y === null) {
+      // X and Y not specified? By default, center on current screen
+      var screen = require('screen')
```
feross: Should be `electron.screen`. Requiring electron modules is deprecated.

feross: I fixed this post-merge

feross:
I think it's a little weird that half of the object is calculated on the client and the other in the main process, but we can clean that up later.

dcposch:
The client doesn't know which screen it's on. (We could use `remote` but that does synchronous IPC and is a bit jank.) Conversely the main process doesn't know what size the video is.

feross:
We could have the client just send the size of the video to the main window and let it figure it out. But nevermind. This is not important right now.

Issue #389, opened 2016-04-11
Create Torrent options


Edward-Stryfe:
This is the only thing I have issue with as far as this client goes. It is very well made for a 0.3 beta and functions in all the right ways. It's a little disappointing that for a torrent to work it NEEDS to have DHT or PEX on. But that's not the point it still seeds on regular BitTorrent and the new WebTorrent sides of the spectrum which is amazing.

But when I go to create a torrent it only allows me to use the directory when creating a torrent but that's a beta disadvantage I'm sure the choice was one or the other and directories are better in that case. But I would love to see tracker editing and webseed adding options. Because at this point in time I'm fine with having to make the torrents in a regular client but I can't find one that will support wss:// links. It errors out. Of course I can make a magnet link and add all that. But then there is no torrent I can use for it. Which is a major downside.

I want to setup my website around Web/BitTorrent and thanks to this client it's been made a LOT easier so thanks and kudos to you Feross in that regard. But there's no way to even change the default list of trackers added to torrents. Or webseeds so there's that too.

feross:
This is done now, thanks to @dcposch

PR #431, opened 2016-04-21
Create Torrent page

dcposch:
Done:
* Handle drag-and-drop (gives you an array of W3C File objects) and the Create Torrent menu item (gives you a path to a folder)
* Show the user which torrent they're about to create

Left to do:
- [x] Change from modal to page
- [x] Add options.
  - We can't have a name option, for now. WebTorrent requires that to be the name of the file or folder we're seeding from
  - [x] We should have an option to change the announce list
  - [x] Private bit
  - [x] Comment
  - Order of files (not doing this for now -- doesn't seem important and it's a decent amount of work, doing it right would require a drag-drop ordered list control)


feross:
Fixes #389

dcposch:
### New look

** Basic **

image

** Advanced **

image

dcposch:
I think this is ready to go now.

**Caveats**
* The comment field doesn't work yet. See https://github.com/feross/create-torrent/issues/36
* You can add and remove trackers, but removing the built-in WebTorrent trackers (the ones that start with `wss://`) doesn't work. They get re-added automatically by `create-torrent` because we set the `WEBTORRENT_ANNOUNCE` global variable. However, we can't just remove that variable because we want `webtorrent` to use those trackers for existing torrents.
  I'm thinking we should add a different mechanism to tell `webtorrent` which trackers to always use. Global variables are gross anyway.

dcposch:
@feross might want to review

grunjol:
Nice work on this. My only suggestion (and i am not a designer) is about the style. I faced the same problem when i was doing the settings page (PR #426). 
There is no definition on styling in the app (except for the gorgeous torrent list and player).
I saw you started using some kind of 'material design' but with a light theme, when the app clearly use a dark one. Now i see you go with the dark theme approach but i found the palette kind of high contrast B/W.

What we need to define:

* We go with the 'Material Design' style 
* An 'Atom-alike' one (like you can see in the screenshots of https://github.com/feross/webtorrent-desktop/pull/426)
* A new one, combining styles

If we go with the material design we need to follow it. A couple of days ago google released an example of material design for desktop, using the [Chrome Browser as an example](http://www.theverge.com/2016/4/23/11495442/chrome-os-material-design-update-rolling-out-screenshots). 

To be clear,it is not my intention to hijack your PR, nor criticizes your great work, just an attempt to define this thing to allow more people to get involved.

dcposch:
@grunjol I agree. We have a lot of colors, margins, font sizes, and so on that could be cleaned up and made consistent across the app.

* I think it would be sweet to get a review from a designer. They could tell us what specifically needs to be fixed.
* Then, we could go in and try to refactor and simplify `index.css`

I think we can do that in a future PR tho

feross:
👍

PR #434, opened 2016-04-21
Detect files with uppercase extensions

grunjol:


feross:
Thanks @grunjol - looks good.

PR #452, opened 2016-04-30
Use airplayer

watson:
This PR is currently WIP and meant as a way of tracking the progress of improving the AirPlay support.

This PR uses the [airplayer](https://github.com/watson/airplayer) module instead of [airplay-js](https://github.com/guerrerocarlos/node-airplay-js).

As far as I can see, the main feature differences between the two modules currently are:

- airplay-js serves the video using HLS, airplayer just streams the video using HTTPS range requests
- airplay-js supports using ffmpeg (if present on the OS) to transpile and cut up the video into smaller pieces to be served as an HLS playlist. Airplayer does not use ffmpeg, so does not support transpiring, but will therefore use significant less CPU for videos that does not need to be transpiled
- airplay-js supports subtitles if ffmpeg is used
- airplayer does not write anything to stdout or stderr

Thoughts and feedback much appreciated 😃 

## Todo

- [x] ~~support~~ mock volume
- [x] update `status` function to new API
- [x] test this on a Apple TV pre 4th gen
- [x] test this on a 4th gen Apple TV
- [x] subscribe to `events`

### DIFF
--- a/renderer/lib/cast.js
+++ b/renderer/lib/cast.js
@@ -12,7 +12,7 @@ module.exports = {
   setRate
 }
 
-var airplay = require('airplay-js')
+var airplayer = require('airplayer')()
 var chromecasts = require('chromecasts')()
 var dlnacasts = require('dlnacasts')()
 
@@ -41,10 +41,9 @@ function init (appState, callback) {
     state.devices.dlna = dlnaPlayer(player)
   })
 
-  var browser = airplay.createBrowser()
-  browser.on('deviceOn', function (player) {
+  airplayer.on('update', function (player) {
     state.devices.airplay = airplayPlayer(player)
-  }).start()
+  })
 }
 
 // chromecast player implementation
@@ -130,12 +129,12 @@ function chromecastPlayer (player) {
 // airplay player implementation
 function airplayPlayer (player) {
   function open () {
-    player.play(state.server.networkURL, 0, function (res) {
-      if (res.statusCode !== 200) {
+    player.play(state.server.networkURL, function (err, res) {
+      if (err) {
         state.playing.location = 'local'
         state.errors.push({
           time: new Date().getTime(),
-          message: 'Could not connect to AirPlay.'
+          message: 'Could not connect to AirPlay. ' + err.message
         })
       } else {
         state.playing.location = 'airplay'
@@ -145,11 +144,11 @@ function airplayPlayer (player) {
   }
 
   function play (callback) {
-    player.rate(1, callback)
+    player.resume(callback)
   }
 
   function pause (callback) {
-    player.rate(0, callback)
+    player.pause(callback)
   }
 
   function stop (callback) {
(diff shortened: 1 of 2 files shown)


watson:
@feross I didn't know about the volume feature of the airplay protocol. I think it's something only found in the new edition of the Apple TV, but I have to check when I get home. I guess you just allow the user to control the volume no matter what but ignore the 400 http response code from the apple tv in case volume is not supported?

watson:
@feross I just tested the volume thing with the 4th gen Apple TV and it doesn't support it as far as I can see. I get a 400, so I don't know where it comes from. I'll investigate further...

watson:
@feross I haven't found any reference to the volume api anywhere besides the airplay-js client (which got it from the airplay2 module). According to `git blame`, it was part of the first commit with no special mention. All Apple TV's that I've tried it on responds with a 400 status code. I opt for dropping it until someone complains and shows me where it's used.

feross:
@watson Okay, fair enough.

watson:
@feross the new airplayer module emits an event when the user pauses, stops etc. the playback on the Apple TV. I can see that the `chromecastPlayer` and the `dlnaPlayer` uses an `addEvents` function to subscribe to events. Is there a way I can change the `state` object to reflect the current playback state?

feross:
@watson Can you try setting `state.playing.isPaused`?

watson:
@feross airplay have 4 playback states "loading", "playing", "paused" and "stopped".

I'm not sure if "loading" is only used in the initial phase between the request to play have been sent to the Apple TV and it have actually started playing. It might also be used if the Apple TV starts buffering mid-way (I'll check that).

Does webtorrent-desktop have the concept of "loading" or "buffering" and does it differentiate between "paused" and "stopped"?

feross:
@watson Right now, I believe we don't distinguish between all those states. @dcposch can you confirm?

Eventually, we probably should since every video player project I've worked on has always needed all these states to handle edge cases. But we apparently haven't needed them yet.

dcposch:
@feross @watson we detect if the video is stalled, and when that happens we show a spinner and torrent download progress when that happens

we don't distinguish between `loading` and `buffering`

we don't distinguish between `paused` and `stopped`

watson:
@dcposch if I'm told by the Apple TV that the video is buffering, do I need to set a state somewhere accordingly?

dcposch:
@watson you could dispatch a `mediaStalled` event

```javascript
var {dispatch, dispatcher} = require('../lib/dispatcher')
...
dispatch('mediaStalled')
```

--

Right now, we actually never show a loading spinner while we're casting, though.

I don't know if we *need* a loading spinner: the cast device might show its own loading spinner on the big screen.

watson:
@dcposch I've added basic event support now (paused/playing). How do you feel about leaving it here for now and then add more fine-grained support later once/if that is supported by webtorrent-desktop?

dcposch:
LGTM

renderer/lib/cast.js
```diff
@@ -129,13 +128,31 @@ function chromecastPlayer (player) {
...
   function open () {
-    player.play(state.server.networkURL, 0, function (res) {
-      if (res.statusCode !== 200) {
+    player.play(state.server.networkURL, function (err, res) {
+      if (err) {
         state.playing.location = 'local'
         state.errors.push({
           time: new Date().getTime(),
-          message: 'Could not connect to AirPlay.'
+          message: 'Could not connect to AirPlay. ' + err.message
```
dcposch: This might produce ugly error messages, but that's fine, better than just seeing 'Could not connect to Airplay' w/ no indication why. Later if we see what the common errors are we can replace them w friendly looking ones.

renderer/lib/cast.js
```diff
@@ -172,12 +194,13 @@ function airplayPlayer (player) {
   }
 
   function volume (volume, callback) {
-    // TODO remove line below once we can fetch the information in status update
+    // AirPlay doesn't support volume
+    // TODO: We should just disable the volume slider
```
dcposch: Sounds good, we can do that in a future PR

feross:
@watson Thanks for making `airplayer` and sending this PR! This was just released in WebTorrent v0.7.0.

I just added you as a GitHub collaborator on this repo. Thanks for helping out!

Issue #451, opened 2016-04-30
path-exists module not in dependencies


watson:
**What version of WebTorrent Desktop?** (See the 'About WebTorrent' menu)
master branch

**What operating system and version?**
OSX 10.10.5

**What did you do?**
```
$ git@github.com:watson/webtorrent-desktop.git
$ cd webtorrent-desktop
$ npm i
$ npm start
```

**What did you expect to happen?**
Unicorns and rainbows

**What actually happened?**
![screen shot 2016-05-01 at 00 18 16](https://cloud.githubusercontent.com/assets/10602/14938871/bcbbfc34-0f32-11e6-8d11-79198965dca3.png)

After I manually installed the path-exists module everything worked fine. It's currently being required in both config.js and bin/clean.js

dcposch:
Wow, that sucks.

* I've added `path-exists` as a dependency
* I'll try to add CI so that the build can't be green if dependencies are missing

dcposch:
Wow, that sucks. Should be fixed if you update your fork though.

* Fixed in https://github.com/feross/webtorrent-desktop/commit/dd27a32b60c7eea68436bdcf84b8e164bafddde7
* I'll try to add CI so that it can't pass if there's a missing dependency

PR #455, opened 2016-05-03
Check for missing or unused dependencies

dcposch:
Makes sure something like #451 can't happen again


feross:
I think this is a little overkill. There wasn't really any chance that we could have shipped a broken build to users. We should just be more careful, IMO.

feross:
There's also probably overlap with this existing package: https://www.npmjs.com/package/dependency-check

PR #462, opened 2016-05-07
Idea: Set a peerid prefix to track webtorrent desktop clients

yciabaud:
I believe it could be interesting to set a custom peerid to identify the client used in a peer list.

Maybe we can set a character at the end of the prefix to match the client used
- `-WW0000-` or `-WW0000H` for node client (hybrid)
- `-WW0000D` for desktop client
- `-WW0000B` for browser client

I tried it in this PR but its not working for some reason. 

Do you think it can be useful to differentiate clients in the peer id?


feross:
This is an interesting idea, and might be worthwhile.

But your approach won't work because there are only 4 digits reserved for the version. We'll need to think more about how to accomplish this.

More info in this BEP here: http://www.bittorrent.org/beps/bep_0020.html

yciabaud:
Yes, I was trying to copy the model of BitTyrant on this page:
> BitTyrant is an Azureus fork and simply uses AZ2500BT + random bytes as peer ID in its 1.1 version.

That's why I kept WW for the prefix and added D after the version. 
How would you handle this?

dcposch:
@yciabaud thanks. If we do this, we'll use WD- as the prefix for WebTorrent Desktop and keep WW- as the prefix for WebTorrent on the web.

yciabaud:
Sure WD- works for me, do you want a PR in this way?

feross:
@yciabaud - I can take care of sending a PR!

I just sent one to `bittorrent-peerid` to support detecting WebTorrent Desktop: https://github.com/fisch0920/bittorrent-peerid/pull/11

yciabaud:
Great, we will have to update the test in bittorrent-tracker to reflect the version string once it is merged: https://github.com/feross/bittorrent-tracker/blob/master/test/stats.js#L148

feross:
This is done now, in PR https://github.com/feross/webtorrent-desktop/pull/742

PR #463, opened 2016-05-07
[WIP] Add "Open folder" context menu

watson:
Add "Go to folder" menu item in the torrent context menu to open the folder containing the torrent:

screen shot 2016-05-07 at 21 28 43

**Questions:**
- [ ] I could rewrite this to use `ipcRenderer.send('showItemInFolder', folderPath)` if you like?
- [ ] Is "Go to folder" the best title? Please let me know if I should change it or if the casing should be different

**To do:**
- [ ] If the torrent is just a single file, open the OS finder window and select the file
- [ ] If the torrent contains multiple files, open the enclosing folder using the OS finder window
- [ ] Only show context menu, if the torrent have been created on the filesystem

**Nice to have:**
- [ ] If the torrent contains multiple files, allow the user to right click an individual file and have the OS finder window highlight that particular file (I'm not sure if this is possible without major refactoring)

### DIFF
--- a/renderer/index.js
+++ b/renderer/index.js
@@ -38,6 +38,7 @@ var dialog = electron.remote.dialog
 var Menu = electron.remote.Menu
 var MenuItem = electron.remote.MenuItem
 var remote = electron.remote
+var shell = electron.shell
 
 // This dependency is the slowest-loading, so we lazy load it
 var Cast = null
@@ -1002,6 +1003,11 @@ function openTorrentContextMenu (infoHash) {
   var torrentSummary = getTorrentSummary(infoHash)
   var menu = new Menu()
   menu.append(new MenuItem({
+    label: 'Go to folder',
+    click: () => openDownloadFolder(torrentSummary)
+  }))
+
+  menu.append(new MenuItem({
     label: 'Save Torrent File As...',
     click: () => saveTorrentFileAs(torrentSummary)
   }))
@@ -1017,6 +1023,12 @@ function openTorrentContextMenu (infoHash) {
   }))
 
   menu.popup(remote.getCurrentWindow())
+}
+
+function openDownloadFolder (torrentSummary) {
+  var folder = path.parse(torrentSummary.name).name
+  var fullPath = path.join(torrentSummary.path, folder)
+  shell.showItemInFolder(fullPath)
 }
 
 function saveTorrentFileAs (torrentSummary) {


watson:
@feross is there an easy way to know which of the files in the expanded torrent is selected? I haven't been able to figure it out

renderer/index.js
```diff
@@ -1019,6 +1025,12 @@ function openTorrentContextMenu (infoHash) {
   menu.popup(remote.getCurrentWindow())
 }
 
+function openDownloadFolder (torrentSummary) {
+  var folder = path.parse(torrentSummary.name).name
+  var fullPath = path.join(torrentSummary.path, folder)
```
watson: I've found that this approach doesn't work for all torrents. There must be a better way to find the path for the download location. Any pointers greatly appreciated 😃

feross: There's actually already a function called `showFolder` that implements this logic. Let's just use that.

feross:
> I could rewrite this to use ipcRenderer.send('showItemInFolder', folderPath) if you like?

There's already a `ipcMain.on('openItem')` which we could re-use, but `showItemInFolder` is actually a bit  different and it's actually the behavior we want.

> Is "Go to folder" the best title? Please let me know if I should change it or if the casing should be different

On OS X, the typical phrasing is "Show in Finder". For other OSes let's just go with "Show in Folder".

>  If the torrent is just a single file, open the OS finder window and select the file
>  If the torrent contains multiple files, open the enclosing folder using the OS finder window

These should be taken care of by the `showItemInFolder` electron API.

>  If the torrent contains multiple files, allow the user to right click an individual file and have the OS finder window highlight that particular file (I'm not sure if this is possible without major refactoring)

Let's do this in a future PR.

feross:
I suppose the bigger question about this PR is should we have duplicate functionality in the context menu and the UI, since there's already an "Open Folder" item in the UI:

screen shot 2016-05-09 at 01 15 08

I think it's actually fine to put this functionality in the context menu instead of the normal UI since:

1. Clicking it makes you leave WebTorrent
2. I expect it to be in the right click menu -- probably from experience with other clients
3. 80% of users probably won't need to use this

feross:
I'm going to work on a PR based on what you've started here.

feross:
@watson Thanks for the PR! Closed in favor of https://github.com/feross/webtorrent-desktop/pull/469

Issue #464, opened 2016-05-07
Cannot create torrent from file


furstenheim:
**What version of WebTorrent Desktop?** (See the 'About WebTorrent' menu)
Version 0.3.3
**What operating system and version?**
Linux, Ubuntu 14.04
**What did you do?**
In the menu I selected "Create New Torrent..."
**What did you expect to happen?**
To be able to add either a file or a folder
**What actually happened?**
I could only add a folder.

This is an issue with [electron](https://github.com/electron/electron/blob/master/docs/api/dialog.md) (or maybe Linux), a dialog cannot be both "openFile" and "openDirectory": 

> Note: On Windows and Linux an open dialog can not be both a file selector and a directory selector, so if you set properties to ['openFile', 'openDirectory'] on these platforms, a directory selector will be shown.

feross:
Looks like Atom solved this by offering two menu options on Linux and Windows.

https://github.com/atom/atom/pull/1856

feross:
Fixed in #465 by @furstenheim

PR #465, opened 2016-05-07
Allow to torrent a single file

furstenheim:
#464

### DIFF
--- a/main/menu.js
+++ b/main/menu.js
@@ -108,6 +108,18 @@ function getMenuItem (label) {
   }
 }
 
+// Prompts the user for a file, then makes a torrent out of the data
+function showOpenSeedFile () {
+  electron.dialog.showOpenDialog({
+    title: 'Select a file for the torrent file',
+    properties: ['openFile']
+  }, function (filenames) {
+    if (!Array.isArray(filenames)) return
+    var file = filenames[0]
+    windows.main.send('dispatch', 'showCreateTorrent', file)
+  })
+}
+
 // Prompts the user for a file or folder, then makes a torrent out of the data
 function showOpenSeedFiles () {
   // Allow only a single selection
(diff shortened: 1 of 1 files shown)


main/menu.js
```diff
@@ -144,7 +156,12 @@ function showOpenTorrentAddress () {
 function getAppMenuTemplate () {
   var fileMenu = [
     {
-      label: 'Create New Torrent...',
+      label: 'Create New Torrent from File',
```
feross: Should be 'Create New Torrent from File...'

main/menu.js
```diff
@@ -144,7 +156,12 @@ function showOpenTorrentAddress () {
...
+      accelerator: 'CmdOrCtrl+F',
```
feross: No shortcut needed for this one.

main/menu.js
```diff
@@ -144,7 +156,12 @@ function showOpenTorrentAddress () {
 function getAppMenuTemplate () {
   var fileMenu = [
     {
-      label: 'Create New Torrent...',
```
feross: This menu item should only be added on non OS X platforms.

main/menu.js
```diff
@@ -144,7 +156,12 @@ function showOpenTorrentAddress () {
...
+      label: 'Create New Torrent from File',
+      accelerator: 'CmdOrCtrl+F',
+      click: showOpenSeedFile
+    },
+    {
+      label: 'Create New Torrent from Folder',
```
feross: Need a conditional here. Should still be 'Create New Torrent...' on OS X.

Add '...' at the end for non OS X.

feross:
Mostly looks good. See inline comments.

furstenheim:
@feross I've addressed your comments in the last commit. I do not have OS X to check it there, but it should be alright.

feross:
Merged -- nice work!

PR #479, opened 2016-05-11
Fix torrentPath migration

### LINKED ISSUE #448: Test migration code
Getting fs.read errors on Windows because of buggy migration code. The `torrentPath` is `..\\static\\bigBuckBunny.torrent` on Windows, not `bigBuckBunny.torrent` as expected. 

Before shipping v0.4, we need to thoroughly test upgrading from v0.3.3.

dcposch:
Fixes #448 

**We should test this on Windows before merging.**

**Test plan:**
1. Delete config.json
2. Install the current released version
4. Run, add a torrent, then quit
5. Run from this branch. Both default torrents and the new one you added should still work
6. Repeat 1-5 on Windows and on posix (Mac or Linux)

### DIFF
--- a/renderer/index.js
+++ b/renderer/index.js
@@ -126,10 +126,19 @@ function cleanUpConfig () {
     // Migration: replace torrentPath with torrentFileName
     var src, dst
     if (ts.torrentPath) {
+      // There are a number of cases to handle here:
+      // * Originally we used absolute paths
+      // * Then, relative paths for the default torrents, eg '../static/sintel.torrent'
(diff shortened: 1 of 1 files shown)


feross:
Nice! Thanks for figuring out what cases we needed to handle. Was on a plane when you sent this PR. Will check it out and test now.

dcposch:
Tested on 64-bit Windows 7. Installed Webtorrent 0.1.0, added a torrent, installed 0.3.3, added another torrent, then ran from this branch. The config file was migrated correctly.

feross:
LGTM

PR #481, opened 2016-05-11
Allow selecting individual files to torrent

### LINKED ISSUE #360: Ability to select/deselect files
Didn't find an issue (maybe I missed it?)

It'd be a nice feature to have.

dcposch:
Saves bandwidth and disk space when a torrent contains extra files you don't need

Fixes #360 

image

### DIFF
--- a/renderer/index.css
+++ b/renderer/index.css
@@ -113,6 +113,10 @@ table {
   opacity: 0.3;
 }
 
+.clickable {
+  cursor: pointer;
+}
+
 .float-right {
   float: right;
 }
@@ -597,7 +601,7 @@ body.drag .app::after {
 }
 
 .torrent-details {
-  padding: 8em 20px 20px 20px;
+  padding: 8em 12px 20px 20px;
 }
 
 .torrent-details table {
@@ -611,6 +615,10 @@ body.drag .app::after {
   height: 28px;
 }
 
+.torrent-details td {
+  vertical-align: center;
+}
+
 .torrent-details tr:hover {
   background-color: rgba(200, 200, 200, 0.3);
 }
@@ -621,14 +629,14 @@ body.drag .app::after {
   vertical-align: bottom;
 }
 
-.torrent-details td.col-icon {
-  width: 2em;
-}
-
-.torrent-details td.col-icon .icon {
+.torrent-details td .icon {
   font-size: 18px;
   position: relative;
   top: 3px;
+}
+
+.torrent-details td.col-icon {
+  width: 2em;
 }
 
 .torrent-details td.col-name {
@@ -643,6 +651,11 @@ body.drag .app::after {
 
 .torrent-details td.col-size {
   width: 4em;
+  text-align: right;
+}
+
+.torrent-details td.col-select {
+  width: 2em;
   text-align: right;
 }
 
--- a/renderer/index.js
+++ b/renderer/index.js
@@ -153,6 +153,11 @@ function cleanUpConfig () {
       delete ts.posterURL
       ts.posterFileName = infoHash + extension
     }
+
+    // Migration: add per-file selections
+    if (!ts.selections) {
+      ts.selections = ts.files.map((x) => true)
+    }
   })
 }
 
@@ -230,6 +235,9 @@ function dispatch (action, ...args) {
   }
   if (action === 'toggleSelectTorrent') {
     toggleSelectTorrent(args[0] /* infoHash */)
+  }
+  if (action === 'toggleTorrentFile') {
+    toggleTorrentFile(args[0] /* infoHash */, args[1] /* index */)
   }
   if (action === 'openTorrentContextMenu') {
     openTorrentContextMenu(args[0] /* infoHash */)
@@ -709,7 +717,7 @@ function startTorrentingSummary (torrentSummary) {
   }
 
   console.log('start torrenting %s %s', s.torrentKey, torrentID)
-  ipcRenderer.send('wt-start-torrenting', s.torrentKey, torrentID, path, s.fileModtimes)
+  ipcRenderer.send('wt-start-torrenting', s.torrentKey, torrentID, path, s.fileModtimes, s.selections)
 }
 
 //
@@ -818,6 +826,9 @@ function torrentMetadata (torrentKey, torrentInfo) {
   torrentSummary.path = torrentInfo.path
   torrentSummary.files = torrentInfo.files
   torrentSummary.magnetURI = torrentInfo.magnetURI
+  if (!torrentSummary.selections) {
+    torrentSummary.selections = torrentSummary.files.map((x) => true)
+  }
   update()
 
   // Save the .torrent file, if it hasn't been saved already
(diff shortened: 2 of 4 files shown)


dcposch:
**Limitations**

Right now if you deselect a file, it will delete the file on disk.

When you *select a previously deselected file*, it doesn't recreate the file on disk until you restart WebTorrent Desktop.

I think I'll need to modify the underlying webtorrent library.
* Currently the `file.select()` and `file.deselect()` APIs in webtorrent.js are a bit janky
* WebTorrent doesn't really deal with what happens a file is unlinked while it's running. I think the actual inode is not deleted because WebTorrent still has a file handle open.

renderer/index.js
```diff
@@ -153,6 +153,11 @@ function cleanUpConfig () {
       delete ts.posterURL
       ts.posterFileName = infoHash + extension
     }
+
+    // Migration: add per-file selections
+    if (!ts.selections) {
+      ts.selections = ts.files.map((x) => true)
```
feross: We already have a `torrentSummary.files` key. This should go there, IMO.

feross: I guess `fileModTimes` currently lives outside of `files` too. So nevermind, I think this is fine for now.

renderer/index.css
```diff
@@ -113,6 +113,10 @@ table {
   opacity: 0.3;
 }
 
+.clickable {
+  cursor: pointer;
```
feross: Native apps don't use the hand pointer, only web pages do. This will make the app feel less native.

renderer/views/torrent-list.js
```diff
@@ -217,26 +218,38 @@ function TorrentList (state) {
...
       handleClick = dispatcher('openFile', infoHash, index)
     }
+    var rowClass = 'clickable'
+    if (!isSelected) rowClass = 'disabled' // File deselected, not being torrented
+    if (!isDone && !isPlayable) rowClass = 'disabled' // Can't open yet, can't stream
     return hx`
-      <tr onclick=${handleClick} class='${rowClass}'>
-        <td class='col-icon'>
+      <tr>
+        <td class='col-icon ${rowClass}' onclick=${handleClick}>
```
feross: I wonder if we can register `handleClick` once on the `<tr>` tag and just have the onclick handler on `col-select` work? Because dispatcher calls `event.stopPropagation()` we don't need to worry about bubbling.

renderer/webtorrent.js
```diff
@@ -308,6 +315,48 @@ function getAudioMetadata (infoHash, index) {
...
+    var file = torrent.files[i]
+    if (selections[i]) {
+      file.select()
+    } else {
+      console.log('deselecting file ' + i + ' of torrent ' + torrent.name)
+      file.deselect()
+
+      // If we deselected a file, try to nuke it to save disk space
+      var filePath = path.join(torrent.path, file.path)
+      fs.unlink(filePath) // Ignore errors for now
```
feross: This is probably a bad idea because piece verification might fail if a piece goes across the file boundary and we've nuked the neighboring file. I will check what Transmission does.

renderer/index.css
dcposch: Yeah good point. I'll remove it

renderer/index.js
dcposch: So does `progress.files`. Currently `files` is immutable.

feross: Maybe in a future version. Let's not mess with the config right before this release.

renderer/webtorrent.js
dcposch: This needs support from WebTorrent: https://github.com/feross/webtorrent/issues/806

...IMO selecting individual files is to:
* save bandwidth
* make the torrent finish faster
* save disk space

it seems jank to create files (which might be many GB) that are going to stay mostly filled with zeros because we're not actually downloading them

renderer/index.js
dcposch: Yeah for sure wasn't going to do it now.

I actually think it's fine --- if we put everything into `info` then things will become even longer. For example

```
state.saved.torrents[0].files[0].name
```
...would turn into

```
state.saved.torrents[0].info.files[0].name
```

renderer/webtorrent.js
feross: Agree with the first two reasons:

- save bandwidth
- make the torrent finish faster

When deselecting a file in Transmission, it remains on disk and the progress isn't lost. This is so you can:

- resume it later
- for piece verification purposes

Sometimes files that are deselected still get downloaded to 100% if they're small and near a file that **is** selected. This happens if the piece overlaps the boundary.

feross: We do the same thing that Transmission does where we only expand files as necessary to do writes into them.

feross:
Overall LGTM, comments inline. Since this is already merged, we can fix it up in another PR.

I'll try to be better about reviewing things faster in the future so we can stop self-merging. I'll also do the same and wait 24-48 hours before merging my own changes too :)

dcposch:
@feross I agree we should get two sets of eyes on something before merging. I don't think PRs should necessarily stay open for 1+ days, though. If you write something now and someone else reviews it, why not merge it?

renderer/webtorrent.js
dcposch: K, I'll remove the `unlink` for now if that matches the behavior of Transmission. I also want to check what uTorrent does.

Also, 
> we only expand files as necessary to do writes into them

... is that possible when you're writing to both the start and end of a file? In other words, there's a 4GB file, the first and last few KB share a BitTorrent piece with the previous and next files so they get filled in while the middle is all zeros. Does that mean it takes up less than 4GB on disk?

feross: > ... is that possible when you're writing to both the start and end of a file?

I think that will cause the file to expand. You're probably right.

It's also worth noting that Transmission appends a `.part` extension to the end of incomplete files, which might allow them to use a format that lets them store partial files more efficiently than normal. Or that might just be a hack so that users don't try to open an incomplete .mp4 file and think Transmission is broken.

feross:
It's a tradeoff -- merge too fast and contributors in other time zones won't have a chance to give their input. But while we're iterating fast, I think you're right -- we should just merge after one other set of eyes has seen it. 👍

renderer/webtorrent.js
dcposch: uTorrent does what I described-- doesn't create files that are deselected

feross: Yeah, Transmission prompts upfront too, but I like how lightweight adding torrents to WebTorrent Desktop currently feels. Let's keep that feeling!

I wonder where uTorrent stores the piece data that goes off the edge.

grunjol:
> It's a tradeoff -- merge too fast and contributors in other time zones won't have a chance to give their input. 

It happens all the time

feross:
@grunjol @dcposch @ungoldman @Flet Going forward, I think we should start wait for at least one LGTM from another contributor before merging a change. And if the change touches a lot of code, then we should wait 24-48 hours before merging even if there are multiple LGTMs. 👍

ungoldman:
👍 

I think the speed advantage of self-merging (especially to master) is far outweighed by the stability and oversight reviews and merges from fellows maintainers can provide. Also makes it a lot easier for collaborators to keep track of what's going on without having to work at the same. That said I totally get the desire to iterate quickly in early stages.

Maybe an update to https://github.com/feross/webtorrent-desktop/blob/master/CONTRIBUTING.md#project-governance?

PR #482, opened 2016-05-11
bump electron version to v1.0.1

ZakariaRidouh:
Electron has been updated to v1.0.1 :smile:

Relevant:
http://electron.atom.io/blog/2016/05/11/electron-1-0


feross:
Thanks for the PR!

For electron versions, we use an explicit version (i.e. no semver ranges like `^` since there's high potential for breakage and we want to upgrade explicitly.)

I'll update the version in a separate PR.

feross:
https://github.com/feross/webtorrent-desktop/pull/488

PR #487, opened 2016-05-12
Avoid TypeError out of OS X

furstenheim:
win.setSheetOffset at main/window.js is undefined on non OS X environments. 
Just guarding against TypeError


feross:
Thanks -- good catch.

Issue #353, opened 2016-04-07
High CPU utilization 


izdi:
**Version 0.3.3 (0-a5366c3) OS X El Capitan**

Whats up with so high CPU utilization. Threads spawning during some period even after torrent's already downloaded.

![cpuwebtorrent](https://cloud.githubusercontent.com/assets/6037501/14365771/1afac802-fd18-11e5-8f1f-c7f2b896ca5f.png)

After 5 minutes
 
![after5 minutes](https://cloud.githubusercontent.com/assets/6037501/14365785/2812b3ce-fd18-11e5-8c0c-10bae8d108cb.png)

feross:
@izdi This is a known Chromium bug that will be fixed in 6-12 weeks.

See this issue: Creating then destroying connections causes 100% CPU https://bugs.chromium.org/p/webrtc/issues/detail?id=5733

feross:
I made a build **without** WebTorrent trackers (so no WebRTC connections) and will try running it overnight and see how well it fares.

feross:
The build that doesn't create WebRTC connections was still working fine for me come morning time. Hmm...

feross:
Chrome just released version 50. The fix we need is in version 51. Chrome waits 6 weeks between releases, and Electron updates the Chrome version [after 2 weeks](https://github.com/electron/electron/blob/master/docs/faq/electron-faq.md).

So we're looking at 8 weeks before this is fixed.

feross:
As a stopgap solution **for OS X only**, we are going to disable WebRTC peers in the next version of WebTorrent (v0.4.0) since the 100% CPU issue is makes the app look broken, when it's not even our fault.

Once Chrome 51 is included in Electron, we'll undo this change and re-enable WebRTC peers on OS X.

PR here: https://github.com/feross/webtorrent-desktop/pull/511

feross:
Electron 1.2.0 was just released, which includes Chrome 51. PR here: https://github.com/feross/webtorrent-desktop/pull/594

PR #511, opened 2016-05-14
HACK: OS X: Disable WebRTC peers to fix 100% CPU issue

feross:
HACK: OS X: Disable WebRTC peers to fix 100% CPU issue caused by Chrome
bug.

Fixed in Chrome 51, so we can remove this hack once Electron updates
Chrome.

For #353.

### DIFF
--- a/renderer/webtorrent.js
+++ b/renderer/webtorrent.js
@@ -28,7 +28,14 @@ global.WEBTORRENT_ANNOUNCE = defaultAnnounceList
 
 // Connect to the WebTorrent and BitTorrent networks. WebTorrent Desktop is a hybrid
 // client, as explained here: https://webtorrent.io/faq
-var client = window.client = new WebTorrent()
+var client = window.client = new WebTorrent({
+  tracker: {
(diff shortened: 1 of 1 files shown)


feross:
This required changes to `webtorrent`, `bittorrent-tracker`, and `simple-peer` to ensure that the `wrtc: false` option actually disables WebRTC even when the JS environment has WebRTC support on `window.RTCPeerConnection`.

ungoldman:
:(

feross:
@ungoldman This is just going to be around for ~4 weeks. Not the end of the world.

ungoldman:
i trust u 🐶

Lurk:
Alt text
Still have 100% cpu on maс os
Alt text
I did some profiling and
Alt text
Looks like we call LanguageDetect.detect() once at 50 ms. And it is heavy.

Maybe

feross:
@Lurk Thanks for the report and screenshots, but this is really a separate issue than the one we're talking about here.

I went ahead and disabled the offending line here: https://github.com/feross/webtorrent-desktop/commit/8ae4ac47e6f434e7099c54a09a3109788b4bae81

PR #519, opened 2016-05-16
Fix a bad bug when creating multifile torrents

dcposch:
Single-file torrents seeded in place correctly, multi file torrents did not.


feross:
LGTM

PR #520, opened 2016-05-16
Better subtitles support

dcposch:
- [x] Simplify. Remove redundant `state.playing.subtitles.enabled`. Remove hacks like regex-parsing the 'label' of each subtitle, using the label as the identifier, etc
- [x] Detect subtitles next a video file automatically
- [x] Default to the user's language instead of just the first subtitle track

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "electron-prebuilt": "1.0.2",
     "fs-extra": "^0.27.0",
     "hyperx": "^2.0.2",
+    "iso-639-1": "^1.2.1",
     "languagedetect": "^1.1.1",
     "main-loop": "^3.2.0",
     "musicmetadata": "^2.0.2",
--- a/renderer/index.js
+++ b/renderer/index.js
@@ -17,6 +17,7 @@ var appConfig = require('application-config')('WebTorrent')
 var concat = require('concat-stream')
 var dragDrop = require('drag-drop')
 var fs = require('fs-extra')
+var iso639 = require('iso-639-1')
 var mainLoop = require('main-loop')
 var path = require('path')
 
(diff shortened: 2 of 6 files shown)


dcposch:
image

grunjol:
I just tested your code, and the only problem I am getting is when adding multiples subtitles manually. 

* Add First subtitle. Subtitle is added and starts playing.
* Add second subtitle. Subtitle is added and starts playing.
* Add third subtitle. Subtitle is added but second subtitle keeps playing.

grunjol:
Oops, you already merge it. I will open an issue.

dcposch:
@grunjol yeah, right now it replaces the currently-playing subtitle only if it's in the OS language

dcposch:
see https://github.com/feross/webtorrent-desktop/blob/master/renderer/index.js#L639

that means you can drop a folder full of subtitles and it will pick the "right" one (eg if your computer is set to Spanish, it'll pick the Spanish subtitle track)

maybe we want a different behavior

grunjol:
My machine is in English.

I added the subs manually in this order, each one individually

First test
* Start playing video
* Add Portuguese => PT plays
* Add French => FR plays
* Add German => FR keeps playing

Second test
* Start playing video
* Add French => FR plays
* Add German => DE plays
* Add Portuguese => DE keeps playing

I dont know why the second is added and displayed buy not the third, and no matter the order the second is picked if no matching languages.

I agree with you, If you add a lot of subtitles you need to select which one to show, and its fine to try to select the OS-matching one. But if you add a new one, I expect that one to be selected regardless the OS language you are running.

feross:
> But if you add a new one, I expect that one to be selected regardless the OS language you are running.

Agree that this is the right behavior.

dcposch:
Already done in #529

dcposch:
@grunjol i couldn't reproduce what you described with the second subtitle you added

feross:
@dcposch See https://github.com/feross/webtorrent-desktop/issues/534 -- it might need to be closed if it's already fixed

PR #529, opened 2016-05-17
Simplify subtitles code

dcposch:
When manually adding subtitle track(s), always switch to a new track

Also fix a bug I added in 63e7aa8c7f2708b9eb6957267284e5f19073fd96

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
   "dependencies": {
     "airplay-js": "guerrerocarlos/node-airplay-js",
     "application-config": "^0.2.1",
+    "async": "^2.0.0-rc.5",
     "bitfield": "^1.0.2",
     "chromecasts": "^1.8.0",
     "concat-stream": "^1.5.1",
--- a/renderer/index.js
+++ b/renderer/index.js
@@ -14,6 +14,7 @@ var ipcRenderer = electron.ipcRenderer
 setupIpc()
 
 var appConfig = require('application-config')('WebTorrent')
+var Async = require('async')
 var concat = require('concat-stream')
 var dragDrop = require('drag-drop')
 var fs = require('fs-extra')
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -16,6 +16,7 @@
   "dependencies": {
     "airplay-js": "guerrerocarlos/node-airplay-js",
     "application-config": "^0.2.1",
+    "async": "^2.0.0-rc.5",
```
feross: Let's not use `async`. It's a fatty dependency. I have better packages, `run-parallel`, `run-series`, `run-auto`, `run-waterfall`, etc. Most of those are already in our dependency tree from webtorrent, so let's stick with them.

renderer/index.js
```diff
@@ -592,59 +593,73 @@ function addTorrent (torrentId) {
...
-  var srtToVtt = require('srt-to-vtt')
-  var LanguageDetect = require('languagedetect')
-
+function addSubtitles (files, autoSelect) {
   // Subtitles are only supported while playing video
   if (state.playing.type !== 'video') return
 
-  // No dupes allowed
+  // Read the files concurrently, then add all resulting subtitle tracks
+  console.log(files)
```
feross: stray `console.log`

renderer/index.js
```diff
@@ -592,59 +593,73 @@ function addTorrent (torrentId) {
...
   var subs = state.playing.subtitles
-  var filePath = file.path || file
-  if (subs.tracks.some((track) => track.filePath === filePath)) return
+  Async.map(files, loadSubtitle, function (err, tracks) {
+    if (err) return onError(err)
+
+    for (var i = 0; i < tracks.length; i++) {
+      // No dupes allowed
+      var track = tracks[i]
+      if (subs.tracks.some((t) => track.filePath === t.filePath)) continue
```
feross: In general, I prefer not abbreviating object variables that we're later going to modify since it makes it harder to see what's being modified. I think it's fine for other types of variables, or when we're just reading the value.

It's longer, but clearer IMO:

```js
state.playing.subtitles.tracks.push(track)
```

PR #535, opened 2016-05-18
Clean up addSubtitles

dcposch:
Fixes comments from #529

### DIFF
--- a/renderer/index.js
+++ b/renderer/index.js
@@ -14,7 +14,7 @@ var ipcRenderer = electron.ipcRenderer
 setupIpc()
 
 var appConfig = require('application-config')('WebTorrent')
-var Async = require('async')
+var parallel = require('run-parallel')
 var concat = require('concat-stream')
 var dragDrop = require('drag-drop')
 var fs = require('fs-extra')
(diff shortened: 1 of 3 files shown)


renderer/index.js
```diff
@@ -606,22 +606,22 @@ function addSubtitles (files, autoSelect) {
   if (state.playing.type !== 'video') return
 
   // Read the files concurrently, then add all resulting subtitle tracks
-  console.log(files)
-  var subs = state.playing.subtitles
-  Async.map(files, loadSubtitle, function (err, tracks) {
+  parallel(files, loadSubtitle, function (err, tracks) {
```
feross: Does this actually work? `parallel` works a bit differently than `async.map`. It takes an array of functions to execute.

I think this should be:

```js
var tasks = files.map(function (file) {
  return function (cb) {
    loadSubtitle(file, cb)
  }
})
parallel(tasks, function (err, tracks) {
  // ...
})
```

Or with arrow functions:

```js
parallel(files.map((file) => (cb) => loadSubtitle(file, cb)), function (err, tracks) {
  // ...
})
```

feross:
Other than my inline comments this LGTM!

feross:
(Let's not merge this until v0.5 is out since I accidentally already published to npm and tagged a release in git)

PR #537, opened 2016-05-19
add audio visualizer

Flet:
Hello!

I'm unsure if this will cut the mustard, but here is a simple audio visualization to show when playing audio files.

audio-visualizer

- The visualization is done using a `<canvas>` element outside of the regular rendering context.
- Should the `visualizer` instance live in the state instead of where it is now?
- Is this too CPU intensive?

Please review and feel free to push updates to the branch! 👍


feross:
Neat hack! Pretty impressive for such a small amount of code. But I think this needs a little bit more polish before we can release it to users. One thing that could be a better is that the width of the visualization varies depending on the width of the text below it.

Flet:
I think this is getting too far from `master`... closing this PR.

PR #538, opened 2016-05-19
Always handle when the user clicks a magnet link or torrent file, or uses File > Open Torrent

### LINKED ISSUE #523: Clicking magnet link / opening .torrent file does nothing if player is open
Introduced in v0.4.0.

dcposch:
- [x] If they're currently watching a video, kick them back out to the torrent list and create a new torrent
- [x] Close any open modal windows
- [x] Scroll to the bottom of the torrent list, so that the new item is visible

Later, in a separate PR, we'll make it so that Webtorrent Desktop remembers your position in the videos you've watched. That way, it won't be as bad to automatically exit ppl out of their video.

### DIFF
--- a/renderer/index.js
+++ b/renderer/index.js
@@ -87,7 +87,7 @@ function init () {
 
   // OS integrations:
   // ...drag and drop a torrent or video file to play or seed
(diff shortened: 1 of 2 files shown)


feross:
LGTM!

PR #541, opened 2016-05-19
Use lighter-weight simple-concat instead of concat-stream

feross:
These modules do the same thing.

$ browserify -r simple-concat --no-builtins | wc -c
901

$ browserify -r concat-stream --no-builtins | wc -c
91998


feross:

PR #543, opened 2016-05-20
Improve open behavior; Fix bugs in LocationHistory

feross:
Lots of intertwined changes here, might be easier to review commit-by-commit.

---

I don't think it matters whether the open comes from onOpen (opening
magnet, .torrent file, dragging file to dock, menu item) or from
dragging to the window.

These should use the same code path. The only relevant information is
the page of the app that we're on.

This change unifies the two methods, and supports dragging .torrent
files or creating a torrent when the player is active, if the dragged
files are not .srt or .vtt. We go back to the list, or to the create
torrent page in these situations, so it's not confusing for the user.

Always close open modals when handling an open.

### DIFF
--- a/renderer/index.js
+++ b/renderer/index.js
@@ -87,7 +87,7 @@ function init () {
 
   // OS integrations:
   // ...drag and drop a torrent or video file to play or seed
-  dragDrop('body', onDrag)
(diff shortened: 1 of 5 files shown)


feross:

PR #544, opened 2016-05-20
VLC tweaks

feross:
- Start video on top, so it's not obscured by other windows.

- Don't show "video title" which is just "http://localhost:xxxx"

- return after error


feross:

PR #545, opened 2016-05-20
Add announcement feature 

feross:
If there's a message returned by the given remote URL, then it will
show up for the user.

Useful in situations where the auto-updater is not working, or if
there's a security issue.


feross:

PR #551, opened 2016-05-20
Improve subtitle positioning

feross:
Before this commit, we tweaked the subtitle position by modifying the
VTT file, line by line with a regex because I did not know it was
possible to use CSS for it.

But apparently there are Shadow DOM elements that we can use instead.

This new approach improves:

- Wrapping long lines. Before, the text would go off the edge of the
screen. Now it wraps intelligently.

- The subtitles move up to get out of the way of the controls when
those are visible.

### DIFF
--- a/renderer/index.css
+++ b/renderer/index.css
@@ -837,14 +837,21 @@ body.drag .app::after {
   height: 14px;
 }
 
+/**
+ * Set the cue text position so it appears above the player controls.
+ */
+video::-webkit-media-text-track-container {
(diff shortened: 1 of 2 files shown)


dcposch:
Looks reasonable! That regex replace was pretty ugly

grunjol:
Really good change.
The old code was messing the position in the case the subtitle already has a position style defined in the line.

feross:
Thanks guys!

PR #552, opened 2016-05-20
Fewer click handlers

feross:
One more bit of cleanup for PR #529.

We can register `handleClick` once on the `<tr>` tag and just have the
onclick handler on `col-select` too. Because dispatcher calls
`event.stopPropagation()` we don't need to worry about event bubbling
up to the `<tr>`.


feross:
Review: @dcposch

dcposch:
LGTM

PR #553, opened 2016-05-21
Associate .torrent files to WebTorrent Desktop (OS X)

### LINKED ISSUE #542: .torrent files not associated to WebTorrent Desktop

0.5.1

OS X 10.11.5

**What did you do?**

Install the app.

.torrent will be associated with WebTorrent Desktop, since no other torrent clients are installed. This used to work correctly, but must have stopped working at some point.

**What actually happened?**

.torrent files are not associated.

screen shot 2016-05-19 at 18 49 49
screen shot 2016-05-19 at 18 49 52

feross:
Before this change, .torrent files would only be associated to
WebTorrent Desktop if another torrent client (like Transmission) was
installed on the system.

That's because one of the user's apps needs to define
"UTExportedTypeDeclarations".

On a fresh machine, without Transmission, WebTorrent Desktop now
associates .torrent files correctly.

So it will appear in the "Open With..." menu and the webtorrent
.torrent icon will be used for .torrent files.

Closes #542

### DIFF
--- a/bin/package.js
+++ b/bin/package.js
@@ -182,8 +182,6 @@ function buildDarwin (cb) {
     var infoPlistPath = path.join(contentsPath, 'Info.plist')
     var infoPlist = plist.parse(fs.readFileSync(infoPlistPath, 'utf8'))
 
-    // TODO: Use new `extend-info` and `extra-resource` opts to electron-packager,
-    // available as of v6.
     infoPlist.CFBundleDocumentTypes = [
       {
         CFBundleTypeExtensions: [ 'torrent' ],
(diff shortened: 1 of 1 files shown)


dcposch:
LGTM. Eventually we should have a prefs page with an option for whether you want Webtorrent to be the default app for `.torrent` files

feross:
@dcposch I believe that this Info.plist stuff only tells OS X that WebTorrent Desktop *can* open .torrent files, but it doesn't actually change your preference if you have it set to something else. But we can test this.

We do take over magnet links without asking first, though. Issue here: https://github.com/feross/webtorrent-desktop/issues/367

PR #554, opened 2016-05-21
Save video position

### LINKED ISSUE #405: Remember video position
Would be great if playing video and pressing the back button, then playing the video would resume at the last position instead of starting the video from the start.

dcposch:
Fixes #405 

Also cleaned up the styling on the list of files when you expand a torrent

### DIFF
--- a/renderer/index.css
+++ b/renderer/index.css
@@ -592,7 +592,7 @@ body.drag .app::after {
 }
 
 .torrent-details {
-  padding: 8em 12px 20px 20px;
+  padding: 8em 0 20px 0;
 }
 
 .torrent-details table {
@@ -617,7 +617,7 @@ body.drag .app::after {
 .torrent-details td {
   overflow: hidden;
   padding: 0;
-  vertical-align: bottom;
+  vertical-align: middle;
 }
 
 .torrent-details td .icon {
@@ -627,7 +627,8 @@ body.drag .app::after {
 }
 
 .torrent-details td.col-icon {
-  width: 2em;
+  width: 3em;
+  padding-left: 16px;
 }
 
 .torrent-details td.col-name {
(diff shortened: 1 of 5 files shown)


dcposch:
screen shot 2016-05-21 at 8 15 41 am

feross:
Nice, this looks really good! Especially the changes to the file list. Looks really pro.

The only thing I ran into is that automatic subtitle detection throws an exception and that prevents playback.

The default state object only contains `length` and `name` properties for each file, so it's relying on the 'metadata' event to populate the `path` property which is used here (https://github.com/feross/webtorrent-desktop/blob/ba41ce5f9a0404290766d35c19511ba07861cd24/renderer/index.js#L722).

This change (https://github.com/feross/webtorrent-desktop/pull/554/commits/ba41ce5f9a0404290766d35c19511ba07861cd24) prevents the 'metadata' event from updating `torrentSummary.files` for the default torrents, so they'll always lack a `path` property.

Once that's fixed up, and you rebase, this is good to merge. :

dcposch:
> This change (ba41ce5) prevents the 'metadata' event from updating torrentSummary.files for the default torrents, so they'll always lack a path property.

Good call! Fixing

dcposch:
> The only thing I ran into is that automatic subtitle detection throws an exception and that prevents playback

That's probably already on master, right, since this PR doesn't change subtitle behavior?

Was it `language-detect`, `srt-to-vtt`, or something else that threw the exception?

feross:
> That's probably already on master, right, since this PR doesn't change subtitle behavior?

The exception is caused by this PR, since the `path` property is not set on the files in `torrentSummary.files` anymore. The source of the bug is the 'metadata' event, but it doesn't manifest itself until the subtitles code runs because it tries to access `torrentSummary.files[0].path`.

PR #555, opened 2016-05-21
electron-prebuilt@1.1.1

feross:
No new features that affect us.

This release includes Chrome 50, which should improve the webrtc CPU usage situation on Linux. For the OS X improvement, we're still waiting for Chrome 51.


dcposch:
lgtm

Issue #539, opened 2016-05-19
Uncaught TypeError: Cannot read property 'update' of undefined


feross:
**What version of WebTorrent Desktop?** (See the 'About WebTorrent' menu)

0.5.1

**What operating system and version?**

OS X 10.11.5

**What did you do?**

Open the app, look at developer console

**What did you expect to happen?**

No errors.

**What actually happened?**

"Uncaught TypeError: Cannot read property 'update' of undefined"

furstenheim:
It seems that 'leave-full-screen' (window.js:118) is triggered by electron before the application has init and this tries to update vdom before it is defined

PR #557, opened 2016-05-21
Fix bug 539

furstenheim:
This fixes bug #539. However I'm not sure it is the ideal solution


dcposch:
looks reasonable to me

@feross do you know why `setupIPC` was up there with the `require()s`?

feross:
Yes, putting setupIPC() up top improved app startup performance by 300ms.

We should find another way to fix this issue IMO. Startup time is 500ms right now. Don't want to go up to 800ms again.

feross:
Let's try this instead? https://github.com/feross/webtorrent-desktop/pull/567

PR #559, opened 2016-05-21
Add support for pasting a `instant.io` link.

### LINKED ISSUE #547: Pasting instant.io link should work
Let's support pasting a link from `instant.io`. All we need to do is strip the part after the `#` and treat it as the `torrentId`.

demoneaux:
Closes #547.


dcposch:
LGTM

feross:
Nice work, @demoneaux!

PR #562, opened 2016-05-22
Add support for video progress time in player controls.

### LINKED ISSUE #351: show playback time (current/total)
Example from youtube:

screen shot 2016-04-07 at 10 43 54 am

demoneaux:
Closes #351.

### DIFF
--- a/renderer/index.css
+++ b/renderer/index.css
@@ -827,6 +827,11 @@ body.drag .app::after {
   outline: none;
 }
 
+.player-controls .time {
+  float: left;
+  line-height: 33px;
+}
+
 .player .playback-bar:hover .loading-bar {
   height: 5px;
 }
--- a/renderer/views/player.js
+++ b/renderer/views/player.js
@@ -426,6 +426,14 @@ function renderPlayerControls (state) {
     </div>
   `)
 
+  // Show video playback progress
+  var videoProgress = formatPlaybackProgress(state.playing.currentTime, state.playing.duration)
+  elements.push(hx`
+    <span.time>
+      ${videoProgress[0]} / ${videoProgress[1]}
+    </span>
+  `)
+
   // Finally, the big button in the center plays or pauses the video
   elements.push(hx`
     <i class='icon play-pause' onclick=${dispatcher('playPause')}>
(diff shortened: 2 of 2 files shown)


renderer/views/player.js
```diff
@@ -540,3 +548,37 @@ function cssBackgroundImageDarkGradient () {
   return 'radial-gradient(circle at center, ' +
     'rgba(0,0,0,0.4) 0%, rgba(0,0,0,1) 100%)'
 }
+
+// Format the playback time of the video
+function formatPlaybackProgress (currentTime, duration) {
```
feross: Change this function to just format a single time instead of both `currentTime` and `duration` since the code should be exactly the same for both. Then you can just call the function on `currentTime` and `duration` separately.

demoneaux: I think the reason why I did both `currentTime` and `duration` together was to make sure that the time format for both would be the same. Otherwise this could lead to something like `01:02 / 02:20:47` when the user is playing, which can be confusing, compared to `00:01:02 / 02:20:47`.

renderer/views/player.js
```diff
@@ -540,3 +548,37 @@ function cssBackgroundImageDarkGradient () {
...
+  if (typeof currentTime !== 'number' || typeof duration !== 'number') {
```
feross: This won't catch `NaN` since `typeof NaN === 'number'`, so we should check for that case too.

renderer/index.css
```diff
@@ -827,6 +827,11 @@ body.drag .app::after {
   outline: none;
 }
 
+.player-controls .time {
```
demoneaux: This currently causes the play button to be off centre if the length of the progress time text is long and the window size is small. Not too sure how to circumvent that though.

feross:
I'm +1 on this, but see comments inline.

renderer/index.css
feross: I wouldn't worry about it, for now. I plan to send a PR soon to cleanup the player controls.

renderer/views/player.js
feross: Okay, fair enough. 👍

demoneaux:
@feross I've updated the PR to check for NaN.

renderer/views/player.js
feross: Actually, looks like YouTube doesn't do this. Interesting.

screen shot 2016-05-22 at 21 17 00

Issue #561, opened 2016-05-21
H.265 video playback


soofuzz:
Hey was just testing this out on x265 (HEVC) video torrent but i get a black screen on playback. Does the player not support x265? Unless there's a way to stream out onto an external player like mpv.

feross:
This is fixed now, thanks to https://github.com/feross/webtorrent-desktop/issues/564. You'll now see a message about using VLC when a HEVC video is detected.

This will go out in the next release of WebTorrent (v0.6).

jakop345:
Looks like there is x265 codec support in chromium dev builds.
https://github.com/henrypp/chromium/releases

feross:
@jakop345 If that's true, then I guess we should get support for it automatically when we update Electron?

demoneaux:
I think that's a custom build of Chromium specifically enabling support for additional codecs. The default Chromium build used by Electron probably won't have it enabled.

feross:
Makes sense. I don't think Chrome wants to support H.265, and I don't blame them.

PR #563, opened 2016-05-22
Handle unsupported video codec (e.g. H.265)

mathiasvr:
Notifies when a video codec is not supported (such as H.265 in #561), and allows it to be played in vlc similar to how unsupported audio in videos are already handled.


feross:
LGTM, thanks!

PR #566, opened 2016-05-23
More consistent controls

feross:
Tried to match the YouTube player UI more closely. There's still work to do, but this is a huge improvement for consistency:

screen shot 2016-05-22 at 21 59 22

Compare to:

screen shot 2016-05-22 at 22 15 54

- Increased size of all controls
- All buttons consistently get lighter on hover, without redundant css
- Play button moves to left, just like YouTube
- Time tweaks: thinner font, less verbose (0:xx instead of 00:xx, and omit hours when they're 0)
- Bigger volume slider nub
- Show back/forward buttons in top left on Windows/Linux, when in player
  - So it's consistent with OS X version, and the back button doesn't move around
  - Also, there are technically situations in which the forward button is needed from the player
    - Like if you start creating a torrent from the player, then go back, then want to go forward again
  - Make the background of the header bar transparent since it's mostly empty and looks terrible otherwise

This is what it looks like on Windows:

screen shot 2016-05-22 at 22 12 47

### DIFF
--- a/main/menu.js
+++ b/main/menu.js
@@ -134,10 +134,10 @@ function onPlayerOpen () {
   getMenuItem('Increase Volume').enabled = true
   getMenuItem('Decrease Volume').enabled = true
   getMenuItem('Add Subtitles File...').enabled = true
-  getMenuItem('Skip forward 10 seconds').enabled = true
-  getMenuItem('Skip back 10 seconds').enabled = true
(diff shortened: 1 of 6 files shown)


dcposch:
LGTM!

the other changes are really subtle, but putting the play button on the far left definitely feels cleaner

PR #568, opened 2016-05-23
Reduce startup jank, improve startup time

feross:


renderer/index.js
```diff
@@ -42,18 +40,29 @@ appConfig.filePath = path.join(config.CONFIG_PATH, 'config.json')
 // This dependency is the slowest-loading, so we lazy load it
 var Cast = null
 
-// For easy debugging in Developer Tools
-var state = global.state = State.getInitialState()
```
feross: all the variables in this scope are global, so no need for `global.state`

renderer/index.css
```diff
@@ -50,19 +50,22 @@ table {
...
-
 .app {
   -webkit-user-select: none;
   -webkit-app-region: drag;
   height: 100%;
   display: flex;
   flex-flow: column;
-  animation: fadein 0.3s;
   background: rgb(40, 40, 40);
+  animation: fadein 1s;
```
feross: this UI janks a lot less on my macbook 12" during startup

dcposch: just tested it. feels v nice! in fact it feels smoother and more professional than other electron apps now

* Atom: pops up a gray page, then a half second later gray text appears, then a half second later the syntax highlighting activates

* Spotify: pops up a gray page, then some HTML, redoes the layout a couple times, janks around, finally settles down a few seconds later

* Webtorrent Desktop: pops up a gray page, fades in the UI 👍

dcposch:
LGTM

PR #569, opened 2016-05-23
Cleanup unsupported codec detection

feross:
Review: @dcposch


dcposch:
Turns a flat `if else` into a nested `if else`, I don't know if that's cleaner but it's fine either way

The variable that used to be called `video` should def be `elem`, you're right. LGTM

feross:
Oh, you're right about the nesting. Good call.

PR #570, opened 2016-05-23
Cleanup unsupported codec detection (pt 2)

feross:
Review: @dcposch


dcposch:
lgtm, i like this better than the previous one

one edge case we should handle in the future: if you watch a silent film (eg an old black-and-white), then this will incorrectly say "Audio codec unsupported". we want to test whether the video has any audio tracks

Issue #242, opened 2016-03-27
Sort file list alphabetically


feross:

dcposch:
are you sure we want this? some torrents have the files listed in certain order on purpose

feross:
Sure, we can revisit this later if it's an issue.

dcposch:
@feross do we still want this?

feross:
Fixed by https://github.com/feross/webtorrent-desktop/pull/571

PR #571, opened 2016-05-23
#242 add sort by file name

Lurk:


renderer/views/torrent-list.js
```diff
@@ -205,6 +205,15 @@ function TorrentList (state) {
...
 
+  function fileSortHelper (a, b) {
+    var result = 0
+    if (a.name > b.name) {
+      result = 1
+    } else if (a.name < b.name) {
+      result = -1
+    }
+    return result
+  }
```
feross: I think this can function can just be a one-liner `return a.name - b.name`.

feross: I'll fix this up.

feross: Oh nevermind. Can't do that in JS, my bad.

PR #573, opened 2016-05-24
Cast menu

### LINKED ISSUE #301: Cast: support more than one device of the same type
I have more than one device supporting DLNA. While testing the feature, I found that webtorrent-desktop was using any of them randomly. This is because it depends on how fast the device answer the broadcast signalling (discovery), and the one that will play is the latest to answer.

This should happen (didn't test it but the logic is the same) with airplay and chromecast too.

I am thinking about college dorms, or shared connections with the same broadcast domain.

You should be able to select which device you will cast to.

dcposch:
Fixes #301

screen shot 2016-05-23 at 9 37 47 pm

### DIFF
--- a/bin/check-deps.js
+++ b/bin/check-deps.js
@@ -60,10 +60,10 @@ function main () {
   var packageDeps = findPackageDeps()
 
   var missingDeps = usedDeps.filter(
-    (dep) => !packageDeps.includes(dep) && !BUILT_IN_DEPS.includes(dep)
+    (dep) => !includes(packageDeps, dep) && !includes(BUILT_IN_DEPS, dep)
   )
   var unusedDeps = packageDeps.filter(
-    (dep) => !usedDeps.includes(dep) && !EXECUTABLE_DEPS.includes(dep)
+    (dep) => !includes(usedDeps, dep) && !includes(EXECUTABLE_DEPS, dep)
   )
 
   if (missingDeps.length > 0) {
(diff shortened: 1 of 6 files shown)


dcposch:
* Right now, it shows a menu even when there's only one cast device of a given type. Maybe we just want to connect to that device directly in that case.
* We're using the same styling as the subtitle selector. Most other video apps will show a bigger menu in the middle of the screen.

Thoughts?

@feross

grunjol:
Coming from `master` to `dc/cast` branch (expected upgrade path) I get a black screen (and this error on dev console) when resuming a previously started torrent

```
rendering error: Cannot read property 'getDevices' of null
	TypeError: Cannot read property 'getDevices' of null
    at renderPlayerControls (/home/grunjol/workspaces/wt/webtorrent-desktop/renderer/views/player.js:411:31)
    at Object.Player (/home/grunjol/workspaces/wt/webtorrent-desktop/renderer/views/player.js:25:9)
    at getView (/home/grunjol/workspaces/wt/webtorrent-desktop/renderer/views/app.js:82:20)
    at App (/home/grunjol/workspaces/wt/webtorrent-desktop/renderer/views/app.js:45:30)
    at render (file:///home/grunjol/workspaces/wt/webtorrent-desktop/renderer/index.js:186:12)
    at redraw (/home/grunjol/workspaces/wt/webtorrent-desktop/node_modules/main-loop/index.js:65:23)
    at /home/grunjol/workspaces/wt/webtorrent-desktop/node_modules/main-loop/node_modules/raf/index.js:72:10
```

dcposch:
cast

@feross rebased and improved. I've tested Chromecast and Airplay but not DLNA.

PR #587, opened 2016-05-25
Electron 1.2.0

feross:
- Update to Electron 1.2.0 (Chrome 51)
- Use .ico format on Windows for best rendering quality, instead of resized .png files
- Fix crashReporter not working (Windows)
- Fix window resizing when aspect ratio is enforced (OS X)


feross:
Tested on OS X, works fine.

Still failing on Windows, even with the newer Electron version :(

feross:
See https://github.com/electron/electron/issues/5629 for the reason that Windows crashes.

We need to downgrade to an Electron that uses Chrome 49.

feross:
New plan: wait until next Electron release that contains Chrome 51 (supposedly imminent)

feross:
Updated PR to Electron 1.2.0 (Chrome 51)

PR #588, opened 2016-05-26
Show error when drag-dropping hidden files

dcposch:
screen shot 2016-05-25 at 11 13 57 pm

### DIFF
--- a/renderer/views/player.js
+++ b/renderer/views/player.js
@@ -37,7 +37,8 @@ function renderMedia (state) {
 
   // Unfortunately, play/pause can't be done just by modifying HTML.
   // Instead, grab the DOM node and play/pause it if necessary
-  var mediaElement = document.querySelector(state.playing.type) /* get the <video> or <audio> tag */
+  // Get the <video> or <audio> tag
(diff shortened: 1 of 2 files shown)


renderer/views/player.js
```diff
@@ -278,18 +304,19 @@ function renderSubtitlesOptions (state) {
     var isSelected = state.playing.subtitles.selectedIndex === ix
     return hx`
       <li onclick=${dispatcher('selectSubtitle', ix)}>
-        <i.icon>${isSelected ? 'radio_button_checked' : 'radio_button_unchecked'}</i>
+        <i.icon>${'radio_button_' + (isSelected ? 'checked' : 'unchecked')}</i>
```
feross: this is fine, but not necessarily better because it breaks grepping for `radio_button_checked`

feross:
LGTM. Thanks for fixing this!

In the future, we need to think about how to handle errors consistently. We have some errors that show up as a page in the window (like this one), some that are notifications at the top of the window that disappear, and some like in announcement.js that show up as native OS dialog windows.

PR #589, opened 2016-05-26
Simplify shortcuts. Go Back menu item

### LINKED ISSUE #585: Keyboard capture of Space and ESC is super aggressive
g**What version of WebTorrent Desktop?** (See the 'About WebTorrent' menu)

0.6.0

OS X 10.11.5

**What did you do?**

Try to use of an OS shortcut with `ESC` or `Space` while WebTorrent Desktop is running.

OS shortcuts should work.

**What actually happened?**

Any system-wide OS shortcut that makes use of `ESC` or `Space` does not work if WebTorrent Desktop is the active window, because it is captured by the app and the `electron-localshortcut` package.

dcposch:
screen shot 2016-05-25 at 11 32 15 pm

Fixes #585

### DIFF
--- a/main/index.js
+++ b/main/index.js
@@ -12,7 +12,6 @@ var handlers = require('./handlers')
 var ipc = require('./ipc')
 var log = require('./log')
 var menu = require('./menu')
-var shortcuts = require('./shortcuts')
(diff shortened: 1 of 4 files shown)


feross:
LGTM

Issue #581, opened 2016-05-24
Duplicate torrent/magnet can be added if torrent is stopped


mathiasvr:
**What version of WebTorrent Desktop?** (See the 'About WebTorrent' menu)
0.6.0
**What operating system and version?**
OS X 10.11.5
**What did you do?**
Add the same torrent or magnet after stopping the download
**What did you expect to happen?**
Get duplicate error message (as is the case when torrent is downloading)
**What actually happened?**
The torrent was added again (twice in list)

feross:
Thanks, I was just going to post this issue.

feross:
Fixed by @mathiasvr's PR #593!

PR #593, opened 2016-05-26
Adding duplicate torrent when stopped

mathiasvr:
I took a look at the issue #581 i created about adding a stopped torrent that is already in the list.
I was able to fix the issue by replacing the torrentKey of the first torrent with the key of the duplicate, and mimic a toggleTorrent.

This is not an elegant solution, but maybe it could do for now?

Also, I was wondering why a torrent have a torrentKey instead of being identified by the infoHash?

### DIFF
--- a/renderer/main.js
+++ b/renderer/main.js
@@ -878,12 +878,21 @@ function torrentInfoHash (torrentKey, infoHash) {
     torrentSummary ? 'existing' : 'new', torrentKey)
 
   if (!torrentSummary) {
-    torrentSummary = {
-      torrentKey: torrentKey,
-      status: 'new'
+    // Check if another torrent has the same infoHash
+    var duplicate = state.saved.torrents.find((x) => x.infoHash === infoHash)
+
+    if (duplicate) {
+      duplicate.torrentKey = torrentKey
+      duplicate.status = 'new'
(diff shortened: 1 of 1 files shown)


demoneaux:
Any updates for this PR?

mathiasvr:
Just updated PR to accommodate refactor.
I don't know if there is interest in merging, maybe somebody has a better idea on how to do this?
I tried to keep changes minimal, but it would probably be better to determine the duplicate before starting a new webtorrent session.

feross:
Thanks for this PR, @mathiasvr! And for updating it to accommodate the refactor.

The reason for `torrentKey` is that some `torrentId`s don't have an info hash that can be determined trivially (i.e. synchronously). For example, an http torrent url or filesystem path where the file has to be fetched asynchronously. So we need a way to identify the torrent before it's info hash has been determined.

We should document this better in the comments. PR welcome for that 👍

feross:
Upon further thought, I think we should error when they adda duplicate torrent instead of setting the state to `new`, so that it's consistent with how the app behaves when adding a duplicate torrent and the current one is actively torrenting.

feross:
https://github.com/feross/webtorrent-desktop/pull/612

Issue #574, opened 2016-05-24
Remember volume level


grunjol:
**What version of WebTorrent Desktop?** (See the 'About WebTorrent' menu)
Version 0.5.1 (0.94.0)
**What operating system and version?**
Ubuntu 16.04
**What did you do?**
* Play any torrent
* Adjust volume to 0.5 (50%)
* Go back to torrent list
* Go forward (or select a new torrent)

**What did you expect to happen?**

* Volume should be 0.5 (50%)

**What actually happened?**
* Volume is 1.0 (100%)

grunjol:
IMHO, volume level be a global value

feross:
Fixed by @furstenheim's PR #600

PR #600, opened 2016-05-27
Persist volume across videos

furstenheim:
The volume preference is preserved during the same session.
#574


renderer/index.js
```diff
@@ -458,7 +458,8 @@ function setVolume (volume) {
   if (isCasting()) {
     Cast.setVolume(volume)
   } else {
-    state.playing.setVolume = volume
```
grunjol: I think this is the way to dispatch that you need to change the volume to the <video> tag . 
Also, this is only for the local renderer and not the cast devices

furstenheim:
Which tag do you refer to? As for the cast devices, do you have any idea how can I mock one to do that part as well?

grunjol:
setVolume holds the value to be set in the <video> DOM element. thats why is set and cleared in the UI.

furstenheim:
@grunjol it should be better now. In any case, only for the local renderer.

feross:
Thanks for the PR, @furstenheim!

This works for now, but it would be better if this was a permanent setting and if it was rendered directly into the volume control html as a default starting volume.

Thanks for taking a stab at this!

feross:
Rebased and merged your commits onto master. GitHub is showing this PR as not merged, but it really is.

PR #603, opened 2016-05-30
Fix issue with tray on Windows.

demoneaux:
Note: There is still an issue with the tray, when the WebTorrent app is hidden, the tray item "Hide to tray" does not change to "Show WebTorrent", but I can't seem to figure out how to fix it.


feross:
Thanks -- this was my fault from the recent refactor.

PR #604, opened 2016-05-30
Fix issue with file sorting.

### LINKED ISSUE #591: Wrong file order due to sorting
feross/webtorrent-desktop@7de3d3c

Windows 10

**What did you do?**
I opened a video by pressing play on the torrent list.

Video plays.

**What actually happened?**
A grey screen shows up with the following error:
```
rendering error: Cannot read property 'length' of undefined
	TypeError: Cannot read property 'length' of undefined
    at renderMedia (C:\Users\weeseng\Documents\GitHub\webtorrent-desktop\renderer\views\player.js:64:31)
    at Object.Player (C:\Users\weeseng\Documents\GitHub\webtorrent-desktop\renderer\views\player.js:24:21)
    at getView (C:\Users\weeseng\Documents\GitHub\webtorrent-desktop\renderer\views\app.js:82:20)
    at App (C:\Users\weeseng\Documents\GitHub\webtorrent-desktop\renderer\views\app.js:45:30)
    at render (file:///C:/Users/weeseng/Documents/GitHub/webtorrent-desktop/renderer/index.js:186:12)
    at redraw (C:\Users\weeseng\Documents\GitHub\webtorrent-desktop\node_modules\main-loop\index.js:65:23)
    at C:\Users\weeseng\Documents\GitHub\webtorrent-desktop\node_modules\raf\index.js:72:10
```

Seems like there is an error when there are no subtitle tracks.

demoneaux:
Closes #591.

### DIFF
--- a/renderer/views/home.js
+++ b/renderer/views/home.js
@@ -191,12 +191,13 @@ function TorrentList (state) {
     } else {
       // We do know the files. List them and show download stats for each one
       var fileRows = torrentSummary.files
+        .map((file, index) => ({ file, index }))
         .sort(function (a, b) {
-          if (a.name < b.name) return -1
-          if (b.name < a.name) return 1
+          if (a.file.name < b.file.name) return -1
+          if (b.file.name < a.file.name) return 1
           return 0
         })
-        .map((file, index) => renderFileRow(torrentSummary, file, index))
+        .map((object) => renderFileRow(torrentSummary, object.file, object.index))
 
       filesElement = hx`
         <div class='files'>


demoneaux:
Note: this does not have any effect on already downloaded torrents which have their order messed up.

feross:
Wow, this issue was caused by `array.sort()` being mutable in JS. I forgot about that...

demoneaux:
Yeah :( This means that some of the old torrents downloaded by people are unusable. Maybe we should recommend deleting older torrents/at least give a notice of the bug?

feross:
I think that we can fix this by deleting `torrentSummary.files` and `torrentSummary.defaultPlayFileIndex` for all torrents that have a `defaultPlayFileIndex` property. Then, the next time they start the torrent, these will get set again.

feross:
Commit here: https://github.com/feross/webtorrent-desktop/pull/610/commits/4ee36f459f55ff2aca7c6942b8498ef852ff3e41

@demoneaux @dcposch Let me know if you think this could cause any issues.

Issue #246, opened 2016-03-27
Consider making add button accept non .torrent files and start seeding them


feross:
That way it's consistent with drag and drop.

feross:
Closed by @mathiasvr in https://github.com/feross/webtorrent-desktop/pull/607

PR #607, opened 2016-05-31
Make add button accept non .torrent files (seeding)

mathiasvr:
Making the add button consistent with drag and drop as described in #246, is this correct?

Note that the `openTorrentFile` dispatch is no longer used, should it be removed?

### DIFF
--- a/main/dialog.js
+++ b/main/dialog.js
@@ -2,7 +2,8 @@ module.exports = {
   openSeedFile,
   openSeedDirectory,
   openTorrentFile,
-  openTorrentAddress
+  openTorrentAddress,
+  openAddFiles
 }
 
 var electron = require('electron')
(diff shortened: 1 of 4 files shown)


demoneaux:
I think the original idea was to allow it to accept both torrents and files to seed?

mathiasvr:
@demoneaux The open file dialog now accepts any file type, if only .torrent files are selected they are added, otherwise the files are prepared for creating a new torrent (to seed) - just like when you drag the files into window (using the same function `onOpen`). Is this not the idea, or am I missing something?

feross:
Looks good to me. Thanks for the nice PRs, @mathiasvr!

mathiasvr:
@feross No problem! Thanks for creating this awesome project 😄

PR #609, opened 2016-05-31
Fix 'fullscreen' and 'always on top' menu items

### LINKED ISSUE #608: Uncaught exception throw while trying to use function under Menu-View
0.6.1 master branch

 Windows 10 - 10586

**What did you do?**
Click F11 key / Menu-View-Full Screen / Menu-View-Float on Top.

In the main view, it should become fullscreen or can be set the form on top.

**What actually happened?**
show a error dialog as bellow.
Error

mathiasvr:
Fixes #608.

Should the menu item `Float on Top` be renamed to `Always on Top` as the function?

### DIFF
--- a/main/menu.js
+++ b/main/menu.js
@@ -149,12 +149,12 @@ function getMenuTemplate () {
           accelerator: process.platform === 'darwin'
             ? 'Ctrl+Command+F'
             : 'F11',
-          click: () => windows.toggleFullScreen()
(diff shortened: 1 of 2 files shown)


feross:
Thanks -- the refactor of the main process code was pretty large and I knew that there would be a couple issues like this. Many thanks for opening a PR to fix this!

We can keep the existing name "Float on Top". I just wanted the function name to match Electron's `alwaysOnTop`.

PR #610, opened 2016-05-31
Semver-based migrations

feross:
- Put migrations into a separate file: migrations.js
- Add semver, for more specific version-based migrations.
- Change `state.saved.version` to semver app version.

Also:

- Fix exception caused by incorrect file ordering  https://github.com/feross/webtorrent-desktop/pull/604#issuecomment-222805214


demoneaux:
LGTM.

feross:
Thanks @demoneaux!

PR #612, opened 2016-06-01
Error on duplicate add (fixes to PR #593)

feross:
See https://github.com/feross/webtorrent-desktop/pull/593#issuecomment-222897517


feross:
Self-merging to get v0.7.0 out the door. Post-review welcome! @dcposch @ungoldman @grunjol

PR #613, opened 2016-06-01
Fix IPC race condition in startup 

feross:
This should eliminate startup bugs caused by `ipcReady` being sent too eagerly from the renderer. I originally moved `ipcReady` to as early as possible in order to make the app launch faster.

But if we make the window show by default, instead of being hidden and waiting for `ipcReady`, then startup doesn't actually feel any slower.

### DIFF
--- a/renderer/main.css
+++ b/renderer/main.css
@@ -65,7 +65,7 @@ table {
   display: flex;
   flex-flow: column;
   background: rgb(40, 40, 40);
-  animation: fadein 1s;
+  animation: fadein 0.5s;
 }
 
 .app:not(.is-focused) {
--- a/renderer/main.js
+++ b/renderer/main.js
@@ -3,18 +3,9 @@ console.time('init')
 var crashReporter = require('../crash-reporter')
 crashReporter.init()
 
-var electron = require('electron')
-
-// Electron apps have two processes: a main process (node) runs first and starts
-// a renderer process (essentially a Chrome window). We're in the renderer process,
-// and this IPC channel receives from and sends messages to the main process
-var ipcRenderer = electron.ipcRenderer
-
-// Listen for messages from the main process
-setupIpc()
-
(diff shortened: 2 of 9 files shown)


feross:
Self-merging to get v0.7.0 out the door. Post-review welcome! @dcposch @ungoldman @grunjol

renderer/main.css
```diff
@@ -65,7 +65,7 @@ table {
   display: flex;
   flex-flow: column;
   background: rgb(40, 40, 40);
-  animation: fadein 1s;
+  animation: fadein 0.5s;
```
dcposch: 👍

renderer/main.js
```diff
@@ -90,6 +86,9 @@ function init () {
   })
   document.body.appendChild(vdomLoop.target)
 
+  // Listen for messages from the main process
+  setupIpc()
```
dcposch: This is the same thing as #557, right?

renderer/main.js
```diff
@@ -486,13 +483,15 @@ function setupIpc () {
   ipcRenderer.on('wt-poster', (e, ...args) => torrentPosterSaved(...args))
   ipcRenderer.on('wt-audio-metadata', (e, ...args) => torrentAudioMetadata(...args))
   ipcRenderer.on('wt-server-running', (e, ...args) => torrentServerRunning(...args))
+
+  ipcRenderer.send('ipcReady')
```
dcposch: This shouldn't matter,  right? Sending the `ipcReady` event and adding all those event handlers all happens on the same tick either way

dcposch:
LGTM

renderer/main.js
feross: Basically.

renderer/main.js
feross: Right. But I think it's clearer.

Issue #560, opened 2016-05-21
Remove torrent with all data


chinabala:
**What did you expect to happen?**
remove torrent and all stream data from HDD
 
**What actually happened?**
it only remove the torrent but not the stream file(s). When torrent stream list is building up on WebTorrent, it getting really hard when HDD storage is not enough and difficult to find what stream has been watched and or to locate where they are.

Can anyone bring up a function that can remove all data of a torrent/stream in once?

demoneaux:
I think this is a good idea. Maybe a modal when people click the delete torrent button, with delete files or just delete torrent only option?

kimo59660:
I think this is a very nice good idea.

feross:
I think we should solve this similar to how Transmission does. Here's a proposal.

Add a right-click menu with two new items "Remove From List..." and "Remove Data File...":

![screen shot 2016-06-05 at 7 08 03 pm](https://cloud.githubusercontent.com/assets/121766/15810228/6349238a-2b51-11e6-8920-bf11c8b78e7c.png)

Clicking the "X" icon or selecting "Remove from List..." from the right-click menu should show a confirmation like this:

![screen shot 2016-06-05 at 7 07 48 pm](https://cloud.githubusercontent.com/assets/121766/15810251/77135a7a-2b51-11e6-9e8e-6cdd285507b7.png)

Selecting "Remove Data File..." from the right-click menu should show a confirmation like this:

![screen shot 2016-06-05 at 7 08 07 pm](https://cloud.githubusercontent.com/assets/121766/15810258/88a8b960-2b51-11e6-8518-cd41ee512734.png)

This way, the default is safe -- we don't delete files unless the user explicitly selects that option.

Thoughts?

dcposch:
We only need the "transfer is active" warning when a file hasn't finished downloading yet. (I don't think it would be that bad to just delete right away, with no modal dialog.)

dcposch:
Done, w/out dialogs for now: #641

PR #615, opened 2016-06-01
Sketch: Remove torrent/data modal

mathiasvr:
I think an option to remove all data, as requested in #560, is a good idea, so I made a simple modal for choosing this, like @demoneaux suggested.

skaermbillede 2016-06-01 kl 16 23 58

Maybe this should just be an option in the right-click menu or preferences, but it might be nice for new users to have the choice at first? A cancel button could also be added?

Thoughts?

### DIFF
--- a/renderer/main.css
+++ b/renderer/main.css
@@ -326,7 +326,6 @@ table {
 }
 
 .create-torrent input.torrent-is-private {
-  width: initial;
   margin: 0;
 }
 
@@ -404,7 +403,7 @@ button.button-raised:active {
  * OTHER FORM ELEMENT DEFAULTS
  */
 
-input {
+input[type='text'] {
   background: transparent;
   width: 300px;
   padding: 6px;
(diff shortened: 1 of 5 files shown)


feross:
@mathiasvr Thanks for taking the initiative and sending this PR!

I think this should be an item in the right-click menu. I think the average user will be confused by an option like this -- "what's the difference between deleting the torrent vs. deleting all the data?" 

I wrote a comment with my thoughts here: https://github.com/feross/webtorrent-desktop/issues/560#issuecomment-223854963

mathiasvr:
@feross I think you are right about these options being confusing. I read your comments on the issue and updated the PR so that the modal now looks like this:
skaermbillede 2016-06-13 kl 15 40 19
I prepared a `keepFiles` modal property so the modal can be used for deleting data as well.

I went for a simple message but we could add more info like Transmission does?

mathiasvr:
Just rebased everything on top of master and squashed the commits. I have integrated the modal with the remove data feature from #641.

dcposch:
Looks great to me. I'll rebase it again and merge.

PR #626, opened 2016-06-03
Restore windows position when app is restarted

### LINKED ISSUE #181: Window position should be restored when app is restarted
On app startup, the main window should start in the same position as when the app was quit.

rparrapy:
Closes issue #181.
I am not sure about the cleanness of reading saved state in the main window module, feedback/alternative approaches are more than welcome :)

### DIFF
--- a/main/windows/main.js
+++ b/main/windows/main.js
@@ -22,14 +22,24 @@ var config = require('../../config')
 var log = require('../log')
 var menu = require('../menu')
 var tray = require('../tray')
+var State = require('../../renderer/lib/state')
 
 var HEADER_HEIGHT = 37
 var TORRENT_HEIGHT = 100
+var state
 
 function init () {
   if (main.win) {
     return main.win.show()
   }
+
+  State.load(createWindow)
+}
+
+function createWindow (err, _state) {
+  if (err) return onError(err)
+  state = _state
+
   var win = main.win = new electron.BrowserWindow({
     backgroundColor: '#1E1E1E',
     darkTheme: true, // Forces dark theme (GTK+3)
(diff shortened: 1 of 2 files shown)


renderer/main.js
```diff
@@ -314,6 +314,8 @@ function dispatch (action, ...args) {
     saveStateThrottled()
   }
   if (action === 'saveState') {
+    state.saved.window = state.saved.window || {}
+    state.saved.window.bounds = args[0]
```
mathiasvr: To avoid dispatching bounds as an argument we could do something like this instead:
```
{ x: window.screenX, y: window.screenY, width: window.outerWidth, height: window.outerHeight }
```
Thoughts?

main/windows/main.js
```diff
@@ -80,6 +92,9 @@ function init () {
 }
 
 function dispatch (...args) {
+  if (args[0] === 'saveState') {
```
mathiasvr: If we choose to go with dispatching the bounds (otherwise see my other comment), I think we should pass the bounds directly in `main/index.js` like this: ` windows.main.dispatch('saveState', windows.main.win.getBounds())`, or create a separate method to avoid special cases in the `dispatch` method.

main/windows/main.js
```diff
@@ -22,14 +22,24 @@ var config = require('../../config')
...
 function init () {
   if (main.win) {
     return main.win.show()
   }
+
+  State.load(createWindow)
+}
+
+function createWindow (err, _state) {
+  if (err) return onError(err)
```
mathiasvr: Instead of returning on state.load error, we should just load the default window settings.

mathiasvr:
Hi, I added some comments to your PR.

However an alternative to the current approach could be to always load the window with default settings, and then dispatch the bounds back from renderer/main.js during initialization. This way we could avoid adding `State` in `main/window/main.js` since it currently belongs in `renderer/lib/`.

Probably need more opinions on this though! 😅

renderer/main.js
rparrapy: I agree with your suggestion, did not notice that _window_ was a available in the renderer. Dispatching bounds is more cumbersome.

rparrapy:
Thanks for the feedback! I agree the line comments you shared.

Maybe not including ```State``` in ```main/window/main.js``` would be more appropriate. I tried something along those lines, but I could see the window moving from the _default_ position to the _saved_ one, not really a nice effect. I probably messed up somewhere.

If we decide to go with this PR, I can fix the minor issues you pointed out in your comments.

mathiasvr:
If the movement is visible I think loading `State` in `main/window/main.js` is better 👍 
We could also load the state in `main/index.js` and pass it to `windows.main.init()`.
Either way the state module should probably be moved then, but maybe someone has a better idea?

Flet:
Closing this PR as this was fixed in #827 👍

PR #632, opened 2016-06-10
handle play/pause when window is hidden

### LINKED ISSUE #598: Play "media key" does not pause video when window is in background

0.6.1

OS X 10.11.5

**What did you do?**

Start playing Sintel, then focus another window so that the WebTorrent window is fully blocked. Then press the Play media key on the keyboard.

The video should pause (i.e. the audio should stop playing)

**What actually happened?**

Video continues to play until the window is focused. This is probably because of `requestAnimationFrame` again.

mathiasvr:
Fixes #598.

This could also be done using the already implemented `state.window.isFocused`, however I added a `webkitvisibilitychange` event listener, since we only need to force the render when the window is completely hidden and not just out of focus. The event does not fire often so I think this is reasonable.

### DIFF
--- a/renderer/main.js
+++ b/renderer/main.js
@@ -91,6 +91,9 @@ function onState (err, _state) {
   window.addEventListener('focus', onFocus)
   window.addEventListener('blur', onBlur)
 
+  // ...window visibility state.
+  document.addEventListener('webkitvisibilitychange', onVisibilityChange)
+
   sound.play('STARTUP')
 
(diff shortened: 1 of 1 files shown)


feross:
Thanks @mathiasvr, I think this is reasonable!

Also, I added you as a collaborator to WebTorrent Desktop. 👍

mathiasvr:
Thanks @feross! 😄

PR #636, opened 2016-06-13
fix modal inconsistencies

mathiasvr:
I noticed that the modals had some minor inconsistencies. Do these changes seem alright?


feross:
Looks good. Only other thing I would change is making the default (raised) button go on the right side.

feross:
Fixed that myself :)

Issue #83, opened 2016-03-08
Add webtorrent:// protocol handler


feross:
_Moved into it's own issue from https://github.com/feross/webtorrent-app/issues/54#issuecomment-193571322..._

Since torrent streaming isn't supported by most popular torrent managers; developers and websites could make an integration for this without requiring users to rethink their default torrent configurations.

Websites that share torrent files generally have links such as 'Download Torrent' and 'Open Magnet'. Imagine an extra link saying 'Stream Torrent'. That's what I'd like to see.

**Bonus suggestion**: decide on a general torrent streaming protocol (i.e. `str://<magnet>`) with https://github.com/jaruba/PowderPlayer so that developers can simply add 'Stream Torrent` links that support any torrent app with the protocol!

giuscri:
+1

ngoldman:
Here's some info on that: https://github.com/electron-userland/electron-packager/issues/59

See also @jden's https://www.npmjs.com/package/register-protocol-handler (currently win32 only)

jden:
Thanks for the mention, @ngoldman - disclaimer, that was kind of experimental and I don't know if it's actually being used anywhere currently, ymmv.
Electron does something that I think works crossplatform for this

Flet:
Probably http://electron.atom.io/docs/v0.36.8/api/protocol/ ?

ngoldman:
Looks like github issue google juice was better than atom docs, thanks for finding that Flet! That looks a lot more up to date.

feross:
@Flet @ungoldman The `protocol` module only works for the URL requests inside Electron. There are OS-specific ways to make this work. Here's how I did it for OS X: https://github.com/feross/webtorrent-app/pull/164/files#diff-9109eccc4c03527f26f8ead0327367e8R175

feross:
Opened an issue to ask Electron to make this easier: https://github.com/atom/electron/issues/4857

feross:
@jden Your module is nice and simple, but it has a native dependency to link to the .NET API, and it fails to install on non-Windows platforms, obviously.

I'm going to try using [`winreg`](https://www.npmjs.com/package/winreg) to modify the registry. I've used this package before for `webtorrent-cli`, to locate the path to launch VLC. `winreg` is great because it just calls out to the `REG` command line program with `child_process`, so it's pure JavaScript. No native deps.

feross:
[...]

PR #640, opened 2016-06-14
new protocol handler: stream-magnet

anonymlol:
Related to  #83 

Only tested on Windows.

### DIFF
--- a/main/handlers.js
+++ b/main/handlers.js
@@ -36,7 +36,7 @@ function installDarwin () {
 
   // On OS X, only protocols that are listed in `Info.plist` can be set as the
   // default handler at runtime.
-  app.setAsDefaultProtocolClient('magnet')
+  app.setAsDefaultProtocolClient('stream-magnet')
 
   // File handlers are defined in `Info.plist`.
 }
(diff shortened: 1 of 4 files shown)


renderer/main.js
```diff
@@ -549,7 +549,7 @@ function onOpen (files) {
 function isTorrent (file) {
   var name = typeof file === 'string' ? file : file.name
   var isTorrentFile = path.extname(name).toLowerCase() === '.torrent'
-  var isMagnet = typeof file === 'string' && /^magnet:/.test(file)
+  var isMagnet = typeof file === 'string' && /^magnet:/.test(file) || /^stream-magnet:/.test(file)
```
feross: I would put parens around the magnet regexes for clarity -- I can never remember the order of ops.

main/handlers.js
```diff
@@ -36,7 +36,7 @@ function installDarwin () {
 
   // On OS X, only protocols that are listed in `Info.plist` can be set as the
   // default handler at runtime.
-  app.setAsDefaultProtocolClient('magnet')
+  app.setAsDefaultProtocolClient('stream-magnet')
```
feross: This shouldn't remove magnet, since we support both in Windows/Linux.

feross:
LGTM, with a couple of exceptions inline. I'll fix them up after merge.

Thanks for the contribution!

feross:
I also had to add support for `stream-magnet` to `parse-torrent`

renderer/main.js
Anmo: This also could be like `var isMagnet = typeof file === 'string' && /^(stream-)?magnet:/.test(file)`

PR #641, opened 2016-06-14
Remove torrent data support

mathiasvr:
Support for moving torrent data to trash, and context menu items as proposed in #560.

### DIFF
--- a/renderer/main.js
+++ b/renderer/main.js
@@ -1096,8 +1096,13 @@ function toggleTorrent (infoHash) {
 }
 
 // TODO: use torrentKey, not infoHash
-function deleteTorrent (infoHash) {
+function deleteTorrent (infoHash, deleteData) {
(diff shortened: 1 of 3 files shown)


dcposch:
This doesn't have the confirmation dialogs, but I think that's fine, we can add those later. LGTM

renderer/main.js
```diff
@@ -1124,6 +1129,20 @@ function openTorrentContextMenu (infoHash) {
   var torrentSummary = getTorrentSummary(infoHash)
   var menu = new electron.remote.Menu()
 
+  menu.append(new electron.remote.MenuItem({
```
dcposch: Nitpick, I would probably put these two menu items at the bottom and keep Show In Finder on top

dcposch:
Thanks!

feross:
Just getting a chance to review this now. Excellent PR! 👍

PR #644, opened 2016-06-15
Support .wmv video via VLC

### LINKED ISSUE #625: Support .wmv format
We should add .wmv to the list of video formats so that it will prompt users to open it in VLC. Right now it's not treated as a video file.

dcposch:


dcposch:
@feross Node Security says this PR introduced a vulnerability (seems unlikely... check out the diff)

When I click link I get a 403

feross:
@dcposch That's weird. The node security link should be publicly visible, IMO. @evilpacket -- can we make the node security links publicly visible for open source repos, sort of how travis and greenkeeper work?

@dcposch I opened an issue on the responsible repo: https://github.com/thibauts/node-castv2/issues/28 The issue doesn't affect WebTorrent.

dcposch:
@evilpacket it would also be awesome to have a Sign Up With Github button as an alternative to username + password

evilpacket:
Yes, that's being worked on right now as I type this. It was an unintended side effect of changing our pricing model last minute ;) sorry for the trouble cc @nlf

nlf:
status for public repos will be visible to everyone later this week :)

as for a 'sign up with github' that would require more changes, i'll make a note of it though and we'll consider it for sure

feross:
@evilpacket @nlf You guys rock 👍

PR #647, opened 2016-06-17
Make auto-resize match video aspect ratio

### LINKED ISSUE #565: Auto-resize does not match video aspect ratio (Windows)

0.5.1

Windows 10

**What did you do?**

Start the Sintel torrent. The window automatically resizes the match the video size.

The aspect ratio should match the video.

**What actually happened?**

The aspect ratio was slightly off, as if the header size is not being taken into account.

Observe the black bars on the sides:

screen shot 2016-05-22 at 22 10 40

Goldob:
Fixes #565

### DIFF
--- a/main/windows/main.js
+++ b/main/windows/main.js
@@ -5,6 +5,7 @@ var main = module.exports = {
   send,
   setAspectRatio,
   setBounds,
+  setContentSize,
   setProgress,
   setTitle,
   show,
@@ -100,6 +101,26 @@ function send (...args) {
 function setAspectRatio (aspectRatio) {
   if (!main.win) return
   main.win.setAspectRatio(aspectRatio)
+}
+
+function setContentSize (size) {
+  // Do nothing in fullscreen
+  if (!main.win || main.win.isFullScreen()) {
+    log('setContentSize: not setting content size because we\'re in full screen')
+    return
+  }
+
+  log('setContentSize: setting content size to ' + JSON.stringify(size))
+  var pos = main.win.getPosition()
+  var contentSize = main.win.getContentSize()
+  main.win.setContentSize(size.width, size.height, true)
+
+  // Keep the window centered at the same position as before resize
+  var x = Math.round(pos[0] + (contentSize[0] - size.width) / 2)
+  var y = Math.round(pos[1] + (contentSize[1] - size.height) / 2)
+
+  log('setContentSize: position changed to ' + JSON.stringify({ x, y }))
+  main.win.setPosition(x, y, false)
 }
 
 /**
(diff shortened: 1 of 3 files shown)


main/windows/main.js
```diff
@@ -102,6 +103,26 @@ function setAspectRatio (aspectRatio) {
...
+  var pos = main.win.getPosition()
+  var contentSize = main.win.getContentSize()
+  main.win.setContentSize(size.width, size.height, true)
+
+  // Keep the window centered at the same position as before resize
+  var x = Math.round(pos[0] + (contentSize[0] - size.width) / 2)
+  var y = Math.round(pos[1] + (contentSize[1] - size.height) / 2)
+
+  log('setContentSize: position changed to ' + JSON.stringify({ x, y }))
+  main.win.setPosition(x, y, false)
```
Goldob: We need to call this function because `BrowserWindow.setContentSize` keeps windows origin at the same position. This however poses a question if subsequently calling two functions modifying window state doesn't disturb resize animation on OS X.

Please confirm.

mathiasvr: On OS X the resize animation is not so smooth anymore. The window is stretched to the right and then jumps back to center the window on screen. Without PR the window is stretched and moved continuously without the jump. Hope it makes sense.

demoneaux:
Possibly we could only use `setContentSize` on non-OS X environments. Though I’m not sure how animations for other environments are affected.

Goldob:
Yes, that's what I was thinking. Other environments shouldn't pose problems, because the `animate` argument [concerns only OS X](http://electron.atom.io/docs/api/browser-window/#winsetcontentsizewidth-height-animate).

We could test it on Linux just to be sure, however. I'll do that a bit later.

main/windows/main.js
Goldob: Thanks for the info. Did auto-resize work properly before the changes? Or were there black bars, like those on Windows?

mathiasvr: Yes it works fine, no black bars.

main/windows/main.js
```diff
@@ -102,6 +103,26 @@ function setAspectRatio (aspectRatio) {
   main.win.setAspectRatio(aspectRatio)
 }
 
+function setContentSize (size) {
```
dcposch: Looks like we still need `setBounds` to restore window size and position after exiting from a video, but we can probably simplify it. It no longer needs to do autocentering, where `x` and `y` are optional arguments.

dcposch:
LGTM

demoneaux:
I believe this is closed by #772. Thanks for your work @Goldob!

PR #650, opened 2016-06-18
Exit media when user closes window

### LINKED ISSUE #649: Playback continues after minimize (close button)
latest
windows 8.1
**What did you do?**
press close button during playback
webtorrent to minimize and playback to pause
**What actually happened?**
playback continues

Goldob:
Fixes #649

### DIFF
--- a/main/windows/main.js
+++ b/main/windows/main.js
@@ -75,6 +75,7 @@ function init () {
     } else if (!app.isQuitting) {
       e.preventDefault()
       win.hide()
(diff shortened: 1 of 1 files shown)


feross:
Thanks for the PR -- but I think the cleaner way to solve it is to change the `win.hide()` call to `hide(), as described here: https://github.com/feross/webtorrent-desktop/commit/62cb304971cb867e5923044df9b7afa2c5f35e78#commitcomment-17965133

PR #655, opened 2016-06-20
Telemetry

dcposch:
Logs system info like # of cores and amount of RAM, OS version, and some usage stats.

I've tried to keep it light. It should never log anything personally identifying or otherwise private.

It makes a single POST request a few seconds after the app loads.

@watson @feross

### DIFF
--- a/bin/check-deps.js
+++ b/bin/check-deps.js
@@ -3,7 +3,48 @@
 var fs = require('fs')
 var cp = require('child_process')
 
-var BUILT_IN_DEPS = ['child_process', 'electron', 'fs', 'os', 'path']
+// We can't use `builtin-modules` here since our TravisCI
+// setup expects this file to run with no dependencies
+var BUILT_IN_NODE_MODULES = [
+  'assert',
+  'buffer',
+  'child_process',
+  'cluster',
+  'console',
+  'constants',
+  'crypto',
+  'dgram',
+  'dns',
+  'domain',
+  'events',
+  'fs',
+  'http',
+  'https',
+  'module',
+  'net',
+  'os',
+  'path',
+  'process',
+  'punycode',
+  'querystring',
+  'readline',
+  'repl',
+  'stream',
+  'string_decoder',
+  'timers',
+  'tls',
+  'tty',
+  'url',
+  'util',
+  'v8',
+  'vm',
+  'zlib'
+]
+
+var BUILT_IN_ELECTRON_MODULES = [ 'electron' ]
+
+var BUILT_IN_DEPS = [].concat(BUILT_IN_NODE_MODULES, BUILT_IN_ELECTRON_MODULES)
+
 var EXECUTABLE_DEPS = ['gh-release', 'standard']
 
 main()
--- a/renderer/main.js
+++ b/renderer/main.js
@@ -15,8 +15,8 @@ var diff = require('virtual-dom/diff')
 var patch = require('virtual-dom/patch')
 
 var config = require('../config')
-var telemetry = require('../telemetry')
 var App = require('./views/app')
+var telemetry = require('./lib/telemetry')
 var errors = require('./lib/errors')
 var sound = require('./lib/sound')
 var State = require('./lib/state')
@@ -96,7 +96,8 @@ function onState (err, _state) {
   document.addEventListener('webkitvisibilitychange', onVisibilityChange)
 
   // Log uncaught JS errors
-  window.addEventListener('error', (err) => telemetry.logUncaughtError('window', err.error), true)
+  window.addEventListener('error',
+    (e) => telemetry.logUncaughtError('window', e.error), true)
 
   // Done! Ideally we want to get here < 500ms after the user clicks the app
   sound.play('STARTUP')
--- a/renderer/main.js
+++ b/renderer/main.js
@@ -95,9 +95,11 @@ function onState (err, _state) {
   // ...window visibility state.
   document.addEventListener('webkitvisibilitychange', onVisibilityChange)
 
-  sound.play('STARTUP')
+  // Log uncaught JS errors
+  window.addEventListener('error', (err) => telemetry.logUncaughtError('window', err.error), true)
 
   // Done! Ideally we want to get here < 500ms after the user clicks the app
+  sound.play('STARTUP')
   console.timeEnd('init')
 }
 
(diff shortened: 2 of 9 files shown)


bin/check-deps.js
```diff
@@ -3,7 +3,46 @@
...
+  'string_decoder',
+  'timers',
+  'tls',
+  'tty',
+  'url',
+  'util',
+  'v8',
+  'vm',
+  'zlib'
+]
```
sindresorhus: https://github.com/sindresorhus/builtin-modules

dcposch:
Example telemetry message, from a 2016 12" Macbook running El Capitan

```
{
  "userID": "65766b64554109ba271c849cd0e51f27b9ef553c2a0d1c6967f7a6869fbcb405",
  "uncaughtErrors": [],
  "playAttempts": {
    "total": 0,
    "success": 0,
    "timeout": 0,
    "error": 0
  },
  "timestamp": "2016-06-21T04:28:37.567Z",
  "localTime": "21:28:37 GMT-0700 (PDT)",
  "screens": [
    {
      "width": 1280,
      "height": 800,
      "scaleFactor": 2
    }
  ],
  "system": {
    "osPlatform": "darwin",
    "osRelease": "Darwin 15.5.0",
    "architecture": "x64",
    "totalMemoryMB": 8192,
    "numCores": 4
  },
  "approxNumTorrents": 4
}
```

dcposch:
I added logging for uncaught exceptions. Tested by running webtorrent-www and webtorrent-desktop on the same laptop and adding a hack to point one at the other. The server appends telemetry reports to a file, and the saved data looks correct.

## Next steps
* Ship it
* Plot the data. We could do this quick and dirty using Google Sheets, or we could make something really sweet with Crossfilter

Flet:
This should be very helpful data!

Would it be prudent let the user know about that data is being collected? 

Ideally they should be able toggle this feature via preferences, but having this stated somewhere is probably a good start. Maybe in via the "Learn more about WebTorrent" link under the Help menu?

dcposch:
We want to be upfront and open about what we're collecting. I'll add a
section to the README. I don't want to complicate the app UI, though.

* We're not collecting personally identifying information
* Lots of common software collects this kind of telemetry, with no prompts
and often with no prefs setting.
* Using a torrent client inherently broadcasts much more sensitive info
anyway---to peers, trackers, and the DHT, which is publically visible.
Specifically, which swarms you're part of.

mathiasvr:
Looks good so far, I think this will be very useful 👍

Maybe the telemetry should be disabled when WebTorrent Desktop is not run in a production environment, or add a property that specifies which environment WebTorrent Desktop is running in, since various (misleading) exceptions might be introduced during development?
Alternatively we could specify whether or not the app has been signed?

dcposch:
@mathiasvr great point. I'll make sure we only collect telemetry from the production app

telemetry.js
```diff
@@ -0,0 +1,146 @@
+// Collects anonymous usage stats and uncaught errors
+// Reports back so that we can improve WebTorrent Desktop
+module.exports = {
+  init,
+  logUncaughtError,
+  logPlayAttempt,
+  getSummary
```
feross: Does this need to be exposed?

telemetry.js
```diff
@@ -0,0 +1,146 @@
...
+// Get the number of torrents, rounded to the nearest power of two
+function getApproxNumTorrents (state) {
+  var exactNum = state.saved.torrents.length
+  if (exactNum === 0) return 0
+  // Otherwise, return 1, 2, 4, 8, etc by rounding in log space
+  var log2 = Math.log(exactNum) / Math.log(2)
+  return 1 << Math.round(log2)
+}
+
+// An uncaught error happened in the main process or one in one of the windows
```
feross: "or in one of the windows"

telemetry.js
```diff
@@ -0,0 +1,146 @@
...
+function logUncaughtError (process, err) {
```
feross: `process` is the name of a global in node.js. Can you rename this so it's clearer?

renderer/webtorrent.js
```diff
@@ -63,6 +63,9 @@ function init () {
 
   ipc.send('ipcReadyWebTorrent')
 
+  window.addEventListener('error', (err) =>
```
feross: This event listener really calls your function with an event object, not an error, so naming it `e` would be clearer.

renderer/main.js
```diff
@@ -94,15 +95,18 @@ function onState (err, _state) {
   // ...window visibility state.
   document.addEventListener('webkitvisibilitychange', onVisibilityChange)
 
-  sound.play('STARTUP')
+  // Log uncaught JS errors
+  window.addEventListener('error', (err) => telemetry.logUncaughtError('window', err.error), true)
```
feross: Why are you setting the third argument to `window.addEventListener` to `true`? Was that intentional? It's setting `useCapture` which is pretty obscure: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

telemetry.js
```diff
@@ -0,0 +1,146 @@
+// Collects anonymous usage stats and uncaught errors
```
feross: This file looks like it's only required from the renderer -- can you move it under `/renderer`?

feross:
LGTM, but see inline comments.

Nice work! I'm really curious to see the percentage of successful video plays.

telemetry.js
dcposch: Fixed

telemetry.js
dcposch: fixed

telemetry.js
dcposch: Renamed the argument to `procName`, the JSON object still says `process`.

It's either 'main' (main process), 'window' (renderer process for the app window), or 'webtorrent' (renderer process for the webtorrent hidden window)

renderer/webtorrent.js
dcposch: fixed

renderer/main.js
dcposch: ya the MDN docs say that you need to capture the error event to get all the errors. errors that happen on an img or video tag etc won't bubble up to the window automatically

dcposch: https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror

telemetry.js
dcposch: done

renderer/main.js
feross: :

Issue #653, opened 2016-06-19
LocationHistory not truly asynchronous


Goldob:
The way it is implemented right now, `LocationHistory.back` doesn't wait for `onbeforeload` to complete. It immediately [pops](https://github.com/feross/webtorrent-desktop/blob/master/renderer/lib/location-history.js#L33) the last page from history, automatically making target page the current one (see: [this function](https://github.com/feross/webtorrent-desktop/blob/master/renderer/lib/location-history.js#L14)). That can cause renderer to [open](https://github.com/feross/webtorrent-desktop/blob/master/renderer/lib/location-history.js#L14) the previous page before it is ready (and throw some errors, which happened to me during development).

Also, I think we shouldn't render the page after we start unloading it. Look how `LocationHistory._go` is implemented. We call [_unload](https://github.com/feross/webtorrent-desktop/blob/master/renderer/lib/location-history.js#L86), but the page stays as current until we have finished unloading. This might potentially cause trouble if renderer uses some variables which `onbeforeunload` clears. Maybe we could add `onafterunload` property for that case.

I see two possible solutions here:

- whenever changing page, first load the new page, then mark it as current, then unload the old page
- keep the current order (unload, load, change), but display loading page until the transition is complete

My bet is the first option, especially that it allows to maitain the current behavior (like, stay in home screen until we are ready to play the selected video), but I leave the decision to you, guys.

feross:
All good points, fixed in https://www.npmjs.com/package/location-history

PR #656, opened 2016-06-20
Test suite and fix proposal for LocationHistory

Goldob:
I set up a test suite to illustrate issue #653. See travis output.

This PR could also be used for fix development once the fate of `LocationHistory` is decided.


Goldob:
The latest commits includes fix proposal. It implements the solution I suggested in the issue.

> whenever changing page, first load the new page, then mark it as current, then unload the old page

The above includes changing page via either `LocationHistory.go`, `back` or `forward` method. Also, `page.onbeforeunload` poperty is replaced with `page.onafterunload` and works accordingly to the name.

Goldob:
Moved to #687

PR #663, opened 2016-06-23
Seeding: sort files by path

dcposch:
Fixes a bug where you could create duplicate torrents by adding the same folder multiple times, because the file order & therefore the infohash was nondeterministic

### DIFF
--- a/renderer/main.js
+++ b/renderer/main.js
@@ -756,6 +756,7 @@ function findFilesRecursive (paths, cb) {
       findFilesRecursive([path], function (fileObjs) {
         ret = ret.concat(fileObjs)
         if (++numComplete === paths.length) {
+          ret.sort((a, b) => a.path < b.path ? -1 : a.path > b.path)
           cb(ret)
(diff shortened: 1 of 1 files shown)


renderer/main.js
```diff
@@ -756,6 +756,7 @@ function findFilesRecursive (paths, cb) {
       findFilesRecursive([path], function (fileObjs) {
         ret = ret.concat(fileObjs)
         if (++numComplete === paths.length) {
+          ret.sort((a, b) => a.path < b.path ? -1 : a.path > b.path)
```
amio: Why not just `ret.sort((a, b) => a.path < b.path ? -1 : 1)`?

dcposch: it has to return 0 when a.path === b.path

amio: I just thought there won't be two file with same path 😌

dcposch: Feels a little janky but yeah, you're def right

Issue #483, opened 2016-05-12
VLC Play Support


saru95:
Navigating through the OSX app I couldnt find an option to play the media through VLC . This comes in handy when the audio is too low or video is in some weird format .
Is this being worked upon ? Would it be fine if I work on this ?

karimdag:
+1

rom1504:
see #368 and https://github.com/feross/webtorrent-desktop/pull/424

feross:
Yep, implemented. Will be released very soon.

aurimus:
The menu in VLC does not appear except the context menu.

PR #666, opened 2016-06-23
Windows Thumbnail Bar

dcposch:
* While in the player view, show a play/pause toggle in the thumbnail

This is just a squashed and rebased version of #483 

Thanks @gediminaspetrikas!


main/menu.js
```diff
@@ -33,6 +34,8 @@ function onPlayerClose () {
   getMenuItem('Increase Speed').enabled = false
   getMenuItem('Decrease Speed').enabled = false
   getMenuItem('Add Subtitles File...').enabled = false
+
+  thumbnail.showPlayerThumbnailBar()
```
feross: This really isn't the right place for this. It's not related to the menu. I'll move it.

PR #668, opened 2016-06-23
Only send telemetry from production

dcposch:
Also fix a bug where Remove Data File doesn't work when the first file in the torrent is in a subfolder.

### DIFF
--- a/renderer/lib/telemetry.js
+++ b/renderer/lib/telemetry.js
@@ -30,7 +30,13 @@ function init (state) {
   telemetry.system = getSystemInfo()
   telemetry.approxNumTorrents = getApproxNumTorrents(state)
 
-  postToServer(telemetry)
+  if (config.IS_PRODUCTION) {
+    postToServer()
+  } else {
+    // Development: telemetry used only for local debugging
+    // Empty uncaught errors, etc at the start of every run
+    reset()
+  }
 }
 
 function reset () {


renderer/lib/telemetry.js
```diff
@@ -17,6 +17,15 @@ const config = require('../../config')
 var telemetry
 
 function init (state) {
+  // Running the app in development?
+  // Make a dummy telemetry object, don't modify config.json, don't post
+  if (!config.IS_PRODUCTION) {
+    telemetry = createSummary()
```
feross: Minor nit: I think this doesn't set the other telemetry properties like `timestamp` which will make it harder to debug those in dev. I'll fix up post-merge.

feross: Nevermind -- can't merge because of the issue with the Remove Data File change.

renderer/main.js
```diff
@@ -1213,11 +1213,7 @@ function openTorrentContextMenu (infoHash) {
 }
 
 function getTorrentPath (torrentSummary) {
-  var itemPath = path.join(torrentSummary.path, torrentSummary.files[0].path)
-  if (torrentSummary.files.length > 1) {
-    itemPath = path.dirname(itemPath)
-  }
-  return itemPath
+  return path.join(torrentSummary.path, torrentSummary.name)
```
feross: Is this the bug?

1. create a folder that contains a single file
2. drop the folder onto WTD
3. right click -> "remove data file"
4. File is deleted, but folders remain

This is unfortunately not easily fixable. The torrent spec for single file torrents doesn't allow specifying a path. See [the spec here](https://wiki.theory.org/BitTorrentSpecification#Info_in_Single_File_Mode). If it's a single file torrent, you can just have a name, no path.

We could fix by storing the original folder the user dropped, and deleting that when they delete a single file torrent, but this isn't ideal because it's complicated and other files might have been added to the folder in the meantime.

mathiasvr: I don't think the bug is about single file torrents.
In a multi file torrent, when the first file `files[0]` is within a subfolder, the resolved path will be that of the subfolder and not root torrent folder.

dcposch: I'll move that commit to its own PR.

--

Currently the Remove Data Files is pretty broken. If you drag, say, two files from your Downloads folder onto WebTorrent Desktop, create a torrent containing just those two, and then click Remove Data Files it'll remove your entire downloads folder (!)

I think the fix is to delete each file in the torrent individually

renderer/lib/telemetry.js
dcposch: Fixed

renderer/main.js
mathiasvr: That's pretty bad. Just tried dragging two files in, I noticed that the file paths was like `Downloads/file1`, and the torrent path was `/Users/mathias`. It would make more sense to make the torrent path `/Users/mathias/Downloads` and the file paths simply `file1`. Does this make sense?

This way I think it would be easier to avoid deleting the dowloads folder.

mathiasvr: Just realized this probably is because multi file torrents must be within a folder path?
I think deleting each file individually sound like the best solution, but what about the empty folders remaining? If we just remove them when they are empty the Downloads folder might still be removed?

dcposch:
this is a small change now, with the `getTorrentPath` fixes moved to their own separate PR. merging

PR #669, opened 2016-06-24
Open in VLC

codealchemist:
Added new checkbox menu item: _Playback / Open in VLC_.
While _Open in VLC_ is checked all video playback will be sent to VLC.
This is really useful if you prefer playing video in VLC instead of the embedded player.

### DIFF
--- a/config.js
+++ b/config.js
@@ -75,8 +75,7 @@ module.exports = {
   WINDOW_WEBTORRENT: 'file://' + path.join(__dirname, 'renderer', 'webtorrent.html'),
 
   WINDOW_MIN_HEIGHT: 38 + (120 * 2), // header height + 2 torrents
-  WINDOW_MIN_WIDTH: 425,
-  OPEN_IN_VLC: false
+  WINDOW_MIN_WIDTH: 425
 }
 
 function getConfigPath () {
--- a/config.js
+++ b/config.js
@@ -75,7 +75,8 @@ module.exports = {
   WINDOW_WEBTORRENT: 'file://' + path.join(__dirname, 'renderer', 'webtorrent.html'),
 
   WINDOW_MIN_HEIGHT: 38 + (120 * 2), // header height + 2 torrents
-  WINDOW_MIN_WIDTH: 425
+  WINDOW_MIN_WIDTH: 425,
+  OPEN_IN_VLC: false
 }
 
 function getConfigPath () {
--- a/main/menu.js
+++ b/main/menu.js
@@ -59,6 +59,10 @@ function onToggleFullScreen (flag) {
   getMenuItem('Full Screen').checked = flag
 }
 
+function onToggleOpenInVlc (flag) {
+  getMenuItem('Open in VLC').checked = flag
+}
+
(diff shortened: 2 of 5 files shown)


anonymlol:
Tested it, works nice, but resets after quitting and re-opening webtorrent-desktop.

codealchemist:
Good catch @anonymlol!
I'll look into it, thx!

mathiasvr:
I don't think this feature belongs in the `Playback` menu because these items should be disabled when the player is inactive. I suggest making it a setting in the preferences window instead.

codealchemist:
That's a good observation @mathiasvr, thx!
This is the playback menu from VLC:
image

They don't follow the _"disable all playback items when player is inactive"_ approach.
But, having the _"Play in VLC"_ option in settings sounds good too.
I'll implement it in another branch so we can choose the best approach.
Thanks!

codealchemist:
Same option added in Preferences:
feross/webtorrent-desktop#674

Cheers!

config.js
```diff
@@ -75,7 +75,8 @@ module.exports = {
   WINDOW_WEBTORRENT: 'file://' + path.join(__dirname, 'renderer', 'webtorrent.html'),
 
   WINDOW_MIN_HEIGHT: 38 + (120 * 2), // header height + 2 torrents
-  WINDOW_MIN_WIDTH: 425
+  WINDOW_MIN_WIDTH: 425,
+  OPEN_IN_VLC: false
```
dcposch: Doesn't look like this is used anywhere?

main/menu.js
```diff
@@ -17,12 +17,32 @@ var dialog = require('./dialog')
...
+var State = require('../renderer/lib/state')
 
-var menu
+var menu, state
 
 function init () {
   menu = electron.Menu.buildFromTemplate(getMenuTemplate())
   electron.Menu.setApplicationMenu(menu)
+
+  State.load(onState)
```
dcposch: We def don't want to load `config.json` in two different places. 

Does "Open in VLC" have to be a configuration pref? What if we just added it to the right-click menu in the torrent list

anonymlol:
We should be able to select VLC as default player. This would be especially good in combination with autoplay for stream links (issue #671).

main/menu.js
anonymlol: We should be able to select VLC as default player. This would be especially good in combination with autoplay for stream links (issue #671).

config.js
codealchemist: Good catch, thx!

main/menu.js
codealchemist: @dcposch I you refer to:
`var config = require('../config')`
We're already loading it in several files.

If you meant `State`, I also felt it wasn't the best approach to load state more than once.
I'd need to spend more time looking into that.
Maybe we can inject it where's needed somehow, or require it using the dispatcher.

codealchemist:
Probably now we have a better way of **using state in the main process**.
What do you guys think?

dcposch:
duplicates #674

PR #670, opened 2016-06-24
Clean up thumbar (thumbnail) code

feross:
Misc code cleanup

- rename thumbnail method names for succinctness
- Get rid of 'updateThumbnailBar' event -- use existing events
- Get rid of 'blockPowerSave' and 'unblockPowerSave' events -- use a
new combined 'onPlayerPlay' and 'onPlayerPause' events which apply to
power save and updating the thumbbar
- Consistent naming for enable/disable methods


renderer/main.js
```diff
@@ -1106,8 +1104,6 @@ function closePlayer (cb) {
   // Tell the WebTorrent process to kill the torrent-to-HTTP server
   ipcRenderer.send('wt-stop-server')
 
-  // Tell the OS we're no longer playing media, laptops allowed to sleep again
-  ipcRenderer.send('unblockPowerSave')
```
dcposch: I remember seeing this earlier (two IPCs, one for unblockPowerSave and one for onPlayerClose) and thinking it could be cleaner 👍

dcposch:
LGTM

main/thumbar.js
```diff
@@ -0,0 +1,54 @@
...
+ * Show the Windows thumbnail toolbar buttons.
+ */
+function enable () {
+  update(false)
+}
+
+/**
+ * Hide the Windows thumbnail toolbar buttons.
+ */
+function disable () {
```
Goldob: I think the function's name is misleading, since it's supposed to be called only on player close. If we call that function in some other context and then pause or resume playback, the thumbar will reappear.

PR #672, opened 2016-06-24
Fix state.playing.jumpToTime behavior

Goldob:
Previously, `state.playing.jumpToTime = 0` didn't do anything. This commit fixes that.

### DIFF
--- a/renderer/views/player.js
+++ b/renderer/views/player.js
@@ -43,7 +43,7 @@ function renderMedia (state) {
       mediaElement.play()
     }
     // When the user clicks or drags on the progress bar, jump to that position
-    if (state.playing.jumpToTime) {
+    if (state.playing.jumpToTime != null) {
       mediaElement.currentTime = state.playing.jumpToTime
       state.playing.jumpToTime = null
     }


renderer/views/player.js
```diff
@@ -43,7 +43,7 @@ function renderMedia (state) {
       mediaElement.play()
     }
     // When the user clicks or drags on the progress bar, jump to that position
-    if (state.playing.jumpToTime) {
+    if (state.playing.jumpToTime != null) {
```
dcposch: I guess the single equals is a fancy way to check for null or undefined. Standard allows it so 👍

dcposch:
LGTM

renderer/views/player.js
Goldob: I found it on [Stack Overflow](http://stackoverflow.com/a/15992131/4132790) and liked it better than `jumpToTime === null || jumpToTime === undefined` or anything close to that. Brief investigation confirmed this should work as desired.

mathiasvr: Yes this is quite common and actually more reliable since `undefined` is not a reserved word in javascript.

feross: Yep `!= null` is fine by `standard`. Useful shorthand.

Issue #664, opened 2016-06-23
Modules/plugins etc. environment


iadj:
It'd be awesome if for `1.0` there's some sort of an API environment that allows us to easily write modules and plug-ins for `webtorrent-desktop`.

dcposch:
we might consider that after 1.0. for now we're prioritizing reliability, performance, and a minimalist UX

if we had this feature today, what would you use it for?

amio:
Maybe something like #631 , it's better to be a plugin rather than built-in.

dcposch:
#631 is two totally different things. First, the title and description make it sound like it's about live streaming. I think P2P live streaming is a cool idea, but it should be its own separate app.
 
Second, #631 suggests a kind of discovery or feed feature, where you could find new torrents within WebTorrent Desktop. If we ever add that I think it'll just be part of the main app.

As far as I can tell, none of the popular torrent clients have a plugin system: uTorrent, Transmission, etc. And we want to be clean and minimalist compared to those.

amio:
+1 for clean and minimalist
and I mean the second one of #631 😅 

When I imagine the plugin system, I think it might be:

1. Some UI hooks
1. Expose some api

and of course, it should be under a condition that:

1. After 1.0
1. Clean & minimalist

Maybe there's a way to achieve it, anyway, we never thought about a torrent client in the browser while it's true now 😄

feross:
I can see a plugin system being used for things like:

- search
- more casting options
- fancier subtitles

But this is def a post 1.0 feature IMO.

PR #677, opened 2016-06-26
Auto download subtitles

codealchemist:
Added subtitles autoloading.
Uses OpenSubtitles API.
All available subtitles for configured language are downloaded into the _subs_ folder on the torrent path.
Then, one after one they are loaded into WebTorrent.
It's a _brute force_ approach to subtitle loading.

**TODO:** add subtitles preferences with language selector and total amount of downloaded subtitles limit.


codealchemist:
Another todo: avoid re-downloading existing subs.

codealchemist:
I've found issues with single video files.
Fixing that.

codealchemist:
Another TODO: add app notifications for errors.

mathiasvr:
I also thought about adding a subtitle downloader, but I think it would make more sense as a `plugin` in relation to #664. However if many people is interested in this I guess it would make sense to include it by default.

Also I edited the title, since I initially thought this was about loading embedded subtitles (like in mkv files).

codealchemist:
Thanks for the title adjustments @mathiasvr!
If we can create a plugin repository for WebTorrent that would be **awesome**. 
I'm thinking about something like Sublime Text's _Package Control_, which is integrated into the app.
I'd like to help with the subtitles downloader either way.
Whatever the community and the project maintainers prefer.

dcposch:
i don't like hitting a central server every time a file is opened. not ideal for privacy or performance.

what about checking OpenSubtitles only after the user clicks the CC button?

currently, that pops open an Open File dialog

we could make it show a menu instead: Open Subtitles File, Search OpenSubtitles. If the user clicks that second option, it would show a loading spinner, then a list of available languages. they can select one to download and use that subtitle track.

--

to be honest, i'm not sure we need this. i don't know any video player that has this feature. even VLC supports it only via a third-party plugin called VLsub.

codealchemist:
@dcposch, the latest code on the PR allows *enabling / disabling* automatic subtitles downloading from *preferences*.
I still think we should not download already downloaded subtitles, that's not terrible but it's not good enough.

But in general it's a cool and useful feature that might be appreciated by anyone who uses subtitles.
It makes it *tremendously easy* to get subtitles, even easier than with VLSub where you need to make several clicks and write stuff manually to identify the video file _(we can't just use hashes when streaming since the file isn't downloaded yet and can't be hashed)_.

That being said, I really like the idea of having a plugin repository integrated in the app.
And I'm looking forward to help with that if I can and if that's something wanted.
:)

feross:
Thanks for the PR, but we're not going to add this feature to WebTorrent Desktop, because the legality of OpenSubtitles is not clear to us.

Once there's a plugin system, I suppose you could build this as a plugin. But it's not something the WebTorrent team can endorse.

codealchemist:
I see. Thanks for the explanation @feross!
:)

feross:
Sure thing. 👍

janat08:
There're some other players that use this that are open source, likely even copyleft. I don't know which one I use that has it. Perhaps ask them about it.

janat08:
SMPlayer has it and is GNU.

janat08:
I say if you let user decide what and if he wants the subtitles choosing "find subtitles" option, then it becomes his responsibility what happens right?

PR #681, opened 2016-06-27
v0.8.1

dcposch:
Tested @Goldob's DLNA bugfix by installing Kodi (aka XMBC). DLNA doesn't crash anymore, and I can stream from WebTorrent Desktop to Kodi.

Tested `stream-magnet://` using Sintel.


dcposch:
@feross

Issue #531, opened 2016-05-17
Option to use a third-party media player


chinabala:
**What version of WebTorrent Desktop?** (See the 'About WebTorrent' menu)
0.4.0
**What operating system and version?**
Mac OS 10.10
**What did you do?**
Audio not supported, switch to VLC
**What did you expect to happen?**
Using VLC with .ass subtitle file
**What actually happened?**
Some reason my VLC don't work on some Chinese .ass file. Wondering if webstorrent-desktop could allow users to change default VLC to other preferred media player?

feross:
Which other media player do you use?

chinabala:
I'm using MPlayerX on Mac. And I guess other users may using different player on their own preference. That'd be really nice if user can go to WebTorrent's Setting page and port stream to supported media player.

feross:
This is implemented in #682

feross:
Fixed by #682, will be released in WebTorrent Desktop v0.12.

Note @chinabala, you're using v0.4 which has a broken auto-updater, so you will need to update manually.

Issue #368, opened 2016-04-09
External players like VLC


diimdeep:
Until media playback will be feature rich, with subtitles, audio track selection etc., might be good to support external players. 
Although manual open file in VLC works just fine.

dcposch:
#424

lewisl9029:
I would personally prefer using the "Open With" dialog to allow users to choose their own video player, rather than hard coding VLC detection and support. Thoughts?

dcposch:
@lewisl9029 sounds reasonable, might implement that in a future version

The great thing about VLC is that it can play all common media formats. If WebTorrent Desktop can't play a file, that means it probably uses the AC3 or DTS codecs or the AVI container format. Either way, most other video players that aren't VLC won't play it either.

rom1504:
mpv and mplayer are pretty good at reading every format too (https://github.com/feross/webtorrent-cli/blob/master/bin/cmd.js#L191)

feross:
Yep, we'll add this in a future version.

feross:
This is done now, thanks to @dcposch

JaKXz:
Is being able to specify `mpv` as the default player going to be an option soon? I'm using v0.7.2 and I don't have VLC installed at all.

feross:
@JaKXz See #682

PR #682, opened 2016-06-27
Custom external media player

mathiasvr:
Adding support for selecting a custom external media player, defaulting to VLC.

Much of it is renaming of VLC specific code to more generic sounding 'external player'.

* Added preference `External Media Player` for specifying path to any program.
* Refactored `ipc.js` and `vlc.js` to `external-player.js` 

Related to issues #368 and #531.


main/external-player.js
```diff
@@ -0,0 +1,65 @@
...
+var windows = require('./windows')
+
+// holds a ChildProcess while we're playing a video in an external player, null otherwise
+var proc
+
+function checkInstall (path, cb) {
+  // check for VLC if external player has not been specified by the user
+  // otherwise assume the player is installed
+  if (path == null) return vlcCommand((err) => cb(!err))
+  process.nextTick(() => cb(true))
```
dcposch: Do we need `process.nextTick`? Could we just call `cb(true)`?

main/external-player.js
```diff
@@ -0,0 +1,65 @@
...
+}
+
+function kill () {
+  if (!proc) return
+  log('Killing external player, pid ' + proc.pid)
+  proc.kill('SIGKILL') // kill -9
+  proc = null
+}
+
+function _spawn (path, args) {
```
dcposch: I'd call this `spawnExternal`. Makes sense that it's not exported. We aren't using underscores for private functions.

main/external-player.js
```diff
@@ -0,0 +1,65 @@
...
+    windows.main.dispatch('exitModal'), 1000)
+
+  proc.on('close', function (code) {
+    clearTimeout(closeModalTimeout)
+    if (!proc) return // Killed
+    log('External player exited with code ', code)
+    if (code === 0) {
+      windows.main.dispatch('backToList')
+    } else {
+      windows.main.dispatch('externalPlayerNotFound')
```
dcposch: A nonzero exit code doesn't necessarily mean it wasn't found. It could also mean the external player ran, but encountered an error, for example an unplayable file.

There's a special exit code for not found (127), but I don't know if that works on all OSs.

renderer/views/preferences.js
```diff
@@ -40,6 +42,27 @@ function renderDownloadDirSelector (state) {
...
+  return renderFileSelector({
+    label: 'External Media Player',
+    description: 'Progam that will be used to play media externally',
+    property: 'externalPlayerPath',
+    options: {
+      title: 'Select media player executable',
+      properties: [ 'openFile' ]
+    }
+  },
+  state.unsaved.prefs.externalPlayerPath || '<VLC>', // TODO: should we get/store vlc path instead?
```
dcposch: It may be confusing to people that they can't set the path back to `<VLC>` after using a different player

renderer/views/player.js
```diff
@@ -269,9 +269,9 @@ function renderCastScreen (state) {
     castIcon = 'tv'
     castType = 'DLNA'
     isCast = true
-  } else if (state.playing.location === 'vlc') {
+  } else if (state.playing.location === 'external') {
     castIcon = 'tv'
-    castType = 'VLC'
+    castType = 'External' // TODO: Should we use the player name instead? see `unsupported-media-modal.js`
```
dcposch: 'External' is fine. If you want, you can extract the filename from the external player path and use that, defaulting to 'VLC' if no external player path is set.

This would show eg. 'QuickTime.app', 'mplayer', or 'VLC' instead of 'External'

dcposch:
LGTM apart from comments

dcposch:
I think one of the biggest things we could do to improve WebTorrent Desktop is to give it codec support comparable to VLC. If we did that, I wonder if we'd still need this. I always hoped "Open in VLC" would be a temporary patch until we could support all the common file types.

anonymlol:
A good external player not only supports more codecs, it also offers higher quality playback and more customisation.

I think this needs a "set as default" checkbox.

main/external-player.js
mathiasvr: It is there to ensure that the callback is always asynchronous. Right now it doesn't really make a difference, but it is generally discouraged to mix sync/async callbacks.

main/external-player.js
mathiasvr: Okay I will rename it. We use underscores in `location-history.js though?

main/external-player.js
mathiasvr: This is what we are using for VLC right now, but let's look into this.

renderer/views/preferences.js
mathiasvr: Yes, there are some differences between selecting the VLC path and using the 'default' behavior. Should we make a way to 'reset' this option?

renderer/views/player.js
mathiasvr: I would like to have a more convenient way to get the player name, compared to how it is done in `unsupported-media-modal.js`. So I guess I will stick with 'External' for now.

feross:
Merged! I rebased the PR so that it would merge cleanly, and also committed this: https://github.com/feross/webtorrent-desktop/commit/2d3673ea33c41eb244c1df3831a96c793c1ba951

feross:
Also, sorry for the delay in getting this in, @mathiasvr!

Btw, @dcposch and I are trying to get as much done as possible in the next few days to help get 1.0 out the door!

mathiasvr:
Nice work guys! 🎉

JaKXz:
I'm guessing this will be in `v1.0` right? [since I'm not seeing it in `v0.11.0`] - really looking forward to it, thanks guys!

feross:
@JaKXz Yeah, it will definitely be in the next release. We'll probably do a v0.12 before 1.0.

PR #687, opened 2016-06-28
LocationHistory rework proposal

Goldob:
Relates with #653.

This is really just #656 after rebase. I decided to set up a new PR because Travis stopped responding to new commits and his output is crucial to illustrate the point (feel free to take a look at the unit tests).

--

With the proposed changes, a procedure for transitioning from `old` to new `page` will work as follows:
1. Call `new.onbeforeload` and wait for callback<sup>1</sup>
2. Mark `new` as current page
3. Call `old.onafterunload`<sup>2</sup>

--

<sup>1</sup> This was not always the case in the old version
<sup>2</sup> I decided it is more suitable than `onbeforeunload`, because when we clean view-related variables, we want to do it **after** we stop rendering the page (or we risk errors otherwise). Also, this eliminates redundant complexity from having two `onbefores` - e.g. what if one has completed successfully, but the other returned an error? Do we just quietly abort the page change? But then the view whose function worked properly doesn't even know it's not being (un)loaded anymore. Maybe we should call the opposing function to notify it? What if it returns an error? You get the point.


Goldob:
@dcposch @mathiasvr

mathiasvr:
LGTM, but why do we need babel?

test/location-history-test.js
```diff
@@ -0,0 +1,176 @@
+/* eslint-env mocha */
```
mathiasvr: We should not need this because we are using standard.js?

renderer/lib/location-history.js
```diff
@@ -12,14 +12,16 @@ LocationHistory.prototype.url = function () {
...
-  return this._history[this._history.length - 1]
+  return this._current
 }
 
 LocationHistory.prototype.go = function (page, cb) {
   if (!cb) cb = noop
   if (this._pending) return cb(null)
 
-  console.log('go', page)
+  if (process.env.NODE_ENV !== 'test') {
```
mathiasvr: Can we avoid adding test related code here?

Goldob: I added this because log littered console on `npm test`. I guess there are two ways to avoid that:
- Don't call log function in unit-testable classes (we can always do that outside)
- Create a log wrapper that only writes to log when not in test environment

More details [here](http://stackoverflow.com/questions/22709882/how-to-suppress-application-logging-messages-from-a-node-js-application-when-run).

Also, if for some reason we don't want tests, we can just cherry-pick `ab32d04` commit. It doesn't  include any test code.

test/location-history-test.js
Goldob: [Yes](https://github.com/feross/standard/blame/master/README.md#L323).

Goldob:
> LGTM, but why do we need babel?
> \- @mathiasvr

Actually, we don't. Thanks for pointing that out.

Mocha threw errors on legit code and I thought it had some problems with ES6. However, as it turns out, adding `--use_strict` flag to `mocha` is enough to fix that.

renderer/lib/location-history.js
mathiasvr: I think we should keep the tests. On second thought since this is the only place we do this, I wouldn't mind keeping it in, we can always change it later.

dcposch:
this looks very reasonable to me. thanks for adding tests for LocationHistory---they look thorough!

webtorrent desktop hasn't had any tests so far. that was fine for prototyping. now that the app has good number of users (and features) i think it's worth the extra effort to make sure nothing breaks.

two questions--not for this PR specifically but for WebTorrent Desktop in general
* do we want unit tests or only integration tests
* do we want mocha (describe...it style) or tape

@feross thoughts?

keeping the build simple is great. no babel ftw.

bin/check-deps.js
```diff
@@ -45,7 +45,13 @@ var BUILT_IN_ELECTRON_MODULES = [ 'electron' ]
 
 var BUILT_IN_DEPS = [].concat(BUILT_IN_NODE_MODULES, BUILT_IN_ELECTRON_MODULES)
 
-var EXECUTABLE_DEPS = ['gh-release', 'standard']
+var EXECUTABLE_DEPS = [
+  'gh-release',
+  'standard',
+  'mocha',
+  'babel-core',
+  'babel-preset-es2015'
```
mathiasvr: You forgot these 😉

Goldob: Ooops! Thanks for that.

Goldob:
Travis clearly doesn't like me. Yet another PR of mine to which he stopped responding.

.travis.yml
```diff
@@ -1,4 +1,4 @@
 language: node_js
 node_js:
   - 'node'
-install: npm install standard
+install: npm install standard mocha
```
feross: I don't prefer `mocha` because it pollutes the global namespace.

feross:
@Goldob Great changes! This is good work. Thanks for doing this. I prefer `tape` to `mocha` because it's a better citizen. It doesn't leak a bunch of globals and force you to write code that assumes they'll just be there.

@dcposch I don't think we should have any unit tests in WebTorrent Desktop. This app should be mostly glue code. That can be tested best with a few sanity-check integration tests.

If there's a file that feels like it needs unit tests, then it should probably just be it's own npm package with it's own README, unit tests, etc. `LocationHistory` should definitely be it's own npm package. Probably, torrent-poster and a few others in `lib/` as well.

I'm going to pull out `location-history` into it's own package today.

renderer/lib/location-history.js
```diff
@@ -28,20 +30,23 @@ LocationHistory.prototype.go = function (page, cb) {
...
+    self._forward.push(current)
+    self._current = previous
+    self._unload(current)
+    cb(null)
   }
 }
 
 LocationHistory.prototype.hasBack = function () {
-  return this._history.length > 1
+  return this._back.length
```
feross: This should return a boolean, so `this._history.length > 0` is better.

feross:
https://www.npmjs.com/package/location-history

feross:
New PR here: https://github.com/feross/webtorrent-desktop/pull/748

PR #693, opened 2016-06-30
Refactor main.js

dcposch:
* Simplify startup
* Split out controllers that handle dispatch(...) events. So far we have a MediaController (for local media playback events) and UpdateController (for the new version available UI).

The goal is to make `main.js` small and readable. Currently it's a 1600 line file that does all kinds of things.

### DIFF
--- a/renderer/controllers/media-controller.js
+++ b/renderer/controllers/media-controller.js
@@ -1,0 +1,56 @@
+const electron = require('electron')
+
+const ipcRenderer = electron.ipcRenderer
+
+// Controls local play back: the <video>/<audio> tag and VLC
+// Does not control remote casting (Chromecast etc)
+module.exports = class MediaController {
+  constructor (state) {
+    this.state = state
+  }
+
+  mediaSuccess () {
+    this.state.playing.result = 'success'
+  }
+
+  mediaStalled () {
+    this.state.playing.isStalled = true
+  }
+
+  mediaError (error) {
+    var state = this.state
+    if (state.location.url() === 'player') {
+      state.playing.result = 'error'
+      state.playing.location = 'error'
+      ipcRenderer.send('checkForVLC')
+      ipcRenderer.once('checkForVLC', function (e, isInstalled) {
+        state.modal = {
+          id: 'unsupported-media-modal',
+          error: error,
+          vlcInstalled: isInstalled
+        }
+      })
+    }
+  }
+
+  mediaTimeUpdate () {
+    this.state.playing.lastTimeUpdate = new Date().getTime()
+    this.state.playing.isStalled = false
+  }
+
+  mediaMouseMoved () {
+    this.state.playing.mouseStationarySince = new Date().getTime()
+  }
+
+  vlcPlay () {
+    ipcRenderer.send('vlcPlay', this.state.server.localURL)
+    this.state.playing.location = 'vlc'
+  }
+
+  vlcNotFound () {
+    var modal = this.state.modal
+    if (modal && modal.id === 'unsupported-media-modal') {
+      modal.vlcNotFound = true
+    }
+  }
+}
--- a/renderer/main.js
+++ b/renderer/main.js
@@ -181,11 +181,13 @@ const dispatchHandlers = {
   'toggleSelectTorrent': (infoHash) => controllers.torrentList.toggleSelectTorrent(infoHash),
   'openTorrentContextMenu': (infoHash) => controllers.torrentList.openTorrentContextMenu(infoHash),
 
-  'startTorrentingSummary': startTorrentingSummary,
+  'startTorrentingSummary': (torrentSummary) =>
+    controllers.torrentList.startTorrentingSummary(torrentSummary),
 
   // Playback
   'playFile': (infoHash, index) => controllers.playback.playFile(infoHash, index),
   'playPause': () => controllers.playback.playPause(),
+  'skip': (time) => controllers.playback.skip(time),
   'skipTo': (time) => controllers.playback.skipTo(time),
   'changePlaybackRate': (dir) => controllers.playback.changePlaybackRate(dir),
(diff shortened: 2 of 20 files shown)


dcposch:
@feross

dcposch:
@feross `main.js` is down to 640 lines now

dcposch:
@feross best place to start for reviewing this PR is probably just to read the new version of `main.js`. I think what happens on startup, what happens when we get a `dispatch()` message, and what happens when we get an IPC message are more clear now.

I think this PR is ready to go. Eventually, want `main.js` to have *only* those three things (coordinating startup, sending dispatch events to the right place, sending IPC messages to the right place)

renderer/controllers/media-controller.js
```diff
@@ -0,0 +1,54 @@
...
+    this.state.playing.isStalled = true
+  }
+
+  mediaError (error) {
+    if (this.state.location.url() === 'player') {
+      this.state.playing.result = 'error'
+      this.state.playing.location = 'error'
+      ipcRenderer.send('checkForVLC')
+      ipcRenderer.once('checkForVLC', function (e, isInstalled) {
+        this.state.modal = {
```
feross: Is this actually the right `this`? We're inside an old-style function.

feross:
Hey, sorry I can't do a deep review this right now while I'm on my break from open source. https://twitter.com/feross/status/747981655504105473

Things to consider:

- How does this increase/reduce startup time?
- Should do a decent "smoke test" and try all the various functionality to make sure this doesn't break anything
- Should probably wait a week after merging this to `master` to actually do a release to shake out any bugs

Feel free to merge without me if you get another LGTM from a collaborator. That's the power of having the commit bit!

dcposch:
@mathiasvr @Goldob

renderer/main.js
```diff
@@ -151,290 +166,122 @@ function updateElectron () {
...
-    State.save(state)
-  }
-  if (action === 'setTitle') {
-    state.window.title = args[0] /* title */
-  }
-  if (action === 'uncaughtError') {
-    telemetry.logUncaughtError(args[0] /* process */, args[1] /* error */)
-  }
+  var handler = dispatchHandlers[action]
+  if (handler) handler.apply(null, args)
```
mathiasvr: Nitpick: ES6 has `handler(...args)`

renderer/views/torrent-list.js
```diff
@@ -153,7 +153,7 @@ function TorrentList (state) {
         <i.button-round.icon.play
```
mathiasvr: @feross renamed this file exactly opposite in https://github.com/feross/webtorrent-desktop/commit/6e240b3fd492481eb6c7fc464468ceb15edca380?

mathiasvr:
LGTM, I did experience increased startup time though (about 300ms).

Goldob:
I confirm increased startup time. In my case it's ~1.5x (1000ms more), but I have a relatively slow PC.

I did a smoke test and detected the following issues:
- Adding a valid torrent file or posting a magnet link results in `Invalid torrnet identifier` error
- `Step Forward` and `Step Backward` commands from `Playback` menu don't do anything
- When I select and deselect some files and then restart the app, the changes are reverted. If you won't be able to reproduce this problem, maybe investigate if it's correlated with shutdown timeout. I know there was such thing before refractor, but I didn't dive to the new code deep enough to check it yet.

The test also included other functionalities, which worked correctly:
- Removing a torrent from list
- Playing the default or specified file from torrent
- Playback navigation (other than mentioned above)
- Views navigation (back/forward)
- VLC playback (missing codecs case)

renderer/views/torrent-list.js
dcposch: ya, i thought it was a bit confusing since the function it exports is still called TorrentList. Then in `app.js`, it used to say `const TorrentList = require('home')`. I think TorrentList is nice and descriptive

dcposch:
@mathiasvr @Goldob thanks for checking it out! 

I think we're loading some of the modules too soon now, instead of doing it lazily. I'll try to get the startup time back down to where it was.

renderer/main.js
dcposch: fixed

dcposch:
Fixes:
* #706 is merged. @Goldob is the boss!
* Save state on exit works again. There was a bug in the `saveState` handler in `main.js`.
* Startup time is fast again. I was requireing the Cast module from playback-controller.js, which means it was no longer being lazy loaded. I changed `lib/cast.js` so that it can be require()d pretty much instantly, and the heavy stuff that takes 300ms all happens in Cast.init() now.

Numbers for a 2016 Macbook on battery power:
* Cast module lazy loaded: ~800ms main process, ~250ms window
* Cast module loaded immediately: ~1100ms main process, ~550ms window

dcposch:
Testing:
- Create a torrent
  - Single file torrent
  - Multi file torrent
  - Start creating a torrent, but then click Cancel
- Delete a torrent
- Select and deselect files. Restart the app. Confirm the changes are saved.
- Play a video
  - Play/pause
  - Scrub
  - Skip forward / backward
  - Change playback speed (**Bug:** negative playback speeds don't work. This is not a problem with this PR tho, confirmed that he same thing happens on master. I'll just remove negative playback speeds in a separate PR)
  - Quit while playing. Sound should stop.
  - Hit ESC while playing. Video should stop, app should return to torrent list, restore window size. (**Bug:** window size restore usually works but is sometimes wonky. Existing bug.)
- Cast
  - Chromecast
  - Airplay
  - DLNA
  - (Technically we should test play/pause, scrub, skip, volume for each of these
- Menu
 - Click thru the menus
 - Click the top bar buttons: < (back), > (forward), and + (add file)

dcposch:
I like @feross suggestion of waiting a few days after the merge before we send out our next release.

feross:
LGTM -- nice work!

PR #702, opened 2016-07-02
Save selected subtitle

### LINKED ISSUE #701: Subtitles have been disabled after "Back" and "Start streaming"
0.8.1
Ubuntu Desktop 16.04
**What did you do?**
I pasted a magnet link and after a few seconds I clicked "Start streaming". I clicked on "CC" and double-clicked on the subtitle file (the video file and the subtitle file did not have the same name). The video file played fine with subtitles. After a few minutes I clicked "Back" to go back to the main screen. Then I finally clicked "Start streaming" again. The video file resumed playing from the point I had left it.
I expected that the subtitles I had previously selected would be showing, since resuming from the last played position suggests that a state is saved for a specific magnet link.
**What actually happened?**
No subtitles were selected and I had to re-select them by clicking on "CC".

Great project by the way! ;)

mathiasvr:
Fixes #701.
The selected subtitle index is stored in the same way as the current time of the playing file. It might be overkill to do this on every render, but I could not see other obvious places to put it, since subtitles can be added/selected in multiple ways.

### DIFF
--- a/renderer/controllers/playback-controller.js
+++ b/renderer/controllers/playback-controller.js
@@ -224,6 +224,11 @@ function openPlayerFromActiveTorrent (state, torrentSummary, index, timeout, cb)
   // if it's video, check for subtitles files that are done downloading
   dispatch('checkForSubtitles')
 
+  // enable previously selected subtitle track
+  if (fileSummary.selectedSubtitle) {
+    dispatch('addSubtitles', [fileSummary.selectedSubtitle], true)
+  }
+
   ipcRenderer.send('wt-start-server', torrentSummary.infoHash, index)
   ipcRenderer.once('wt-server-' + torrentSummary.infoHash, (e, info) => {
     clearTimeout(timeout)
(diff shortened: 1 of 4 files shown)


dcposch:
I don't think this works. The subtitles list can come from two different places: either from the torrent, or added by the user. You can drag `.srt` and `.vtt` files onto the player, or select them from the open file dialog after clicking CC.

The subtitles list can change from one time you play a file to the next, so we can't save an index into that list.

dcposch:
instead of index, you could save the subtitle path. make sure to distinguish whether it's a relative path (for a file inside the same torrent) or an absolute path (for user supplied subtitles)

mathiasvr:
@dcposch You are right, I changed the PR so now it saves the absolute path for the selected subtitle, is that okay?
Also I changed the `addSubtitles` function: if the added file is a duplicate it is still possible to auto-select it. This also makes it possible to re-enable a subtitle by dragging it.

dcposch:
LGTM!

mathiasvr:
Rebased for #693

dcposch:
Thanks!

xnakos:
Good job, guys! ;)

PR #706, opened 2016-07-06
Fixes for dc/refactor branch

Goldob:
Bugfixes for `dc/refactor` branch (#693):
- `Invalid torrent identifier` error when adding new torrent
- `Step Forward` and `Step Backward` commands not working

### DIFF
--- a/renderer/webtorrent.js
+++ b/renderer/webtorrent.js
@@ -14,7 +14,6 @@ var path = require('path')
 var crashReporter = require('../crash-reporter')
 var config = require('../config')
 var torrentPoster = require('./lib/torrent-poster')
-var TorrentSummary = require('./lib/torrent-summary')
 
 // Report when the process crashes
 crashReporter.init()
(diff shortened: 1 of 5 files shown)


renderer/webtorrent.js
```diff
@@ -73,23 +72,20 @@ function init () {
 
 // Starts a given TorrentID, which can be an infohash, magnet URI, etc. Returns WebTorrent object
 // See https://github.com/feross/webtorrent/blob/master/docs/api.md#clientaddtorrentid-opts-function-ontorrent-torrent-
-function startTorrenting (torrentSummary) {
```
Goldob: We can't use `torrentSummary` as parameter here, because we don't have one when adding new torrent. It is created after `wt-infohash` message.

Goldob: We can't use `torrentSummary` as parameter here, because we don't have one when adding new torrent. It is created after `wt-infohash` message.

dcposch:
LGTM. About to cherry pick. Thanks!

x550j:
W dniu środa, 6 lipca 2016 Adam Gotlib <notifications@github.com>
napisał(a):

PR #707, opened 2016-07-08
Improve error logging

mathiasvr:
When I tried to play an unsupported video the uncaught error was `undefined` for some reason.

### DIFF
--- a/renderer/main.js
+++ b/renderer/main.js
@@ -109,7 +109,7 @@ function onState (err, _state) {
 
   // Log uncaught JS errors
   window.addEventListener('error',
-    (e) => telemetry.logUncaughtError('window', e.error), true)
+    (e) => telemetry.logUncaughtError('window', e.error || e.target), true)
 
   // Done! Ideally we want to get here < 500ms after the user clicks the app
   sound.play('STARTUP')
--- a/renderer/lib/telemetry.js
+++ b/renderer/lib/telemetry.js
@@ -123,12 +123,12 @@ function logUncaughtError (procName, err) {
   if (!telemetry) return
 
   var message, stack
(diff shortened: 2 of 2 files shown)


renderer/main.js
```diff
@@ -109,7 +109,7 @@ function onState (err, _state) {
 
   // Log uncaught JS errors
   window.addEventListener('error',
-    (e) => telemetry.logUncaughtError('window', e.error), true)
+    (e) => telemetry.logUncaughtError('window', e.error || e.target), true)
```
dcposch: Shouldn't an ErrorEvent always have `e.error`?
https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent

What kinds of errors are there where we need to use target instead?

dcposch:
Otherwise LGTM

renderer/main.js
mathiasvr: When I play an unsupported video (xvid) I get this exception in the console:
```
Uncaught (in promise) DOMException: Failed to load because no supported source was found.
```
The file and line is reported as `undefined:1`, I don't know if this is a bug in chromium?
No matter how I tried to catch this, the event never had the `error` property, so i added target (the `<video>` element) to at least provide some info about where the error happened, because the error message from the console is not saved anywhere.

We might be able to solve this in another way since I also get the _"Sorry, we can't play that file"_ modal instead of _"Video codec unsupported"_.

I will try to look into this.

mathiasvr: Okay so this seems to happen when the video element does not know how to interpret the file. For example changing the extension of some garbage file to `.avi` will cause this. The `mediaError` function will still be called with an `undefined` error though, if this is okay I can remove `e.target`?

PR #708, opened 2016-07-09
Fix unhandled 'error' dispatch

mathiasvr:
Fix unhandled 'error' dispatch used by the `subtitles-controller` and `playback-controller`, and be consistent in the `torrent-list-controller`.

### DIFF
--- a/renderer/main.js
+++ b/renderer/main.js
@@ -237,7 +237,7 @@ const dispatchHandlers = {
 
   // Everything else
   'onOpen': (files) => onOpen(files),
-  'onError': (err) => onError(err),
+  'error': (err) => onError(err),
   'uncaughtError': (proc, err) => telemetry.logUncaughtError(proc, err),
   'saveState': () => State.save(state)
 }
--- a/renderer/main.js
+++ b/renderer/main.js
@@ -231,13 +231,13 @@ const dispatchHandlers = {
   'forward': () => state.location.forward(),
 
   // Controlling the window
-  'setDimensions': (dimensions) => setDimensions(dimensions),
+  'setDimensions': setDimensions,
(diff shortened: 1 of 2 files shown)


mathiasvr:
Also 'mediaEncrypted' is not handled, and 'openTorrentFile' is not used?

mathiasvr:
Just as a side note, I created this messed up hack of a script to determine unhandled+unused dispatch calls:
```bash
#!/bin/bash

diff <(
  # list dispatch calls
  grep "\(dispatch\|dispatcher\)('" *.js bin/ main/ renderer/ -R |
  grep '.js:' |
  sed -E "s/.*(dispatch|dispatcher)\('([^'\/]*).*/\2/g" |
  sort |
  uniq
) <(
  # list dispatch handlers
  grep "':" renderer/main.js |
  sed "s/.*'\(.*\)':.*/\1/" |
  sort)
```

It would be nice to have this checked during testing, because standard cannot check the dispatch calls.

renderer/main.js
```diff
@@ -231,13 +231,13 @@ const dispatchHandlers = {
   'forward': () => state.location.forward(),
 
   // Controlling the window
-  'setDimensions': (dimensions) => setDimensions(dimensions),
+  'setDimensions': setDimensions,
```
mathiasvr: This was to be consistent with `backToList` and `escapeBack`. Do we want this? If so do we also want to call state functions directly?

dcposch: Either way is fine

dcposch:
LGTM

feross:
@mathiasvr Did you send a PR to check the `dispatcher` calls with your script? I don't recall seeing one.

mathiasvr:
@feross No, I wasn't really satisfied with it.

feross:
@mathiasvr Okay, no worries. Just checking.

PR #711, opened 2016-07-11
Remove torrent file and poster

### LINKED ISSUE #709: Poster and torrent is not deleted upon torrent deletion

Version 0.8.1 (0.95.2)

OS X 10.11

**What did you do?**

screenshot

screen shot 2016-07-10 at 01 44 33

Nothing in these directories

**What actually happened?**

All thumbnails of all my downloaded torrents along with torrents files are kept.

It can lead to:
- Privacy issues on shared computers
- Eating hard drive space slowly over time (80mb here)

mathiasvr:
Fixes #709.

### DIFF
--- a/renderer/controllers/torrent-list-controller.js
+++ b/renderer/controllers/torrent-list-controller.js
@@ -122,10 +122,8 @@ module.exports = class TorrentListController {
       var summary = this.state.saved.torrents[index]
 
       // remove torrent and poster file
-      var torrentFilePath = TorrentSummary.getTorrentPath(summary)
-      var posterFilePath = TorrentSummary.getPosterPath(summary)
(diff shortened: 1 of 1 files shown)


renderer/controllers/torrent-list-controller.js
```diff
@@ -229,6 +236,13 @@ function findFilesRecursive (paths, cb) {
   })
 }
 
+function deleteFile (path) {
+  if (!path) return
+  fs.unlink(path, function (err) {
+    if (err) dispatch('onError', err)
```
dcposch: This should be 'error' instead of 'onError' now. I'll fix it after merging

dcposch:
LGTM

PR #721, opened 2016-07-16
Refactor main.js: down to 450 lines

dcposch:
* Move all the torrent event handlers to their own file `torrent-controller.js`
* Move all the file extension checking to the same place

### DIFF
--- a/renderer/controllers/torrent-controller.js
+++ b/renderer/controllers/torrent-controller.js
@@ -1,0 +1,192 @@
+const path = require('path')
+const ipcRenderer = require('electron').ipcRenderer
+
+const TorrentSummary = require('../lib/torrent-summary')
+const TorrentPlayer = require('../lib/torrent-player')
+const sound = require('../lib/sound')
+const {dispatch} = require('../lib/dispatcher')
+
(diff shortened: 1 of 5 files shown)


dcposch:
@feross @mathiasvr

renderer/controllers/torrent-controller.js
```diff
@@ -0,0 +1,192 @@
...
+
+  // Gets a torrent summary {name, infoHash, status} from state.saved.torrents
+  // Returns undefined if we don't know that infoHash
+  getTorrentSummary (torrentKey) {
+    return TorrentSummary.getByKey(this.state, torrentKey)
+  }
+}
+
+function getTorrentPath (torrentSummary) {
+  var itemPath = path.join(torrentSummary.path, torrentSummary.files[0].path)
```
mathiasvr: We probably want to use TorrentSummary.getFileOrFolder here.

mathiasvr:
LGTM 👍

renderer/controllers/torrent-controller.js
dcposch: good call, done

PR #722, opened 2016-07-17
Automatic download subtitles

alamcordeiro:
Automatic download subtitles from opensubtitles. 
Based on checksum from movie files

captura de tela 2016-07-16 as 8 40 04 pm


demoneaux:
This is similar to #677. For now, I don't think WebTorrent Desktop will include this feature in its core. Perhaps after the 1.0 release, we'll look into adding extensions such as these.

xnakos:
@alamcordeiro Just a snack for thought; while the file is being downloaded, the checksum is bogus. So, if someone wanted to download subtitles for a video file that is still being downloaded, using the checksum is not an option. If I remember correctly, using SMPlayer to download subtitles while a file was being downloaded provided no results, so it made me think that this method was used. On the other hand, VLC's extension for downloading subtitles provides two methods for searching. One by checksum and one by filename. I think that the latter plays well too.

grunjol:
The opensubtitle hash is based on size + 64bit cheksum of the first and last 64k of the file.
You can check piece availability to ensure hash is not bogus (or maybe force first and last blocks).

xnakos:
@grunjol Thanks for the correction! Here is the source too: https://trac.opensubtitles.org/projects/opensubtitles/wiki/HashSourceCodes (a JavaScript implementation is also contained in the page).

feross:
Thanks for the PR, but we're not going to add this feature to WebTorrent Desktop, because the legality of OpenSubtitles is not clear to us.

alamcordeiro:
Okay, thanks

septemberboy7:
@alamcordeiro Please develop a plugin.

codealchemist:
@septemberboy7, I contributed with a plugin system in this PR #1125 .
Once we can move forward with it I'd like to create a plugin with what I did on #677 .
I also think that auto downloading subs in WebTorrent Desktop will be awesome!
We need @feross and @dcposch to review the PR.
I also hope to get some feedback on that too :)

@alamcordeiro, if #1125 gets approved maybe we can work together on the subtitles plugin.
Cheers!

alamcordeiro:
Excelent. I at disposal

PR #723, opened 2016-07-17
fix showItemInFolder

mathiasvr:
Fix strange error where clicking 'Show In Finder' would also show the 'confirm delete torrent' modal. Also make it show the torrent content and not the cached torrent file.


dcposch:
LGTM ugh i'm a tool

PR #727, opened 2016-07-19
play first video file

anonymlol:
The files inside a torrent are usually sorted, so I believe playing the first file as default makes more sense than playing the biggest.

### DIFF
--- a/renderer/lib/torrent-player.js
+++ b/renderer/lib/torrent-player.js
@@ -47,14 +47,11 @@ function isPlayableTorrent (torrentSummary) {
 
 // Picks the default file to play from a list of torrent or torrentSummary files
 // Returns an index or undefined, if no files are playable
+// Play the first video file
 function pickFileToPlay (files) {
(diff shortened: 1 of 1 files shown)


mathiasvr:
Yes when there are multiple video files this is often desirable, but maybe we should select biggest file if one is considerably larger than the others, or just if there are only 1 or 2 playable files?

anonymlol:
I think we should stick to the file order inside the torrent by default. Playing something other than the first file should be set by a specific parameter. There was a discussion about this [here](https://github.com/feross/webtorrent-desktop/issues/83#issuecomment-228266609).

mathiasvr:
Selecting an index is still only a proposal for the proposed 'stream+magnet' protocol and will not work with normal magnet or torrent links. I was thinking that selecting the first file would be a problem for torrents that contains only the main video and a sample, but it seems that the main content is usually ordered first, so this PR is probably fine 👍

Issue #725, opened 2016-07-18
Extremely sluggish scrolling


dcposch:
**What version of WebTorrent Desktop?** d0733d33705c4cbf5dee26c43cda2e8febc80419 (after 0.8.1)

**What operating system and version?** Mac OS El Capitan

**What did you do?** Opened the detail view on this torrent ([Huckleberry Finn](https://archive.org/details/huckleberry_mfs_librivox)), which contains 701 files.

**What did you expect to happen?** A long list of files with smooth scrolling.

**What actually happened?** Very choppy, nearly unusable scrolling and increased CPU load.

The app just froze for ~1 second at a time and became mostly unresponsive to inputs.

Collapsing the detail view made the app fast again.

The torrent wasn't finished downloading, so torrent progress events were causing the whole file list to redraw with new % complete for each file. Updating 700 `<divs>` once per second shouldn't be a problem, though.

**It looks like HyperX has poor performance.**

Here's the result of profiling during the situation described above:

<img width="1074" alt="screen shot 2016-07-17 at 5 53 22 pm" src="https://cloud.githubusercontent.com/assets/169280/16904357/2ab58dd8-4c49-11e6-9ae7-bb2b1329f344.png">

The biggest time sinks are all inside hyperx. It's parsing our HTML templates not only at runtime, but every single time they're called --- in this case, that means it's parsing the hyperx template in `renderFileRow()` 700+ times per call to update(). The parser is written in Javascript and is not particularly efficient.

The original and most popular virtual DOM library, React, parses its templates at compile time (JSX). That seems more reasonable.

(React also has other advantages. For example, our virtualdom + hyperx setup replaces every event listener in the app on every call to update() by default. We've partially worked around that by memoizing the event handlers using `dispatcher(...)`. In general React is a lot more widely used and more actively developed.)

--

WebTorrent Desktop needs to feel snappy and reliable. I suggest we do one of two things:
- Remove hyperx and go back to using virtualdom directly. (See https://github.com/feross/webtorrent-desktop/commit/b5561f13b4181e8ad9144fd882af0c7b5df5aac9)
- Switch to React

Thoughts?

feross:
I've never used React on a project before, so I have no idea if I'll like it. At a high level, I agree with your reasoning. I'm happy to switch to it if it solves the perf issues.

dcposch:
fixed!

[...]

PR #729, opened 2016-07-20
Switch to React

dcposch:
The goal is to make the app faster and smoother. As a bonus, React is more standard and more actively maintained than hyperx/virtual-dom/main-loop.

See #725 for motivation

@feross


dcposch:
## Work left to do
- [ ] Fix all React warnings
- [ ] Test speed improvement
- [ ] Fix the packaging scripts

dcposch:
We should also use React in a more idiomatic way. I ported it with minimum changes, but later on we should convert function calls like `{Header(..)}` into their own React components like `<Header ... />`

dcposch:
## Speed test
* Load [this torrent](https://archive.org/details/huckleberry_mfs_librivox) and expand it
* Open the dev tools, go to Timeline view, start recording
* Scroll for a few seconds, then stop recording

The timeline view shows FPS and breaks down where time is being spent.

### hyperx
Can't really scroll. See #725 
**Computing the vdom:** ~500ms
**Applying the vdom to the actual dom:** didn't measure


### React
With React in production mode (extra checks and warnings disabled with NODE_ENV=production, which is what our users see):
**Computing the vdom:** ~10ms
**Applying the vdom to the actual dom:** ~60ms

This is better, but unfortunately the scrolling is still choppy. First, 70ms is a lot longer than 16ms, so you are freezing for a few frames (compared to 60fps) every time you call update(). Second, even when the DOM is not updating at all, Chrome can't scroll it smoothly.

Below, the orange parts correspond to time spent in JS (our code and React). The purple parts are time spent in Chrome rendering frames of scrolling animation while the DOM isn't changing.

screen shot 2016-07-20 at 10 59 35 pm

In short it's better now, but still not smooth. IMO this PR is a good first step. Next we'll have to make the DOM more lightweight (eg by capping the number of file rows we show).

dcposch:
The dom is just SLOW

I saved the torrent list view as static HTML:

screen shot 2016-07-22 at 5 21 44 pm

A few stats:
- `torrent-details.html` is only 97KB
- There are only 351 rows in the table
- The table layout is really easy to compute
  - `table-layout: fixed` in the CSS, so no autosizing columns
  - The rows have a fixed pixel height
  - The side columns have fixed pixel width. The middle column takes the remaining width

And this is plain HTML + CSS, no Javascript of any kind.

Anyway despite that, when mousing over the list, the `:hover` state is very choppy.

As you can see below it runs at about 10FPS.

screen shot 2016-07-22 at 5 34 07 pm

feross:
@dcposch Here's an idea: remove the background image. It's hugely stretched. I bet that's causing this.

src/renderer/controllers/torrent-controller.js
```diff
@@ -129,8 +129,6 @@ module.exports = class TorrentController {
     //       files which are completed after a video starts to play to be added
     //       dynamically to the list of subtitles.
     // checkForSubtitles()
-
-    dispatch('update')
```
feross: Why was this removed?

src/renderer/main.js
```diff
@@ -52,7 +52,7 @@ State.load(onState)
 // the dock icon and drag+drop.
 function onState (err, _state) {
   if (err) return onError(err)
-  state = _state
+  state = window.state = _state // Make available for easier debugging
```
feross: Interesting. The `window.state = _state` line wasn't required before.

Is this file no longer executing in the top-level scope because of the react build step?

src/renderer/main.js
```diff
@@ -74,24 +74,15 @@ function onState (err, _state) {
...
-
   // Listen for messages from the main process
   setupIpc()
 
   // Calling update() updates the UI given the current state
   // Do this at least once a second to give every file in every torrentSummary
   // a progress bar and to keep the cursor in sync when playing a video
   setInterval(update, 1000)
+  window.requestAnimationFrame(renderIfNecessary)
+  app = ReactDOM.render(<App state={state} />, document.querySelector('#body'))
```
feross: `body` shouldn't need an id. There's only ever one body.

src/renderer/main.js
```diff
@@ -132,21 +123,23 @@ function lazyLoadCast () {
...
-  try {
-    return App(state)
-  } catch (e) {
-    console.log('rendering error: %s\n\t%s', e.message, e.stack)
+// Calls render() to go from state -> UI, then applies to vdom to the real DOM.
+// Runs at 60fps, but only executes when necessary
+var needsRender = 0
+
+function renderIfNecessary () {
+  if (needsRender > 1) console.log('combining %d update() calls into one update', needsRender)
```
feross: If this log gets printed out often, it will be a perf hit and spam the console. We might want to axe it once we're sure this code is working correctly.

src/renderer/main.js
```diff
@@ -302,6 +295,7 @@ function backToList () {
     var contentTag = document.querySelector('.content')
     if (contentTag) contentTag.scrollTop = 0
 
+    // TODO dcposch: is this still required with React?
```
feross: It's probably still required because we're still using a `requestAnimationFrame` loop. But maybe there's a cleaner way to tell react to redraw with `app.setState(state)`.

src/renderer/views/app.js
```diff
@@ -0,0 +1,95 @@
...
+        <div key='content' className='content'>{getView(state)}</div>
+        {getModal(state)}
+      </div>
+    )
+
+    return vdom
+  }
+}
+
+function getErrorPopover (state) {
```
feross: These helper functions should probably be private methods on the `App` class so that `state` doesn't need to be passed in each time.

src/renderer/views/create-torrent-error-page.js
```diff
@@ -0,0 +1,26 @@
+const React = require('react')
+
+const {dispatcher} = require('../lib/dispatcher')
```
feross: Is this dispatcher library still the best way to accomplish this? Doesn't react offer something cleaner?

src/renderer/views/unsupported-media-modal.js
```diff
@@ -0,0 +1,39 @@
...
+      : err
+    var actionButton = state.modal.vlcInstalled
+      ? (<button className='button-raised' onClick={dispatcher('vlcPlay')}>Play in VLC</button>)
+      : (<button className='button-raised' onClick={() => this.onInstall}>Install VLC</button>)
+    var vlcMessage = state.modal.vlcNotFound
+      ? 'Couldn\'t run VLC. Please make sure it\'s installed.'
+      : ''
+    return (
+      <div>
+        <p><strong>Sorry, we can't play that file.</strong></p>
```
feross: Look at this GitHub rendering fail.

Let's name all the files with JSX in them with a `.jsx` extension so this doesn't happen in the future?

static/main.html
```diff
@@ -3,9 +3,13 @@
   <head>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1">
-    <link rel="stylesheet" href="main.css" charset="utf-8">
+    <title>WebTorrent Desktop</title>
+    <link rel="stylesheet" href="main.css">
+    <script>
+	require('../build/renderer/main.js')
```
feross: Why is this no longer a script tag with an `async` property? I did LOTS of perf testing and found that that was fastest. Is this change required to get React to work?

static/main.html
```diff
@@ -3,9 +3,13 @@
...
+    </script>
   </head>
   <body>
-    <script async src="main.js"></script>
+    <div id='body'></div>
```
feross: Why not just tell react to use the `<body>` directly?

static/main.html
```diff
@@ -3,9 +3,13 @@
   <head>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1">
-    <link rel="stylesheet" href="main.css" charset="utf-8">
+    <title>WebTorrent Desktop</title>
```
feross: The actual browser window gets set with Electron APIs, but I suppose this is fine

static/main.html
```diff
@@ -3,9 +3,13 @@
...
+    <link rel="stylesheet" href="main.css">
```
feross: Why is charset not specified anymore? Was this an intentional change?

static/webtorrent.html
```diff
@@ -14,9 +15,11 @@
         height: 140px;
       }
     </style>
+    <script>
+      require('../build/renderer/webtorrent.js')
```
feross: Same question: why not `<script async ...>`

feross:
Hey @dcposch, this is nice work! 👍 👍 👍 

Never used React before, but curious to see how this goes. It looks pretty nice so far. I left lots of comments and questions inline.

Please address them, then we can merge this. I think we shouldn't release for a few days after this is merged since I wasn't able to compare line-by-line for some of the files. Some of the file diffs just show every line removed, then every line added. :( So we should wait a few days for any bugs to get shaken out.

static/webtorrent.html
dcposch: Doing that breaks `require()`. If we want `<script ...>` to work, we either have to:
* Write ugly things like `require('../build/renderer/<etc>')` in the actual source files
* Move `main.html` to be next to `main.js`

Doing it this way lets us separate `src/` for JS sources only and `static/` for HTML, CSS, etc

static/main.html
dcposch: I modified `main.html` to pass validation: https://html5.validator.nu/

static/main.html
dcposch: Added b/c HTML5 validator requires it: https://html5.validator.nu/

static/main.html
dcposch: That's what I did before, but React prints a warning when you do that. (For silly reasons that don't apply to us--apparently some ppl have browser extensions that break React when used on `<body>`--but I also didn't see a better way to turn off the warning.)

src/renderer/main.js
dcposch: This is for the wrapper `<div>`, needed to avoid React's warning about using `<body>` directly

src/renderer/main.js
dcposch: It's no longer in top-level scope because of the `require()` path change, because I moved `main.html` over to `static/`

dcposch:
@feross I tried removing the background image, but that wasn't the cause of the slowness.

src/renderer/main.js
dcposch: Sure, removed.

src/renderer/main.js
dcposch: I'll try it and see if it's slower. Reason I haven't done that before to is to allow collapsing multiple `update()` calls into one.

feross:
👍

src/renderer/controllers/torrent-controller.js
dcposch: It was causing double update()s.

src/renderer/main.js
dcposch: Added a comment in `main.html` to explain

static/main.html
dcposch: I'll test and post the results below

dcposch:
## Startup time
Checking the logs for the `init` timer.

Measured on a 2016 12" Macbook running Mac OS El Capitan.

### Before (18aadf9d236ab19309cbaead44559d5fdae27b01)
main process: 850ms
renderer: 25oms

### After (b51dada8e1c3db8136ca67909d94943e72645cbf)
main process: 900ms
renderer: 350ms

dcposch:
And scrolling thru the 350 files in Adventures of Huckleberry Finn is fast now!

feross:
👍 👍 👍

PR #730, opened 2016-07-21
v0.9.0

dcposch:
@feross


mathiasvr:
Out of curiosity, why is this released before it's merged to master? This means that any changes would have to be done for the next release right?

dcposch:
@mathiasvr my B. Thought I had already merged it. There were two more bugfixes @feross & I found last night while testing just before the release

mathiasvr:
@dcposch Thanks, no problem! This just confused me about the relation between branches and releases, makes perfect sense now.

PR #733, opened 2016-07-23
application-config@1

feross:
This protects against corrupting the configuration file if the
application crashes before the stream finishes writing to the file

Especially important for large configuration files

See https://github.com/LinusU/node-application-config/pull/3


demoneaux:
LGTM

feross:
@demoneaux thanks for the review!

PR #738, opened 2016-07-25
bump versions of electron and fs-extra

mathiasvr:


feross:
`fs-extra` changelog includes breaking changes, but I don't think they affect us: https://github.com/jprichardson/node-fs-extra/blob/master/CHANGELOG.md

Whenever I've bumped the Electron version in the past, I've looked through the release logs for new features and changes that we should make in WebTorrent Desktop. Lots of APIs get changed/added in each release. For this bump, it looks like these are relevant:

### General

Update to Chrome 51.0.2704.106
Improve performance for looking up paths in `require`
`accelerator` and `label` are now optional when `role` is specified for `MenuItem`

### macOS

Fix the fullscreen button showing in frameless window on OS X 10.9.
Fix transparent window having border on OS X 10.9.
Fix crash when using frameless window on OS X 10.9.
Update Squirrel.Mac to 0.2.1 (fixes situations in which updates would not get applied)
Add 'quit' role to MenuItem (https://github.com/electron/electron/pull/6154)
Add 'togglefullscreen' role to MenuItem (https://github.com/electron/electron/pull/6166/files)
Add 'delete' role to MenuItem (https://github.com/electron/electron/pull/5870)
Fix frameless window not showing in Window menu.
Fix context menu always choosing first item by default.
Fix title of frameless window showing as empty under fullscreen

### Linux

Fix menubar not hiding after entering fullscreen. (Some Linux distros)
Show badge counts on Linux (Use `app.setBadgeCount()` API. On macOS, `app.dock.setBadge()` which is macOS-specific can be changed to `app.setBadgeCount()` now)
Fix startup crashes (Some Linux distros)
Fix message boxes and file dialogs not being modal.
Fix `BrowserWindow.focus()` not moving focus to the window

mathiasvr:
@feross Thanks for going through this! I regret not providing this information with the PR.

feross:
It's all good. 👍  I'm about to send a PR that makes use of the relevant new APIs.

feross:
@mathiasvr See https://github.com/feross/webtorrent-desktop/pull/739

PR #739, opened 2016-07-25
Electron: Updates for Electron 1.2.8

feross:
Followup PR to #738.

- Electron: Use default labels and accelerators
- macOS: add missing Edit menu roles
- Use 'quit' role
- Linux: Support showing badge count

Please review: @mathiasvr

### DIFF
--- a/src/main/dock.js
+++ b/src/main/dock.js
@@ -30,12 +30,11 @@ function downloadFinished (path) {
 }
 
 /**
- * Display string in dock badging area. (OS X)
+ * Display a counter badge for the app. (OS X, Linux)
  */
-function setBadge (text) {
-  if (!app.dock) return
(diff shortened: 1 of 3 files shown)


mathiasvr:
LGTM 👍
But I think the undo/redo menu items are a little confusing. They are always active even though there is nothing to undo/redo, and they cannot be used to undo something like removing or adding a torrent. Can they actually be used for anything else than undoing/redoing text in the `Open Torrent Address` modal? If not, I would remove them.

feross:
We also have Cut/Copy/Select All which have the same problem.

It seems like most macOS apps tend to include these items even when they do nothing most of the time, so they can be used in text fields. I agree with you though. We should look at how we can disable these items when the user's not focused in a text field.

PR #740, opened 2016-07-25
electron-prebuilt@1.3.0

feross:
Another Electron was just released. Let's bump from 1.2.8 to 1.3.0.

Changelog:

- Upgrade to Chrome 52
- Update to Node.js 6.3.0


mathiasvr:
Nice, LGTM 👍

PR #742, opened 2016-07-27
Set peer ID to start with "-WD-"

feross:
To distinguish WebTorrent Desktop (WD) from WebTorrent in the browser
(WW).

See the spec:

http://www.bittorrent.org/beps/bep_0020.html
https://wiki.theory.org/BitTorrentSpecification

### DIFF
--- a/src/renderer/webtorrent.js
+++ b/src/renderer/webtorrent.js
@@ -2,14 +2,16 @@
 // process from the main window.
 console.time('init')
 
-var WebTorrent = require('webtorrent')
-var defaultAnnounceList = require('create-torrent').announceList
 var deepEqual = require('deep-equal')
+var defaultAnnounceList = require('create-torrent').announceList
 var electron = require('electron')
 var fs = require('fs-extra')
+var hat = require('hat')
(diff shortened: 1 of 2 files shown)


src/renderer/webtorrent.js
```diff
@@ -26,9 +28,36 @@ global.WEBTORRENT_ANNOUNCE = defaultAnnounceList
...
+
+/**
+ * Version number in Azureus-style. Generated from major and minor semver version.
+ * For example:
+ *   '0.16.1' -> '0016'
+ *   '1.2.5' -> '0102'
+ */
+var VERSION_STR = VERSION.match(/([0-9]+)/g)
+  .slice(0, 2)
+  .map(function (v) { return zeroFill(2, v) })
```
mathiasvr: Nitpick: Use arrow function: `.map((v) => zeroFill(2, v))`.

mathiasvr:
LGTM

src/renderer/webtorrent.js
feross: Will do!

yciabaud:
Nice!

PR #743, opened 2016-07-27
OS X -> Mac

feross:
OS X is no longer the marketing name. It's technically macOS, but I'm using Mac in the website marketing since that's what I think users will expect.


mathiasvr:
You should update the GitHub repository description as well. :v:

feross:
Done.

PR #746, opened 2016-07-27
Fix 700

### LINKED ISSUE #700: VLC won't minimize and have title
Version 0.8.1 (0.95.2)
Ubuntu 16.04
**What did you do?**
Open unsupported format in vlc
vlc to have title displaying played file in top bar rather than localhost and minimize player
**What actually happened?**
played file isn't shown, and it won't minimize (vlc).

feross:
Fix #700

The title is set with the `--meta-title` flag to VLC.

### DIFF
--- a/src/main/dialog.js
+++ b/src/main/dialog.js
@@ -117,5 +117,5 @@ function setTitle (title) {
 }
 
 function resetTitle () {
-  window.main.dispatch('resetTitle')
+  windows.main.dispatch('resetTitle')
 }
--- a/src/main/dialog.js
+++ b/src/main/dialog.js
@@ -118,5 +118,5 @@ function setTitle (title) {
 }
 
 function resetTitle () {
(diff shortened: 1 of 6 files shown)


src/main/dialog.js
```diff
@@ -118,5 +117,5 @@ function setTitle (title) {
 }
 
 function resetTitle () {
-  setTitle(config.APP_WINDOW_TITLE)
+  window.main.dispatch('resetTitle')
```
mathiasvr: Correction: `window` should be `windows`

mathiasvr:
LGTM aside from comment, nice use of `dispatch` 👍

src/main/dialog.js
feross: Good catch!

feross:
@mathiasvr thanks for the review!

PR #747, opened 2016-07-27
electron-prebuilt@1.3.1

feross:
Electron changelog: https://github.com/electron/electron/releases/tag/v1.3.1

The only bugfix that potentially affects us is:

- Window position doesn't restore correctly

None of the new features is relevant to us.


feross:
Thanks for the 👍 @mathiasvr

PR #758, opened 2016-08-04
Allow dragging magnet links (Fix #284)

### LINKED ISSUE #284: Allow dragging magnet links
It would be handy to have the app accept dragging of torrent links instead of doing a manual copy-paste.

feross:


mathiasvr:
Works well for me, but we might want to handle incompatible text at some point.

codealchemist:
You were ahead of me by 10hs on this one @feross xD
Works ok, thanks!

feross:
@mathiasvr handle incompatible text? Do you mean show a friendlier message instead of `ENOENT: no such file or directory`? Open a new issue?

mathiasvr:
@feross yes exactly, I think errors displayed in the UI should be short and descriptive, but I don't think it's an important issue right now.

PR #760, opened 2016-08-05
Replace deprecated `react-tools` with `babel`

feross:
- Switch to babel, since react-tools has been deprecated since June 12, 2015. See https://facebook.github.io/react/blog/2015/06/12/deprecating-jstransform-and-react-tools.html

- Move babel command to "npm run build"

- Move commands for package into "bin/package.js"

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -23,7 +23,11 @@ var windows = require('./windows')
 var shouldQuit = false
 var argv = sliceArgv(process.argv)
 
-if (!argv.includes('--dev')) process.env.NODE_ENV = 'production'
+if (config.IS_PRODUCTION) {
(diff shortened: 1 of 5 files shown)


dcposch:
LGTM

src/main/index.js
```diff
@@ -23,7 +23,11 @@ var windows = require('./windows')
 var shouldQuit = false
 var argv = sliceArgv(process.argv)
 
-if (!argv.includes('--dev')) process.env.NODE_ENV = 'production'
+if (config.IS_PRODUCTION) {
+  // When Electron is running in produdtion mode (packaged app), then run React
```
dcposch: production

feross: 👍

PR #770, opened 2016-08-10
Fix screen stacking bug

### LINKED ISSUE #665: Don't allow Create Torrent window to open from Prefs window
Currently, you can alternate clicking Create New Torrent and Preferences to create a stack of prefs and new torrent windows as deep as you want, and move between them using the back and forward buttons.

dcposch:
You can no longer open a whole stack of Prefs windows, or Create Torrent windows

Simplifies and fixes behavior when dropping files onto the app or the dock icon. Before, you could use drag-drop to create stacks of Create Torrent windows. Now, you can only create torrents from the home screen.

Fixes #665

### DIFF
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -1,11 +1,10 @@
 module.exports = {
   init,
-  onPlayerClose,
-  onPlayerOpen,
+  setPlayerOpen,
+  setWindowFocus,
+  setAllowNav,
   onToggleAlwaysOnTop,
-  onToggleFullScreen,
-  onWindowBlur,
-  onWindowFocus
+  onToggleFullScreen
 }
 
 var electron = require('electron')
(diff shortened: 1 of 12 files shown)


dcposch:
@feross

src/main/menu.js
```diff
@@ -350,6 +333,17 @@ function getMenuTemplate () {
       click: () => dialog.openSeedFile()
     })
 
+    // Edit -> Preferencs (Windows, Linux)
```
feross: Preferences

src/main/menu.js
```diff
@@ -24,26 +23,28 @@ function init () {
...
+function setWindowFocus (flag) {
+  getMenuItem('Full Screen').enabled = flag
+  getMenuItem('Float on Top').enabled = flag
+}
+
+// Disallow opening more screens on top of the current one.
+function setAllowNav (flag) {
+  getMenuItem('Preferences').enabled = flag
+  getMenuItem('Create New Torrent...').enabled = flag
+  getMenuItem('Open Torrent File...').enabled = flag
```
feross: You're missing the "Create New Torrent from File..." option. This one only appears on Windows/Linux, so you'll need an if-statement to check that it exists before setting the `enabled` property.

src/main/menu.js
feross: This comment should be "Edit menu (Windows, Linux)" to match the style of the others.

src/renderer/controllers/torrent-list-controller.js
```diff
@@ -40,6 +40,11 @@ module.exports = class TorrentListController {
 
   // Shows the Create Torrent page with options to seed a given file or folder
   showCreateTorrent (files) {
+    // You can only create torrents from the home screen.
+    if (this.state.location.url() !== 'home') {
+      return dispatch('error', 'Please go back to the torrent list before creating a new torrent.')
```
feross: It's not ideal that this error appears in two places. Can't it be removed from here? With the changes in this PR, it's not possible for the drag-and-drop handler to call `showCreateTorrent` except from the home screen anyway.

feross: The one other place I think it can be called is via the menu item "Create New Torrent..." when the player is open. Shouldn't the player also do `ipcRenderer.send('setAllowNav', false)`? This seems like the right place to handle it. Then this line can be removed here.

feross:
LGTM -- but see inline comments.

src/main/menu.js
```diff
@@ -24,26 +23,28 @@ function init () {
...
-  getMenuItem('Add Subtitles File...').enabled = true
+function setWindowFocus (flag) {
+  getMenuItem('Full Screen').enabled = flag
+  getMenuItem('Float on Top').enabled = flag
+}
+
+// Disallow opening more screens on top of the current one.
+function setAllowNav (flag) {
+  getMenuItem('Preferences').enabled = flag
+  getMenuItem('Create New Torrent...').enabled = flag
```
feross: You're missing the "Create New Torrent from File..." option. This one only appears on Windows/Linux, so you'll need an if-statement to check that it exists before setting the `enabled` property.

src/main/menu.js
```diff
@@ -24,26 +23,28 @@ function init () {
...
+  getMenuItem('Open Torrent File...').enabled = flag
+  getMenuItem('Open Torrent Address...').enabled = flag
```
feross: These last two "Open..." options don't actually navigate to new pages. They're equivalent to drag-and-dropping a new .torrent file onto the app (which is supported from all pages). For that reason, shouldn't these options remain enabled?

feross:
Nice work! This is a lot more consistent. I like it.

LGTM -- but see inline comments.

src/renderer/controllers/torrent-list-controller.js
dcposch: > Shouldn't the player also do ipcRenderer.send('setAllowNav', false)? 

Then you can't open Prefs from within the player

feross: Oh, that's bad. Okay, this is fine then.

Btw right now, you also can't open Prefs from within the Create Torrent page.

src/main/menu.js
dcposch: Good call, added

src/main/menu.js
dcposch: Sure, I'll put them back

src/renderer/controllers/torrent-list-controller.js
dcposch: Just tried it --- I *can* open Prefs from within Create Torrent. (I just can't open another Prefs from within Prefs, or another Create Torrent from within either of those.)

src/main/menu.js
dcposch: fixed

feross:
👍

feross:

PR #771, opened 2016-08-10
Pref: default torrent file handler

### LINKED ISSUE #367: WebTorrent silently set itself as default magnet/torrent handler
**What version of WebTorrent Desktop?**
0.3.3

OSX 10.11

**What did you do?**
Open app

App should not silently modify user settings

**What actually happened?**
WebTorrent silently set itself as default magnet handler

dcposch:
Before, the app made itself the default torrent file handler automatically, pissing off some of our users. Now, it's not by default, and you can change it in the prefs.

Fixes #367


dcposch:
@feross @diimdeep @mathiasvr

dcposch:
www gifcreator me_hyspcw

src/renderer/views/preferences.js
```diff
@@ -43,6 +44,29 @@ function renderDownloadDirSelector (state) {
...
+      className='btn'
+      onClick={toggleFileHandlers}>
+      {buttonText}
+    </button>
+  )]
+  return renderControlGroup(definition, controls)
+
+  function toggleFileHandlers () {
+    var isFileHandler = state.unsaved.prefs.isFileHandler
+    dispatch('updatePreferences', 'isFileHandler', !isFileHandler)
```
feross: Should use `setStateValue` to match the other place it's used. Or, better just get rid of that function and call `dispatch('updatePreferences', ...)` directly everywhere.

src/renderer/controllers/prefs-controller.js
```diff
@@ -41,6 +42,9 @@ module.exports = class PrefsController {
   // All unsaved prefs take effect atomically, and are saved to config.json
   save () {
     var state = this.state
+    if (state.unsaved.prefs.isFileHandler !== state.saved.prefs.isFileHandler) {
+      ipcRenderer.send('setDefaultFileHandler', state.unsaved.prefs.isFileHandler)
+    }
```
feross: This is fine, but it would be better if all preferences took effect immediately instead of that the time the preferences page is unloaded/destroyed. That's how it works in all macOS apps -- and also the reason why preferences pages on macOS never have a Save/Cancel button.

feross:
LGTM 👍

mathiasvr:
I think the program should initially associate with the `.torrent` extension but not make itself the default handler, that should be up to the user through the OS. For `.magnet` links I would add a button that makes WebTorrent the default handler when clicked and not every time it is started.

feross:
@mathiasvr The menu label is misleading. This preference affects both .torrent and magnet handling.

On macOS, the .torrent association is automatic (defined in Info.plist). So if there's no other handler, WebTorrent will be default. If there's another default, then the user can still use "Open With..." to select WebTorrent.

On Windows / Linux, the current situation (after this PR) is that WebTorrent does nothing until the preference is toggled. This is not ideal -- thanks for pointing this out.

With this change, now the `.desktop` file will not get installed on Linux systems until the user toggles the preference option. But that file is needed for the app icon to show up correctly in the Unity launcher, and to support pinning etc.

I think we need to change `handlers.js` to distinguish between installing handlers and making WebTorrent the default.

- install: associate with .torrent (but not as the default program), i.e. install .desktop file (linux), add to registry (windows)
- makeDefault: become default .torrent handler, register magnet handler (there can only be one)
- removeDefault: give up .torrent handler, unregister magnet handler
- uninstall: only used by the Windows program uninstaller to completely clean up registry entries, never called on Mac or Linux

@dcposch Does that seem reasonable to you?

feross:
Let's move discussion to this new issue so we don't lose track of it: https://github.com/feross/webtorrent-desktop/issues/773

feross:

PR #776, opened 2016-08-13
Check for missing download path

### LINKED ISSUE #646: File Path Warnings Needed
**What version of WebTorrent Desktop?** Version 0.7.2 (0.94.3)

**What operating system and version?** elementary OS 0.3.2 Freya (64-bit) Built on Ubuntu 14.04

**What did you do?** Start WebTorrent without mounted drive, where Download Path resides

**What did you expect to happen?** Error, Download Path does not exist, confirmation to create path or choose alternative.

**What actually happened?** Folder created in /media/user/, where files were downloaded off away from desired location

**Details** 
Obviously I am saving all my torrents to a separate drive on my PC, where the drive has to be mounted before using. Silly me, I start WebTorrent before this happens. WebTorrent happily chugs along, downloads the files, returns no warnings. I realize after the download that my drive is not mounted, and the downloaded files are in /media/mark/External_Drive1/Torrent_Downloads/ when External_Drive is not mounted. _(Please note that the incremented number may be from mounting the drive after Ubuntu sees the original path as not empty.)_

I safely moved the files to the drive, however WebTorrent still thinks the torrents reside in _Drive1. Closing the application, modifying the .config file, and reopening did not resolve the issue. I had to edit the config file, kill WebTorrent process, and then reopen for changes to make effect. 

**Possible Solutions** 
Upon opening the application, the download path should be tested to see if it exists. If it does not exist, a warning should appear to remap, or create the directory. Verbosely, the warning should also detect if the /media path is being used, to suggest remounting & retrying.

Additionally, right clicking each torrent in the list should provide options to change the Download Path of each torrent, moving the files. This could save someone from having files dumped where they don't want, and modifying the .config as I did.

dcposch:
Also handle the case where the user cancels out of Save Torrent File As

Fixes #646.

### DIFF
--- a/src/renderer/controllers/torrent-list-controller.js
+++ b/src/renderer/controllers/torrent-list-controller.js
@@ -76,21 +76,25 @@ module.exports = class TorrentListController {
   }
 
   // Starts downloading and/or seeding a given torrentSummary.
-  startTorrentingSummary (torrentSummary) {
-    var s = torrentSummary
-
-    // Backward compatibility for config files save before we had torrentKey
-    if (!s.torrentKey) s.torrentKey = this.state.nextTorrentKey++
+  startTorrentingSummary (torrentKey) {
+    var s = TorrentSummary.getByKey(this.state, torrentKey)
+    if (!s) throw new Error('Missing key: ' + torrentKey)
 
     // Use Downloads folder by default
     if (!s.path) s.path = this.state.saved.prefs.downloadPath
 
-    ipcRenderer.send('wt-start-torrenting',
-                      s.torrentKey,
-                      TorrentSummary.getTorrentID(s),
-                      s.path,
-                      s.fileModtimes,
-                      s.selections)
+    fs.stat(TorrentSummary.getFileOrFolder(s), function (err) {
+      if (err) {
+        s.error = 'path-missing'
+        return
+      }
+      ipcRenderer.send('wt-start-torrenting',
+        s.torrentKey,
+        TorrentSummary.getTorrentID(s),
+        s.path,
+        s.fileModtimes,
+        s.selections)
+    })
   }
 
   // TODO: use torrentKey, not infoHash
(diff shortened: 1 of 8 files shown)


src/renderer/controllers/torrent-list-controller.js
```diff
@@ -274,6 +274,7 @@ function saveTorrentFileAs (torrentSummary) {
     ]
   }
   electron.remote.dialog.showSaveDialog(electron.remote.getCurrentWindow(), opts, function (savePath) {
+    if (!savePath) return // They clicked Cancel
```
feross: good catch!

feross:
This is good. One concern though: this only shows the user a message when the download folder is gone, but it doesn't prevent the active torrents from being added to the torrent engine, right? That means that `webtorrent` will make new directories, right? That's not what we want.

We should probably wait for the check to confirm the directory exists before adding to the torrent engine.

But, come to think of it, there's a bigger issue.

The default download path in Preferences is not necessarily the same as the torrents. The preference could have been changed to an external drive after several torrents were already started in ~/Downloads. In that case, the torrents that live in ~/Downloads should show up in the list and be usable by the user.

What do you think about changing this PR `fs.stat` the `torrentSummary.path` before calling `webtorrent.add()`? Torrents whose locations do not exist anymore could still show up in the list, but with warning text in their status line (below the name). This is what Transmission does.

As for the warning about the default download path being missing, I think that can stay. But I don't think it shouldn't prevent rendering of the entire torrent list when it's missing. What about just showing the warning at the top of the list?

dcposch:
Ya, I just wanted to do the easy part first.

I can a per-torrent check similar to the one Transmission has.

dcposch:
gifff

dcposch:
@feross fixed. New error message

screen shot 2016-08-13 at 8 37 56 pm

feross:
LGTM -- we can work on improving the design of this and honing the copy in a future PR.

dcposch:
screen shot 2016-08-13 at 9 22 17 pm
screen shot 2016-08-13 at 8 55 43 pm

feross:

PR #787, opened 2016-08-20
If a power saver block already exists, do nothing

feross:
Before this change, when opening the player, both 'onPlayerOpen' and
'onPlayerPlay' would fire, which enabled, disabled, and re-enabled the
power save blocker in quick succession.

Instead, we just want to activate it once.


demoneaux:
LGTM.

PR #789, opened 2016-08-20
Ignore '.' argument which is annoying in development

feross:
Ignore '.' argument, which gets misinterpreted as a torrent id, when a development copy of WebTorrent is started while a production version is running.


demoneaux:
LGTM.

PR #792, opened 2016-08-21
Telemtry: post at least once a day

dcposch:
This ensures that people who keep the app running in the background
for days at a time are still counted as active users.


feross:
LGTM

Issue #293, opened 2016-04-05
Ubuntu: Add Unity shortcut on .deb install


feross:
> @dcposch: i think nearly all our Linux users will be on Ubuntu, so automatically adding to the launcher on unity is probably worth it

grunjol:
Ubuntu info

https://help.ubuntu.com/community/UnityLaunchersAndDesktopFiles

Basically it requires to execute 

`gsettings get com.canonical.Unity.Launcher favorites`

and it will give you a list of desktop files and unity command shortcuts like this:

`['application://google-chrome.desktop', 'application://skype.desktop', 'application://firefox.desktop', 'application://atom.desktop', 'unity://running-apps', 'unity://expo-icon', 'unity://devices']`

it can be easly parsed as a json array and add the webtorrent-desktop menu item and then execute

`gsettings set com.canonical.Unity.Launcher favorites "['application://google-chrome.desktop', 'application://skype.desktop', 'application://firefox.desktop', 'application://atom.desktop', 'application://webtorrent-desktop.desktop', 'unity://running-apps', 'unity://expo-icon', 'unity://devices'"`

remember to add the double quotes because is single parameter

feross:
We should really do this, because most people don't know that they need to run `webtorrent-desktop` from the terminal the first time in order add the shortcuts, etc.

Proof in this media coverage of the app: http://www.webupd8.org/2016/05/webtorrent-desktop-040-released-with.html

grunjol:
Before adding the 'shortcut' to unity launcher we need to create the the .desktop files so the system is aware of the app path/icons, resources, etc. 
`.desktop` file is dynamically created on startup as part of the initialization of the application, so you can not add (fix) to launcher before the .desktop file itself is there.

The dynamic launcher was created because originally the app only ran in standalone (single user) mode, but now you don't  know if you are running in the app from a 'packaged' version, or a standalone version.

The solution for packaged version can be add a hardcoded `.desktop` file and install the app system-wide or run the app using a parameter that creates the .desktop file dynamically as part of the package installation script. Either way, the app initialization should check if we already have the system-wide `.desktop` file to skip the _local_ standard initialization.

Let me think about this for one day (and get some feedback from other users) and willstart the implementation

mekarpeles:
Howdy @feross, we recently implemented this in @internetarchive's archivecd project. 

We placed our .desktop file in `/usr/share/applications/<app>.desktop`
[...]

PR #793, opened 2016-08-21
Add system-wide menu item for debian and derivates

grunjol:
This PR adds WD into the system, making the application discoverable and enable be launched from the UI and not from the console the first time. 
Fixes the issue mentioned http://www.webupd8.org/2016/05/webtorrent-desktop-040-released-with.html as part of #293


grunjol:
I added some icons to follow the integration guide in https://developer.gnome.org/integration-guide/stable/icons.html.en . They are already ImageOptim'd

static/linux/share/applications/webtorrent-desktop.desktop
```diff
@@ -0,0 +1,33 @@
+[Desktop Entry]
```
feross: Is there any way to unify this with `static/linux/webtorrent-desktop.desktop`? Looks mostly identical.

feross:
LGTM! Merged, will be released as part of WebTorrent Desktop v0.12.0.

static/linux/share/applications/webtorrent-desktop.desktop
grunjol: Yes, they are almost the same. My first attempt was create a command line handler like the squirrel-win32 one, which can invoke the (un)install handler, but i need to pass a parameter to indicate that is called from the system installation (or detect the user invoking was the superadmin) instead of the standard user startup registration. Then, based on that change, modifiy the template variables and after that, modify the entire handlers.js script to return a callback to shutdown the app. Basically, It requires some work and testing in all the supported platforms and had no time, so, it was simplier to create the final 'packaged' version of the resources.

feross: No worries!

PR #795, opened 2016-08-21
A bunch of small fixes

feross:
See individual commits

### DIFF
--- a/src/renderer/lib/migrations.js
+++ b/src/renderer/lib/migrations.js
@@ -38,8 +38,6 @@ function run (state) {
 }
 
 function migrate_0_7_0 (saved) {
-  console.log('migrate to 0.7.0')
-
   var fs = require('fs-extra')
   var path = require('path')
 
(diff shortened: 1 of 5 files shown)


feross:
Review: @dcposch

src/renderer/lib/migrations.js
```diff
@@ -95,15 +91,15 @@ function migrate_0_7_0 (saved) {
 }
 
 function migrate_0_7_2 (saved) {
-  if (!saved.prefs) {
```
dcposch: Was this a bug before? Seems like both ways should work.

Usually I wouldn't change migration steps that have already run

dcposch:
LGTM

src/renderer/lib/migrations.js
feross: Yeah, don't want to get in a habit of changing these after-the-fact, but it looked weird that every migration had a different way of checking for undefined (`!condition` vs. `=== undefined` vs. `== null`).

What we really want is `== null`, IMO.

Now that it's consistent, it's clear what to use in future migrations.

PR #797, opened 2016-08-22
Added channels support

codealchemist:
Channels provide easy access to categorized and curated content.
Channels are added by pasting their URLs in Preferences.
Each channel can be accessed in the channels section and provide its own collection of torrents.

@feross, @dcposch, is anyone exploring this concept already?
I was also thinking about adding a setting to WebTorrent to enable _broadcasting_, to share the existing torrent's list.


dcposch:
@codealchemist check out #856 

> I was also thinking about adding a setting to WebTorrent to enable broadcasting, to share the existing torrent's list.

How would that work? Who would you share it with?

codealchemist:
### Scenario 1:
Let's think about someone who produces video content and wants to share that with anyone.
That person can start a *torrent channel* using something like this:
https://github.com/codealchemist/torrent-channels
Or any other tool we can come up with.

Then they can share a URL to their channel on their website, over chat, etc.
Giving WebTorrent users, or users of any other apps which is compatible with a TBD *torrent channels spec*, the ability to view their channel and stream the torrents listed there.

### Scenario 2:
I watched an open sourced TV show which I really loved.
I want to share all the episodes in a convenient way with my friends.
I can create a *torrent channel* listing all of the episodes and share the link to it.
My friends can use that channel link and input it in WebTorrent to list and stream the episodes.
Which would be easier than sharing the individual torrent links.


This makes me think about the *list* feature in Twitter.
If we create something like that inside WebTorrent it will make it tremendously easy to share content, having each WebTorrent instance serve lists too.
We would be able to add torrents to lists and then share links to those lists.
Users interested in the content of a particular list would subscribe to it and access a curated list of torrents.
Maybe we can automatically create a new torrent with the list contents and share that instead.

codealchemist:
**Feature preview:**
Channels are accessed using the *Channels* main nav item:
image

Available channels list (I only added one):
image

Torrents list inside channel:
image

Channel Preferences (old UI):
image

DiegoRBaquero:
Given the already huge conflicts, I'll close this, but we can discuss the idea further in an issue if you'd want to

codealchemist:
We didn't get any traction on this one.
I'm totally fine letting it go 😄 
Probably it's a good idea to keep WebTorrent Desktop as simple as possible.
Thanks @DiegoRBaquero !

PR #804, opened 2016-08-22
Bugfixes

dcposch:
Here's our current list of top 10 bugs from https://webtorrent.io/desktop/telemetry/:

|Count|Example message|Status|
|------|------------------|------|
|2104|[object Object]|No idea, no stacktrace|
|632|[object HTMLVideoElement]|No idea, no stacktrace|
|359|Cannot read property 'getHostNode' of null|Happens after a React component throws an error in render(). Will go away by itself once we fix those errors, logged separately.|
|284|Failed to set the 'currentTime' property on 'HTMLMediaElement': The provided double value is non-finite.|**Fixed in this PR**|
|78|Cannot read property '0' of undefined|**Fixed in this PR**|
|62|Illegal wire type for field Message.Field...|Filed: https://github.com/thibauts/node-castv2/issues/35|
|59|torrent is destroyed|Bug in webtorrent.js, already filed|
|51|You can't connect to dlna when already connected to another device|Np. Comes from our own UI; should just use addError instead of throwing an uncaught exception.|
|48|val is not defined|Filed: https://github.com/thibauts/node-castv2/issues/36|
|47|Cannot read property 'message' of undefined|Bug in older version of WebTorrent, can't happen in the current telemetry.js|

### DIFF
--- a/src/renderer/controllers/playback-controller.js
+++ b/src/renderer/controllers/playback-controller.js
@@ -93,6 +93,10 @@ module.exports = class PlaybackController {
 
   // Skip (aka seek) to a specific point, in seconds
   skipTo (time) {
+    if (!Number.isFinite(time)) {
+      console.error('Tried to skip to a non-finite time ' + time)
+      return console.trace()
+    }
     if (isCasting(this.state)) Cast.seek(time)
     else this.state.playing.jumpToTime = time
   }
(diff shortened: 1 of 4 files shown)


dcposch:
@feross

feross:
LGTM!

feross:
This is amazing. Telemetry is extremely usefu! It would have taken forever to fix these issues without this, waiting for issues to get filed.

PR #806, opened 2016-08-23
Improve Preferences Page; Use Material UI

feross:
Let's use [Material UI](http://www.material-ui.com) to get consistent buttons and form inputs throughout the app.

New principles for our UI:

- Create components for anything that is reused.

- Components should use inline styles whenever possible (Let's shrink the size of `main.css` over time)

- Always require just the individual component that is needed from Material UI so that the whole library doesn't get loaded (important for app startup performance)

- Always apply the `'control'` class to material component buttons. Use this class on Material UI components to get correct native app behavior:
  - Dragging the button should NOT drag the entire app window
  - The cursor should be default, not a pointer (hand) like on the web


feross:
Before:

screen shot 2016-08-22 at 5 57 34 pm

After:

screen shot 2016-08-22 at 5 47 50 pm

Play in external player (note that help text updates):

screen shot 2016-08-22 at 5 47 56 pm

Pick alternative player (mplayer):

screen shot 2016-08-22 at 5 48 11 pm

Still working on it, but so far this feels a lot better and more consistent.

feross:
Improved the spacing a bit:

screen shot 2016-08-22 at 6 05 40 pm

dcposch:
looks *so much better*

dcposch:
for consistency we may want to drop a shadow from the title bar

feross:
> for consistency we may want to drop a shadow from the title bar

will play around with that

> i would also make the text in the input fields brighter, to match the label text

done

src/main/index.js
```diff
@@ -68,6 +68,8 @@ function init () {
   app.on('ready', function () {
     isReady = true
 
+    // require('electron').BrowserWindow.addDevToolsExtension('/Users/feross/Library/Application Support/Google/Chrome/Default/Extensions/fmkadmapgofadopljbjfkapdkoienihi/0.15.3_0/')
```
demoneaux: What's this? Interested to know :)

dcposch: probably React DevTools

feross: Yeah, sorry I'll remove that. It adds React Dev Tools to the Electron inspector.

feross: https://github.com/feross/webtorrent-desktop/commit/3f7e2c1e4a21eaebe18b8aaf3ef7b8dc52572978

PR #807, opened 2016-08-23
Fix playback + download of default torrents

dcposch:
There was a terrible bug introduced in https://github.com/feross/webtorrent-desktop/commit/0809e20a6e1de917a2f8315c2f1da3b9ef177a43 -- clicking play on any of the default torrents in a fresh install of the app would fail and result in a 'path missing' error.

This fixes the bug, and also adds a migration step to clean up resulting broken config files


feross:
LGTM

PR #813, opened 2016-08-23
Add 'Segoe UI'

avamsi:
Almost every other app in Windows uses 'Segoe UI' and it'd be nice if WebTorrent Desktop uses the same.

### DIFF
--- a/static/main.css
+++ b/static/main.css
@@ -21,7 +21,7 @@ body {
 
 body {
   color: #FFF;
-  font-family: BlinkMacSystemFont, 'Helvetica Neue', Helvetica, sans-serif;
+  font-family: BlinkMacSystemFont, 'Segoe UI', 'Helvetica Neue', Helvetica, sans-serif;
   font-size: 14px;
   line-height: 1.5em;
 }


feross:
Nice, makes a big difference!

feross:
Before:

screen shot 2016-08-23 at 2 51 48 am

After:

screen shot 2016-08-23 at 2 51 35 am

feross:
The only change I'm going to make to this PR is to put the 'Segoe UI' font in an `.app.is-win32` CSS selector so some poor designer who owns a Mac and has 'Segoe UI' installed doesn't somehow end up with the UI using that font. Because I'm not sure if the 'BlinkMacSystemFont' works on prior versions of OS X.

feross:
Fixed here: https://github.com/feross/webtorrent-desktop/commit/9ad0316dffe9ac189a310f4a42d0dd89fbd14259

avamsi:
Thanks for the fix and quick approval!

feross:
Will be part of the next release, probably v0.13.0 or v1.0

PR #817, opened 2016-08-23
Fix error with menus.

demoneaux:


src/main/menu.js
```diff
@@ -42,7 +42,11 @@ function setWindowFocus (flag) {
...
 function setAllowNav (flag) {
   getMenuItem('Preferences').enabled = flag
-  getMenuItem('Create New Torrent...').enabled = flag
+  getMenuItem(
+    process.platform === 'darwin'
+      ? 'Create New Torrent...'
+      : 'Create New Torrent from Folder...'
+  ).enabled = flag
   var item = getMenuItem('Create New Torrent from File...')
   if (item) item.enabled = flag
```
mathiasvr: Nice catch!
I have a suggestion for making it a little more readable, but I don't mind keeping it like this.
```javascript
if (process.platform === 'darwin') {
    getMenuItem('Create New Torrent...').enabled = flag
} else {
    getMenuItem('Create New Torrent from Folder...').enabled = flag
    getMenuItem('Create New Torrent from File...').enabled = flag
}
```

feross: Yeah, good catch @demoneaux. Thanks 👍 

I also prefer @mathiasvr's version. Will fix up after merge.

feross: https://github.com/feross/webtorrent-desktop/commit/b08d2739968f09685af9fe61bb862acd72762c82

demoneaux: :+1:

Issue #371, opened 2016-04-09
i18n


DiegoRBaquero:
We need i18n.

Thinking of https://github.com/mashpie/i18n-node

grunjol:
Totalmente de acuerdo

feross:
@grunjol También estoy de acuerdo

@DiegoRBaquero That looks like it's express specific.

DiegoRBaquero:
@feross It's not bound to express: https://github.com/mashpie/i18n-node#examples-for-common-setups, https://gist.github.com/mashpie/5188567

hicom150:
@RecoX I quote here your Multi-language implementation suggestion 😉

> Create some files under `Languages` folder for each language like `english.json`, `spanish.json`, etc and every file have in it all the texts required like
> 
> ```
> {
>      MAIN_WINDOW_CREATE_TORRENT: 'Create Torrent',
>      MAIN_WINDOW_DELETE_TORRENT: 'Delete Torrent',
> }
> ```
> 
> Then thanks to this webpage:
> https://www.logisticinfotech.com/translate-language-files-online-json/
> 
> with the json generated we can translated straight away to the language we want
> 
> We need to create a select input in the options menu with all the available languages in the application and that's it I think for now.
>

hicom150:
I enumerate here some popular React i18n libraries in case it helps to start "messing" with one 😉

- [i18nify](https://github.com/sealninja/react-i18nify) 🚲
- [react-intl](https://github.com/formatjs/react-intl) 🚗
- [i18next](https://github.com/i18next/react-i18next) 🚜

gfoligna:
I'm available for translations to Spanish. Most likely to check if the results make sense at all.

> @RecoX I quote here your Multi-language implementation suggestion 
> 
> > Create some files under `Languages` folder for each language like `english.json`, `spanish.json`, etc and every file have in it all the texts required like
> > ```
> > {
> >      MAIN_WINDOW_CREATE_TORRENT: 'Create Torrent',
> >      MAIN_WINDOW_DELETE_TORRENT: 'Delete Torrent',
> > }
> > ```
> > 
> > 
> > Then thanks to this webpage:
> > https://www.logisticinfotech.com/translate-language-files-online-json/
> > with the json generated we can translated straight away to the language we want
> > We need to create a select input in the options menu with all the available languages in the application and that's it I think for now.

KovalevArtem:
I suggest using **Weblate**

PR #820, opened 2016-08-23
Add support for i18n with react-intl [IN PROGRESS]

yciabaud:
This PR adds react-intl to translate the UI, I made a little experiment with one message translated to french.

Can you give me some feedback on this?

If it looks good to you, I can set a french translation up on the whole app and let the community translate other languages.

Issue #371


src/renderer/main.js
```diff
@@ -82,7 +84,30 @@ function onState (err, _state) {
   // Do this at least once a second to give every file in every torrentSummary
   // a progress bar and to keep the cursor in sync when playing a video
   setInterval(update, 1000)
-  app = ReactDOM.render(<App state={state} />, document.querySelector('#body'))
+
+  // Setup locale
+  var lang = navigator.language.split('-')[0] || 'en'
```
feross: Let's use `app.getLocale()`, Electron's official way to get the user's locale.

src/config.js
```diff
@@ -65,6 +65,8 @@ module.exports = {
   IS_PORTABLE: isPortable(),
   IS_PRODUCTION: isProduction(),
 
+  LOCALES_PATH: path.join(__dirname, '..', 'static', 'lang'),
```
feross: I don't see this folder checked into the repo. Also, can we call this `locales` instead of `lang`?

src/renderer/main.js
```diff
@@ -7,7 +7,9 @@ const dragDrop = require('drag-drop')
 const electron = require('electron')
 const React = require('react')
 const ReactDOM = require('react-dom')
+const ReactIntl = require('react-intl')
```
feross: Excellent library choice. Looks great!

feross:
@yciabaud Nice!! This looks excellent. Please continue working on this PR. Maybe we can get this done in time for v1!

For menu items, we may need to use the [FormatJS core libraries](http://formatjs.io/github/#core) directly, since those aren't written in React. But I suggest saving that work for a future PR to avoid overcomplicating things at the moment.

yciabaud:
Little update to fix inline comments (damn missing file!). OK when I have some time I will try to localize every message of the app.

yciabaud:
We will have to check the menus for windows and mac, I tested only with linux.

yciabaud:
Well I nearly localized everything except from some error messages and the window title. I am about to localize the titles but the navigation is broken since https://github.com/feross/webtorrent-desktop/pull/820/commits/b629c0dbb31fe847ad63115cf35d0320d4dd3298

Can you help figure what is going on?

I also have several warnings from standard coming from JSX templates, how to get rid of this?

src/renderer/controllers/prefs-controller.js
```diff
@@ -16,7 +18,8 @@ module.exports = class PrefsController {
       url: 'preferences',
       setup: function (cb) {
         // initialize preferences
-        state.window.title = 'Preferences'
+        state.window.title = new IntlMessageFormat(
```
dcposch: Since this looks like a common line, can we simplify it to

    i18n.format('preferences')

src/renderer/pages/App.js
```diff
@@ -82,8 +82,9 @@ class App extends React.Component {
     return (
       <div key='errors'
         className={'error-popover ' + (hasErrors ? 'visible' : 'hidden')}>
-        <div key='title' className='title'><FormattedMessage id='error'
-          defaultMessage='Error'/></div>
```
dcposch: Why do we need `defaultMessage` here (and the `||` defaults above)?

IMO English could just be one of the translation languages (and the default, in case we don't have translations for the user's system language yet)

yciabaud: `defaultMessage` is part of the [React-Intl Message Formatting Fallbacks](https://github.com/yahoo/react-intl/wiki/API#message-formatting-fallbacks), this ensures a message is always retrieved (think of incomplete translation). We could load an english lang file and overwrite it with the local labels but that is not the way the library is designed.

I find it convenient in code navigation and readability to have the labels in the code but if you prefer we can setup this in another way.

What do you think? Can you have a look at my issue refreshing the view? I am a bit stuck ATM...

src/renderer/controllers/prefs-controller.js
yciabaud: Yes, it would be :
```
i18n.format('id', 'defaultmessage', [{formatKey: formatValue])
```

A bit cleaner!

DiegoRBaquero:
Given the number of conflicts and stalled PR, I'm going to close this.

PR #821, opened 2016-08-23
fix player poster image css bug

mathiasvr:
When playing audio files the poster was not displayed.


feross:
LGTM

Nice catch -- can't believe that we didn't notice this before now. We really need to get autoplay working so the app is more useful for playing back entire albums. I've actually never listened all the way through the Creative Commons WIRED CD using WebTorrent yet, because there's no autoplay!

PR #823, opened 2016-08-23
Don't render player controls when using external player

### LINKED ISSUE #790: Do not render player controls when external player is in use

master

All

**What did you do?**

Start video in external player.

WebTorrent Desktop's UI should not mislead me into thinking that I can control the playback from within WebTorrent.

**What actually happened?**

Controls were shown and I thought I could control the playback from there. This is particularly misleading when the file is audio. Then, the external player doesn't even show a UI so all you see is the WebTorrent UI.

Example, here's just the audio from Big Buck Bunny:

screen shot 2016-08-19 at 11 26 46 pm

mathiasvr:
Closes #790


feross:
LGTM

PR #824, opened 2016-08-23
Don't open partially downloaded files in external app

### LINKED ISSUE #810: Partially downloaded files should not be openable in external programs

0.12.0

All

**What did you do?**

Clicked on a partially downloaded audio file.

File should not open.

**What actually happened?**

File opened in iTunes which proceeded to play half of it, then silence for the rest of the file.

cc @dcposch

mathiasvr:
Fixes #810 

It might be cleaner to handle this directly in the PlaybackController`openItem` method, or to create a noop dispatcher function?

### DIFF
--- a/src/renderer/controllers/playback-controller.js
+++ b/src/renderer/controllers/playback-controller.js
@@ -38,7 +38,7 @@ module.exports = class PlaybackController {
     })
   }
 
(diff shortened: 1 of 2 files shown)


feross:
LGTM

PR #825, opened 2016-08-23
Don't render radial progress on path-missing error

mathiasvr:
If the torrent folder has been deleted we show a path missing error and do not render the play and download buttons, however the radial progress (`positionElem`) was still being displayed.

This PR fixes that, but the `renderTorrentButtons` function could still use some refactoring.

### DIFF
--- a/src/renderer/pages/TorrentListPage.js
+++ b/src/renderer/pages/TorrentListPage.js
@@ -184,33 +184,9 @@ module.exports = class TorrentList extends React.Component {
       downloadTooltip = 'Click to start torrenting.'
     }
 
-    // Do we have a saved position? Show it using a radial progress bar on top
-    // of the play button, unless already showing a spinner there:
-    var positionElem
-    var willShowSpinner = torrentSummary.playStatus === 'requested'
-    var defaultFile = torrentSummary.files &&
-      torrentSummary.files[torrentSummary.defaultPlayFileIndex]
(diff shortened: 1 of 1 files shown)


feross:
LGTM 

@mathiasvr you're on a roll today! Nice work! 👍 👍 👍

src/renderer/pages/TorrentListPage.js
```diff
@@ -220,6 +196,29 @@ module.exports = class TorrentList extends React.Component {
           {downloadIcon}
         </i>
       )
+
+      if (TorrentPlayer.isPlayableTorrentSummary(torrentSummary)) {
+        playButton = (
+          <i
+            key='play-button'
+            title={playTooltip}
+            className={'button-round icon play ' + playClass}
```
feross: Just realized there's a bug here. `playClass` is used here, but it's defined later, on line 220. So it gets set to `undefined` which is causing a rendering a bug.

screen shot 2016-08-23 at 5 23 49 pm

The radial is getting rendered *under* the play button instead of over it.

feross: Fixed in https://github.com/feross/webtorrent-desktop/commit/83901eecbaa8441d6dd0d07d104d168a9a5df17e

mathiasvr: Shame on me, thought I was so clever with my last second code reordering.
Thanks for fixing!!!

PR #827, opened 2016-08-24
remember window position

### LINKED ISSUE #181: Window position should be restored when app is restarted

Flet:
Closes #181

- Save the bounds of the window when its resized or moved.
- Stored in config.json under "bounds" key (state.saved.bounds)

One oddity: the window is shown at its default centered position until the renderer starts and the state gets loaded, then the window is moved to its saved bounds. 

Setting the bounds when the window is created is possible, but it means loading the state way earlier... (and possibly loading it again once the renderer starts). Any thoughts on acceptable ways to solve this/mitigate the jank at startup?

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -18,7 +18,7 @@ console.log('Production: %s portable: %s test: %s',
   IS_PRODUCTION, IS_PORTABLE, IS_TEST)
 if (IS_PORTABLE) console.log('Portable path: %s', PORTABLE_PATH)
 
-module.exports = {
+let cfg = {
   ANNOUNCEMENT_URL: 'https://webtorrent.io/desktop/announcement',
   AUTO_UPDATE_URL: 'https://webtorrent.io/desktop/update',
   CRASH_REPORT_URL: 'https://webtorrent.io/desktop/crash-report',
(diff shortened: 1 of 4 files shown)


Flet:
Oops. I just saw #626 :/

Flet:
OK, after reading the comments on #626, loading the state earlier in main/index.js and passing it to the init() of the main window. This gets rid of the jank at startup. The state is loaded a second time in the renderer now, but maybe thats ok? :) Is there a clean way to pass it along from the main window?

src/main/windows/main.js
```diff
@@ -26,10 +26,17 @@ var tray = require('../tray')
 var HEADER_HEIGHT = 38
 var TORRENT_HEIGHT = 100
 
-function init () {
+function init (state) {
   if (main.win) {
     return main.win.show()
   }
+
+  var defaultBounds = {
```
dcposch: This could even be `config.DEFAULT_BOUNDS`. If you move it there, I'd also replace the `TORRENT_HEIGHT * 6` with something like `TORRENT_HEIGHT * (DEFAULT_TORRENTS.length + 1)`

dcposch:
LGTM

dcposch:
I left a comment. Also recommend you squash and rebase, then feel free to merge.

Here's a way to do it that results in just a single commit on master, no merge commit.

```
git fetch
git rebase -i origin/master # squash as needed
git push --force
# a bit later, once CI is finished:
git push origin window-position:master
```

mathiasvr:
@dcposch is this different from pressing "Squash and merge"?

Flet:
Thanks for reviewing @dcposch, I rebased and made updates.

src/config.js
```diff
@@ -93,9 +93,19 @@ module.exports = {
   WINDOW_WEBTORRENT: 'file://' + path.join(__dirname, '..', 'static', 'webtorrent.html'),
 
   WINDOW_MIN_HEIGHT: 38 + (120 * 2), // header height + 2 torrents
```
dcposch: This should use UI_HEADER_HEIGHT and UI_TORRENT_HEIGHT

PR #830, opened 2016-08-24
ImageOptim: losslessly compress all images

feross:


demoneaux:
LGTM as there doesn't seem to be any noticeable change in quality. Also I noticed that [loading.gif](https://github.com/feross/webtorrent-desktop/blob/f9aeb676b46ad746b3b8ab65818db56b69b04ba5/static/loading.gif) doesn't use the official logo as the background?

feross:
> LGTM as there doesn't seem to be any noticeable change in quality. 

Yes, it's lossless compression. No loss in quality. [ImageOptim](https://imageoptim.com/mac) is great!

> Also I noticed that loading.gif doesn't use the official logo as the background?

Yep, intentional. It's the installer spalsh screen, so it's meant to look "half finished" 😉

PR #832, opened 2016-08-24
Allow people to opt out of telemetry.

### LINKED ISSUE #766: Opt-out for telemetry
Is there any possibility to disable telemetry? Would be nice to opt-out of this option.

demoneaux:
Closes #766.

### DIFF
--- a/src/renderer/lib/migrations.js
+++ b/src/renderer/lib/migrations.js
@@ -35,6 +35,10 @@ function run (state) {
     migrate_0_12_0(state.saved)
   }
 
+  if (semver.lt(version, '0.13.0')) {
+    migrate_0_13_0(state.saved)
+  }
+
   // Config is now on the new version
   state.saved.version = config.APP_VERSION
 }
@@ -134,4 +138,11 @@ function migrate_0_12_0 (saved) {
       delete torrentSummary.path
     }
   })
+}
+
+function migrate_0_13_0 (saved) {
+  // Enable telemetry by default
+  if (saved.prefs.optInTelemetry == null) {
+    saved.prefs.optInTelemetry = true
+  }
 }
--- a/src/renderer/pages/PreferencesPage.js
+++ b/src/renderer/pages/PreferencesPage.js
@@ -21,6 +21,9 @@ class PreferencesPage extends React.Component {
 
     this.handleExternalPlayerPathChange =
       this.handleExternalPlayerPathChange.bind(this)
+
+    this.handleOptInTelemetryChange =
+      this.handleOptInTelemetryChange.bind(this)
   }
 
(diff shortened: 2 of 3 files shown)


src/renderer/pages/PreferencesPage.js
```diff
@@ -114,6 +117,23 @@ class PreferencesPage extends React.Component {
     // dispatch('updatePreferences', 'isFileHandler', !isFileHandler)
   }
 
+  optInTelemetryCheckbox () {
+    return (
+      <Preference>
+        <Checkbox
+          className='control'
+          checked={this.props.state.unsaved.prefs.optInTelemetry}
+          label={'Opt in to telemetry'}
```
feross: "Automatically send anonymous usage statistics"

src/renderer/lib/migrations.js
```diff
@@ -135,3 +139,10 @@ function migrate_0_12_0 (saved) {
     }
   })
 }
+
+function migrate_0_13_0 (saved) {
+  // Enable telemetry by default
+  if (saved.prefs.optInTelemetry == null) {
+    saved.prefs.optInTelemetry = true
```
feross: Can you also add this to the default state in `state.js` so that users who install versions after 0.12 will get defaulted to `true`.

src/renderer/pages/PreferencesPage.js
```diff
@@ -133,6 +153,9 @@ class PreferencesPage extends React.Component {
         <PreferencesSection title='Default torrent app'>
           {this.setDefaultAppButton()}
         </PreferencesSection>
+        <PreferencesSection title='Telemetry'>
```
feross: I would title this 'Privacy' instead of 'Telemetry' since most users won't know what the latter means.

feross:
I'm okay with this change, though as I said in #766, I do think it's a bit overkill since the data we're gathering isn't personally identifying in any way.

IMO, one thing we should change: if the user opts out of telemetry, we should still send the most basic info: (1) the randomly-generated user id (so we can count total number of users) and (2) the app version (so we can ensure the auto-updater works).

The rest of the info derived from the user's system should be excluded (like screen size, operating system, local time, approx. number of torrents, etc.)

@dcposch - what do you think?

mathiasvr:
Just make this feature a version 5.0 milestone 😆

feross:
@mathiasvr :trollface:

demoneaux:
Not sure if this is still necessary: there doesn't seem to be much demand for opting out of telemetry, and also the data collected is not personal information and does not compromise the user. What do you all think?

feross:
Agreed. Let's close this for the time being.

PR #836, opened 2016-08-25
Telemetry: log version in errors

dcposch:
@mathiasvr @Flet

### DIFF
--- a/src/renderer/lib/telemetry.js
+++ b/src/renderer/lib/telemetry.js
@@ -33,7 +33,7 @@ function init (state) {
 
   if (config.IS_PRODUCTION) {
     postToServer()
-    // If the user keeps WebTorrent running for a long time, post every 24h
+    // If the user keeps WebTorrent running for a long time, post every 12h
     setInterval(postToServer, 12 * 3600 * 1000)
   } else {
     // Development: telemetry used only for local debugging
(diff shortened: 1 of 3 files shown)


dcposch:
@feross

src/renderer/lib/telemetry.js
```diff
@@ -118,29 +119,60 @@ function getApproxNumTorrents (state) {
...
-  var message, stack
-  if (err instanceof Error) {
-    message = err.message
-    // Remove the first part of each file path in the stack trace.
-    // - Privacy: remove personal info like C:\Users\<full name>
-    // - Aggregation: this lets us find which stacktraces occur often
-    stack = err.stack.replace(/\(.*app.asar/g, '(...')
+  var message
+  var stack = ''
+  if (e.message) {
```
mathiasvr: I don't think that `e` can be null/undefined now, but maybe we should check that anyway to be sure. We can always remove it if we don't see anything in the telemetry reports, but if it happens the report will never be logged or sent.

mathiasvr:
LGTM

src/renderer/lib/telemetry.js
feross: Not a bad idea.

Issue #835, opened 2016-08-25
Start the app automatically on boot


dcposch:
We should add a pref, on by default, that starts WebTorrent Desktop automatically. That helps keep the swarms healthy.

noamokman:
I'm on it, (#837)
Would love to have your input on this @dcposch

obv-mikhail:
Not sure whether it should be on by default. Seems intrusive?

dcposch:
@obv-mikhail will be off by default for now.

For comparison, though, uTorrent has a similar pref that's on by default (people even complained because they changed it to auto-start again with every app update!)

iovis9:
I'd definitely NOT put it on by default ever.

grunjol:
Reasons against set as default are already mentioned in the feature branch https://github.com/feross/webtorrent-desktop/pull/837/commits/d6920c69c9953669d9cedd1783970451fa1b3336

demoneaux:
Closing as this was fixed in #923.

PR #837, opened 2016-08-26
Startup option in preferences (WIP)

noamokman:
Hi there,
This is about #835,
Still WIP,
Need to test and create startup on first boot or install.

Would love to help more along the way :smile: 
Thanks,
Noam

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -21,6 +21,7 @@ const windows = require('./windows')
 
 let shouldQuit = false
 let argv = sliceArgv(process.argv)
+const hidden = argv.includes('--hidden')
 
 if (config.IS_PRODUCTION) {
   // When Electron is running in production mode (packaged app), then run React
@@ -68,7 +69,7 @@ function init () {
   app.on('ready', function () {
     isReady = true
 
-    windows.main.init()
+    windows.main.init({hidden: hidden})
     windows.webtorrent.init()
     menu.init()
 
@@ -156,6 +157,8 @@ function processArgv (argv) {
       dialog.openTorrentFile()
     } else if (arg === '-u') {
       dialog.openTorrentAddress()
+    } else if (arg === '--hidden') {
+      // Igonre hidden argument, already being handled
     } else if (arg.startsWith('-psn')) {
       // Ignore Mac launchd "process serial number" argument
       // Issue: https://github.com/feross/webtorrent-desktop/issues/214
--- a/src/main/startup.js
+++ b/src/main/startup.js
@@ -1,0 +1,27 @@
+module.exports = {
+  install,
+  uninstall
+}
+
+const config = require('../config')
+const AutoLaunch = require('auto-launch')
+
+const appLauncher = new AutoLaunch({
+  name: config.APP_NAME,
+  isHidden: true
+})
+
+function install () {
+  return appLauncher.isEnabled()
+      .then(enabled => {
+        if (enabled) return
+        return appLauncher.enable()
+      })
+}
+
+function uninstall () {
+  return appLauncher.isEnabled()
+      .then(enabled => {
+        if (enabled) return appLauncher.disable()
+      })
+}
--- a/src/renderer/controllers/prefs-controller.js
+++ b/src/renderer/controllers/prefs-controller.js
@@ -17,7 +17,9 @@ module.exports = class PrefsController {
       setup: function (cb) {
         // initialize preferences
         state.window.title = 'Preferences'
-        state.unsaved = Object.assign(state.unsaved || {}, {prefs: state.saved.prefs || {}})
+        state.unsaved = Object.assign(state.unsaved || {}, {
+          prefs: Object.assign({}, state.saved.prefs)
+        })
         ipcRenderer.send('setAllowNav', false)
         cb()
       },
@@ -49,6 +51,9 @@ module.exports = class PrefsController {
     const state = this.state
     if (state.unsaved.prefs.isFileHandler !== state.saved.prefs.isFileHandler) {
       ipcRenderer.send('setDefaultFileHandler', state.unsaved.prefs.isFileHandler)
+    }
+    if (state.unsaved.prefs.startup !== state.saved.prefs.startup) {
+      ipcRenderer.send('setStartup', state.unsaved.prefs.startup)
     }
     state.saved.prefs = Object.assign(state.saved.prefs || {}, state.unsaved.prefs)
     State.save(state)
--- a/src/renderer/lib/state.js
+++ b/src/renderer/lib/state.js
@@ -107,7 +107,8 @@ function setupSavedState (cb) {
       downloadPath: config.DEFAULT_DOWNLOAD_PATH,
       isFileHandler: false,
       openExternalPlayer: false,
-      externalPlayerPath: null
+      externalPlayerPath: null,
+      startup: false
     },
     torrents: config.DEFAULT_TORRENTS.map(createTorrentObject),
     version: config.APP_VERSION /* make sure we can upgrade gracefully later */
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
   "dependencies": {
     "airplayer": "^2.0.0",
     "application-config": "^1.0.0",
+    "auto-launch": "^4.0.0",
     "bitfield": "^1.0.2",
     "chromecasts": "^1.8.0",
     "create-torrent": "^3.24.5",
--- a/src/main/ipc.js
+++ b/src/main/ipc.js
@@ -17,6 +17,7 @@ const shortcuts = require('./shortcuts')
 const externalPlayer = require('./external-player')
 const windows = require('./windows')
 const thumbar = require('./thumbar')
+const startup = require('./startup')
 
 // Messages from the main process, to be sent once the WebTorrent process starts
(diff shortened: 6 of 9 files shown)


src/main/startup.js
```diff
@@ -0,0 +1,16 @@
...
+  install,
+  uninstall
+}
+
+var startup = require('user-startup')
+
+function install () {
+  var appPath = process.execPath
+
+  startup.add('webtorrent-desktop', appPath)
```
feross: Does this work on Windows and OS X?

.gitignore
```diff
@@ -2,3 +2,4 @@ node_modules/
 build/
 dist/
 npm-debug.log.*
+.idea
```
feross: what is this? can you remove it?

noamokman: it's a folder used by webstorm (actually all jetbrains products)

src/main/startup.js
noamokman: Yes

noamokman: waiting for typicode/user-startup#7 to land,
apparently all params are required.

noamokman:
@feross where do you think is a good place to have the initial startup installation?
The default value is `true` so we need to run `startup.install` at least once.

.gitignore
grunjol: You can use `$HOME/.config/git/ignore` or `$GIT_DIR/info/exclude` to include your editor config files (i use eclipse and my .settings and .project files are there) to avoid populate them in every project

noamokman:
switched to use [auto-launch](https://www.npmjs.com/package/auto-launch) due to better electron support

noamokman:
@grunjol thanks!
removed the .idea from the gitignore file

feross:
@dcposch do you have thoughts on this?

src/renderer/lib/state.js
```diff
@@ -107,7 +107,8 @@ function setupSavedState (cb) {
       downloadPath: config.DEFAULT_DOWNLOAD_PATH,
       isFileHandler: false,
       openExternalPlayer: false,
-      externalPlayerPath: null
+      externalPlayerPath: null,
+      startup: true
```
grunjol: Default should not be `true` IMHO

src/renderer/controllers/prefs-controller.js
```diff
@@ -17,7 +17,7 @@ module.exports = class PrefsController {
       setup: function (cb) {
         // initialize preferences
         state.window.title = 'Preferences'
-        state.unsaved = Object.assign(state.unsaved || {}, {prefs: state.saved.prefs || {}})
```
dcposch: Was this causing problems?

src/renderer/lib/state.js
dcposch: Why not? There's a pref to disable

Looks like uTorrent not only starts automatically, but it re-enables that option after every auto-update (!) https://forum.utorrent.com/topic/92281-disable-utorrent-program-auto-start-as-default-setting/

Obv we don't want to do that---if someone turns off the pref, we'll respect it permanently---but I think auto start after they first install WebTorrent is fine.

dcposch:
LGTM! @noamokman thanks for doing this!

For future PRs that add new UI (in this case, new controls on the Prefs page), consider posting a screenshot. Github lets you just paste images directly into a PR, so on Windows you can open the prefs page -> Alt + PrintScr -> paste into Github

src/renderer/controllers/prefs-controller.js
noamokman: yeah, there was an immutability problem,
the saved and unsaved perfs object was the same one.
Fixed :smile:

noamokman:
@dcposch cool! it looks something like that:
image

noamokman:
Hi everyone,
That's the remaining work:
* test startup on all platforms
    * windows
    * linux
    * mac
* disable feature on portable - should I hide hide the pref section ui or show it as disabled?
* add migrations function to handle the startup value?
* run install on app load (install checks if the auto launch is already enabled first so its safe to run it every single time.


I'm having problems packaging the app on my machine.
So helping at testing the startup on other platforms is appreciated :smile: 
Still need to decide about the other stuff on the list though.

src/renderer/lib/state.js
grunjol: Let me be clear, I'm not against create a startup feature, but do not force the user beforehand.
* The application is not a daemon, so the user should indicate when start.
* The user will be surprised with the first autostart (even more if its a previous user).
* A lot of users care about their privacy, hence use VPNs, (manually started in most of the cases) so first autostart will be offensive. 
* Even in the forum post you mention, the users ask for the same expected behavior
image
he is not complaining about not respecting user's preference but the default behavior.

grunjol:
You don't  need to implement the last 2 if the default behavior is not run at startup.
I can test it in debian and ubuntu.

noamokman:
Yeah, that's true.
Thanks!

noamokman:
@dcposch @grunjol @feross 
Final verdict on the default value for startup?
I wanna make adjustments and hopefully merge

dcposch:
> Final verdict on the default value for startup?

We can have it off by default, for now, as @grunjol suggested. Sounds like it'll be easier to ship that way. If we make it the default in the future, it will only affect new installs (not existing users). CC @feross 

> disable feature on portable - should I hide hide the pref section ui or show it as disabled?

I'd just hide it.

> test startup on all platforms

I can test on Mac and Windows

src/main/startup.js
```diff
@@ -0,0 +1,25 @@
+module.exports = {
+  install,
+  uninstall
+}
+
+var AutoLaunch = require('auto-launch')
+
+var appLauncher = new AutoLaunch({
+  name: 'WebTorrent Desktop'
```
grunjol: you should get the app name from config

grunjol:
image

auto-start on linux removes only grabs the first word (WebTorrent) to create the file and launcher... not a big deal.

grunjol:
Working fine on Ubuntu and Debian GNOME.

It starts the app with torrent list. I am guessing... can we make it start 'minimized' from autostart? (maybe with a command line)

noamokman:
> Working fine on Ubuntu and Debian GNOME.

Great :+1:  thanks

>  starts the app with torrent list. I am guessing... can we make it start 'minimized' from autostart? (maybe with a command line)

on it!

grunjol:
Reading a little more, it seems it already has an `isHidden: true` option.
which basically adds a `--hidden` to the command line.

dcposch:
On Mac, I'm getting:

    module.js:442 Uncaught Error: Cannot find module '../components/Header'

...but only when running the packaged app. `npm start` works fine.

grunjol:
already reported https://github.com/feross/webtorrent-desktop/issues/842

grunjol:
@noamokman `processArgv` should digest the `--hidden` parameter and keep it to prevent `main.show()` being excuted on `function onOpen (e, torrentId)`

noamokman:
@grunjol oops, didnt see the messages while working on it, check the latest commit.
opening a torrent should show the main window.

dcposch:
After rebasing on top of #843, auto start works on Mac, but it pops up **two** windows. One is the WebTorrent window (should start minimized), the other is a terminal:

screen shot 2016-08-27 at 9 17 20 am

noamokman:
@dcposch weird... maybe merge #843, I'll rebase my branch.
The window should be hidden now. dont know about the terminal though

dcposch:
Works great on Windows!

Looks like there are three things left to do:
- [ ] disable feature, hide from prefs section in the portable windows app
- [ ] auto-start minimized
- [ ] don't start a terminal on Mac

noamokman:
> disable feature, hide from prefs section in the portable windows app
 auto-start minimized

these two should are pushed

noamokman:
The implementation for linux and windows is an hidden argument.
Not sure what this does for mac:
https://github.com/Teamwork/node-auto-launch/blob/master/src/AutoLaunchMac.coffee#L10

grunjol:
i think you get this wrong, you need to add `isHidden: true` option in auto-launch

noamokman:
@grunjol yeah I did, that argument is used to optionally add the `--hidden` agrument on startup.
not sure it works the same on mac

src/main/index.js
```diff
@@ -21,6 +21,7 @@ var windows = require('./windows')
 
 var shouldQuit = false
 var argv = sliceArgv(process.argv)
+var hidden = argv.includes('--hidden')
```
grunjol: instead of set hidden value here, do it on `function processArgv (argv)`

dcposch:
> these two are already pushed

sweet, i'll test again

grunjol:
https://github.com/Teamwork/node-auto-launch/issues/28

Check the latest message it has a possible solution

noamokman:
waiting for Teamwork/node-auto-launch#38 to land :+1: 
also rebased the branch.
will now sort the process args problem

src/main/index.js
noamokman: need to leave it here so that the main windows starts hidden,
placing the hide function in the processArgv function makes the window flash for a sec.
Made the processArgv function ignore the value as a torrent id

dcposch:
Rebased, added some fixes, merged. See #923 . Thanks @noamokman !

PR #838, opened 2016-08-26
fix: add missing console.timeEnd

ungoldman:
Noticed the webtorrent renderer instance has a `console.time('init')` with a missing `console.timeEnd('init')` counterpart. Thought I'd add it in at the end of the `init()` function.


feross:
Nice, LGTM!

Issue #362, opened 2016-04-08
When player is active, other downloads should temporarily stop


feross:
So all download capacity goes to what the user actually wants to see.

ungoldman:
Disagree! At least for me that is not desirable.

rom1504:
Maybe there could be an option in the preferences window https://github.com/feross/webtorrent-desktop/pull/144

karloluis:
Maybe throttle the rest? This really puts more stress on the preferences issue, specially since the goal is to keep it minimal.

ungoldman:
I think a pause all button would be a good feature, would leave the
decision of how to manage downloads and streaming to the user.

dcposch:
I think we can solve this with a global priority for pieces to download, across all active torrents.

While you're streaming, pieces within a few megabytes of where you're currently playing should get top priority. That way, your media should never pause to buffer as long as your internet connection (and the seeders) are fast enough.

Pieces that are far ahead of where you're currently playing can have the same priority as other torrents running in the background.

@feross thoughts? 

@ungoldman to your point, I think this could be more elegant than just pausing all other torrents while one is streaming. At the same time you never want to your video to buffer because other torrents are taking up too much bandwidth

feross:
@dcposch This issue isn't really about piece selection priority as much as throttling. Piece selection priority affects what pieces we ask peers for next. Each torrent currently asks as fast as it can, based on the measured connection speeds, always asking for a little bit more data than the peer is currently sending.

Each torrent object in WebTorrent is isolated. One torrent object doesn't know anything about another. So, to solve this issue correctly, we need to add per-torrent throttling and a notion of torrent priority to WebTorrent.

We can probably devise a balanced solution where priority for the background torrents increases as the foreground torrent builds up a buffer. But this is a non-trivial change. 

In the meantime, I propose we do the right thing for 99% of users, and that is to make sure streaming works reliably.

Steam does this too. When you launch a game, all game downloads pause so the foreground game can use 100% of the available bandwidth.

That model makes sense for WebTorrent too. When the active torrent is not fully downloaded, the full bandwidth should be made available to it. Once it completes, the background torrent downloads can resume. Also, if the torrent is already completed when the playback beings, then background torrent downloading should not be interrupted.

Moscarda:
[...]

PR #840, opened 2016-08-27
Playback priority

codealchemist:
Added highest playback priority feature; added highest playback priority checkbox on preferences.
Pauses all active torrents when playback starts, giving all bandwidth to the streaming torrent.
Restores paused torrent when playback stops.
This is the _easy way_ to solve #362.


noamokman:
I think we need to resume downloads when the torrent you are watching finished downloading.
It looks like it only resumes after you close the player, am I right?

codealchemist:
Good catch @noamokman!
Yeah, on this PR the paused downloads resume only after the player stops.
It will be better if it also resumes when the streaming torrent finishes downloading.
I'll look into it, thanks!

codealchemist:
@noamokman, now it resumes on `wt-done`, which seems to be working as expected.

noamokman:
Awesome! Great work :smile:

codealchemist:
Thanks :)

codealchemist:
@feross if this is still needed I can spend some time fixing conflicts.
Thx.

toinouu:
Exactly what I want 
+1

codealchemist:
Thanks @toinouu!
Still waiting for team approval before fixing conflicts and merge.
:)

dcposch:
@codealchemist Sorry we missed this. Reviewing today

codealchemist:
No worries @dcposch :)
Thanks for reviewing it!

src/renderer/controllers/torrent-list-controller.js
```diff
@@ -103,18 +103,73 @@ module.exports = class TorrentListController {
     }
   }
 
+  pauseAll ({filter, excluded}) {
```
dcposch: What about making this
```
prioritizeTorrent (infohash) {
```

...the filter arg is always just active torrents. Don't know if we need the ability to pause torrents by arbitrary JSON filter. The excluded arg is always an array containing a single infohash: the torrent we're playing.

codealchemist:
Thanks a lot for your review @dcposch!
I'll work on it.
Cheers!

codealchemist:
Tests passing.
Manual test looks good.

feross:
🎉 🎉 🎉 🎉

PR #841, opened 2016-08-27
use depcheck

noamokman:
Hi all,
This has been bugging me a bit,
we can use the [depcheck](https://www.npmjs.com/package/depcheck) instead of writng our own dependency check.
Thanks :smile:


noamokman:
p.s. this makes the deps check work on windows

dcposch:
LGTM

dcposch:
@noamokman you also want `--ignore-dirs=build,dist`. I'll add that

Issue #842, opened 2016-08-27
Cannot find module '../components/Header'


grunjol:
**What version of WebTorrent Desktop?** (See the 'About WebTorrent' menu)
master
**What operating system and version?**
Ubuntu 16.04
**What did you do?**
Open the App
**What did you expect to happen?**
Start correctly
**What actually happened?**
`module.js:442 Uncaught Error: Cannot find module '../components/Header'`

I think the problem is component name is lowercase header. Linux is case sensitive and can not find the file

dcposch:
Fixed in #843

PR #843, opened 2016-08-27
Fix npm run package

dcposch:
A require() had the wrong case, which apparently works for `npm start` build but fails in the packaged app

@noamokman @mathiasvr


dcposch:
Fixes #842

grunjol:
Same happens on linux, this patch fixes the issue.

Issue #845, opened 2016-08-27
Unable to create torrent


PurgingPanda:
**What version of WebTorrent Desktop?** (See the 'About WebTorrent' menu)
Version 0.12.0 (0.96.4)

**What operating system and version?**
Ubuntu GNOME 16.04

**What did you do?**
Dragged and dropped files on the window. Then on create new torrent display, I clicked on create torrent.

**What did you expect to happen?**
That the new torrent got created and the application went back to the overview screen.

**What actually happened?**
Nothing visible.
Errors in the console:
```CreateTorrentPage.js:87 Uncaught TypeError: Cannot read property 'value' of nullReactErrorUtils.invokeGuardedCallback @ ReactErrorUtils.js:70
telemetry.js:160 Uncaught TypeError: Cannot read property 'length' of undefinedReactErrorUtils.invokeGuardedCallback @ ReactErrorUtils.js:70
main.js:275 toggleDevTools
```

dcposch:
Just tested this, and:
* It works for me in the current released app (0.12)
* It fails with the same TypeError listed above on the master branch

@PurgingPanda just to confirm, did you clone the repo and run `npm start`, or did you install from the deb file?

PurgingPanda:
I cloned the repo, ran ```npm install``` and finally ```npm run build && electron .```.

dcposch:
Looks like this is fixed in #851 and released, let us know if you run into this error again on the latest version of the app. Thanks for the bug report

PR #851, opened 2016-08-29
Telemetry: handle undefined values

mathiasvr:
Follow-up on #836. Ensures that e/stack/message is not undefined.

Also fixes a bug present in  #845 where the stack was `undefined`.


dcposch:
LGTM

PR #855, opened 2016-08-29
WIP: add embedded mkv subtitles

mathiasvr:
This PR should provide basic support for loading embedded mkv subtitles.
- The subtitles will only be loaded if the file is 100% downloaded.
- The subtitle language is determined by the `languagedetect` module and does not use mkv metadata.

Currently the code is overly complicated, because the embedded subtitles are loaded in a different way than subtitle files. `languagedetect` has not worked very well for me so maybe it's better to use the mkv metadata.

Also I wonder how often people actually uses mkv's with embedded subtitles that are also playable by WebTorrent Desktop?

### DIFF
--- a/src/renderer/controllers/subtitles-controller.js
+++ b/src/renderer/controllers/subtitles-controller.js
@@ -2,8 +2,10 @@ const electron = require('electron')
 const fs = require('fs-extra')
 const path = require('path')
 const parallel = require('run-parallel')
+const zeroFill = require('zero-fill')
 
 const remote = electron.remote
+const ipcRenderer = electron.ipcRenderer
 
 const {dispatch} = require('../lib/dispatcher')
 
@@ -74,6 +76,7 @@ module.exports = class SubtitlesController {
     torrentSummary.progress.files.forEach((fp, ix) => {
       if (fp.numPieces !== fp.numPiecesPresent) return // ignore incomplete files
       var file = torrentSummary.files[ix]
+      if (this.state.playing.fileIndex === ix) return this.checkForEmbeddedMKVSubtitles(file)
       if (!this.isSubtitle(file.name)) return
       var filePath = path.join(torrentSummary.path, file.path)
       this.addSubtitles([filePath], false)
@@ -85,12 +88,31 @@ module.exports = class SubtitlesController {
     var ext = path.extname(name).toLowerCase()
     return ext === '.srt' || ext === '.vtt'
   }
+
+  checkForEmbeddedMKVSubtitles (file) {
+    var playing = this.state.playing
+    // var playingFile = this.state.getPlayingFileSummary()
+    // var playingPath = path.join(torrentSummary.path, playingFile.path)
+
+    if (path.extname(file.name).toLowerCase() === '.mkv') {
+      ipcRenderer.send('wt-get-mkv-subtitles', playing.infoHash, playing.fileIndex)
+
+      ipcRenderer.once('wt-mkv-subtitles', function (e, tracks) {
+        tracks.forEach(function (trackEntry) {
+          var track = loadEmbeddedSubtitle(trackEntry)
+          console.log('loaded emb subs', track)
+          playing.subtitles.tracks.push(track)
+        })
+
+        if (tracks.length > 0) relabelSubtitles(playing.subtitles)
+      })
+    }
+  }
 }
 
 function loadSubtitle (file, cb) {
   // Lazy load to keep startup fast
   var concat = require('simple-concat')
-  var LanguageDetect = require('languagedetect')
   var srtToVtt = require('srt-to-vtt')
 
   // Read the .SRT or .VTT file, parse it, add subtitle track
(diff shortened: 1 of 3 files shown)


dcposch:
> Also I wonder how often people actually uses mkv's with embedded subtitles that are also playable by WebTorrent Desktop?

I'll add more detailed play attempts to the telemetry

dcposch:
LGTM

I wonder if we should extract a media player module from WebTorrent Desktop.

Here's what it would do:
* Manage the <video> or <audio> tag
* Provide a React-friendly way to control playback, including volume, seek, subtitles etc. In other words, it would encapsulate some of the hacks we currently have in `player-page.js` and `playback-controller.js`
* Handle edge cases (for example, pausing media when the window is closed)
* Handle hardsubs
* Handle casting
* Handle as many audio and video containers and codecs as possible

Here's what it wouldn't do:
* Show its own UI. The user of the module (starting with WT Desktop) would still have to chose their own rendering library (in our case, React), write their own player controls, etc.

But the complexity of making a full-featured audio / video player for an Electron app would be handled under the hood.

Thoughts?

mathiasvr:
Sounds interesting, I think it could be beneficial to have all the media player logic in a separate module. Since there are already a bunch of media player libraries with various fallbacks, I wonder if any of them already does what we need.

Regarding codec/subtitle support what are your thoughts on native players such as WebChimera.js that doesn't depend on browser support of the different formats such as H.265?

feross:
> Regarding codec/subtitle support what are your thoughts on native players such as WebChimera.js that doesn't depend on browser support of the different formats such as H.265?

I think we should definitely look into our options. Right now, the number one error experienced by users is lack of codec support. That's the last big thing blocking v1.0.0, in my opinion.

> I wonder how often people actually uses mkv's with embedded subtitles that are also playable by WebTorrent Desktop?

I think it's probably pretty often, but we can't say for sure without data, of course. I think this PR is important anyway, since once our codec support improves, we'll still want to support embedded mkv subtitles.

src/renderer/controllers/subtitles-controller.js
```diff
@@ -137,3 +155,49 @@ function relabelSubtitles (subtitles) {
     track.label = counts[lang] > 1 ? (lang + ' ' + counts[lang]) : lang
   })
 }
+
+function detectVTTLanguage (buffer) {
+  var LanguageDetect = require('languagedetect')
+
+  // Detect what language the subtitles are in
+  var vttContents = buffer.toString().replace(/(.*-->.*)/g, '') // remove numbers?
```
feross: Re: your comment, I think this does remove numbers since it's removing everything on the line with the --> on it.

src/renderer/controllers/subtitles-controller.js
```diff
@@ -137,3 +155,49 @@ function relabelSubtitles (subtitles) {
...
+  var langDetected = (new LanguageDetect()).detect(vttContents, 2)
+  langDetected = langDetected.length ? langDetected[0][0] : 'subtitle'
+  langDetected = langDetected.slice(0, 1).toUpperCase() + langDetected.slice(1)
+
+  return langDetected
+}
+
+function loadEmbeddedSubtitle (trackEntry) {
+  // convert to .vtt format
```
feross: Is this the `.ass` format? You could make a module for this like the `srt-to-vtt` module we already use. `ass-to-vtt`?

src/renderer/controllers/subtitles-controller.js
```diff
@@ -137,3 +155,49 @@ function relabelSubtitles (subtitles) {
...
+  }
+
+  var buf = new Buffer(vtt)
+  var langDetected = detectVTTLanguage(buf)
+
+  var track = {
+    buffer: 'data:text/vtt;base64,' + buf.toString('base64'),
+    language: langDetected,
+    label: langDetected,
+    filePath: null
```
feross: The `filePath` is used to persist the subtitle that is selected. See `player-page.js` for where it's saved. And in `playback-controller.js` it's read out as `fileSummary.selectedSubtitle`.

We should change this so it works for subtitles embedded within mkv files too.

feross:
This PR is great, let's get it merged!

benwurth:
I'm really interested in having this feature in webtorrent. Is there anything I can do to help move it along?

MauriceMahan:
Hey, how is this coming along? I could see this being a really useful feature since embedded subtitles via MKV are quite popular.

SecT0uch:
Looking forward this feature..

au5ton:
@mathiasvr

mathiasvr:
Hi, I know this PR is very old by now and seemingly abandoned.

The problem is that a proper way to do this requires a refactoring of the current subtitle implementation, and support for loading subtitles during download may impact the download priority of some files especially when seeking.

Anyway, I had hoped for chrome to add support for more video codecs, because as it is now I believe using (or integrating) an external player is a better solution.

I am curious to know if there are some commonly used mkv videos with subtitles that uses a codec playable by WebTorrent Desktop.

If that is the case, I might attempt to finish this ancient thing 😅

DiegoRBaquero:
Is  this something you'd like to continue?

ThaUnknown:
> Is this something you'd like to continue?

I'll be honest, there's an immense amount of work to do if you want to support subtitles with seeking, you can see a bunch of examples how I did it in some of my libraries like https://github.com/ThaUnknown/webtorrent-player/blob/main/index.js or https://github.com/ThaUnknown/pwa-haven/blob/main/video-player/src/modules/subtitles.js but even this isn't what I'd consider good, because of 1 format, SSA which has rendering as complicated as browser CSS, rendering that with decent performance real time is REALLY hard, and a lot of players like VLC/MPC-HC struggle with it, by far the best lib for it is libass, but it still requires immense amt of work to get it working cleanly.

mathiasvr:
@DiegoRBaquero Ah, no not necessarily so I will just close this. Iirc some general restructuring is necessary because subtitles are tightly linked with files atm.

PR #860, opened 2016-08-31
Changed seek forward/backward shortcuts to just left/right arrows

avatsaev:
because ain't nobody got time for those crazy key combinations


avatsaev:
Well all web players use this shortcut (simple arrow with no combinations) and popcorn time did this too.

PR #863, opened 2016-08-31
Material UI: consistent style

dcposch:
### Made some things consistent
* File names
* JSX

### Added new lint rules
* Normal spacing in JSX tags: `<tag />` and `<tag>`, not `<tag/>` or `<tag >`
* No lines over 100 characters
* No vars, only const and let

I didn't want to mess with our standard setup or write eslint rules, so I added something more basic that goes line-by-line enforcing a few banned regex patterns.

@feross


dcposch:
CC @grunjol @mathiasvr

mathiasvr:
LGTM

feross:
> @dcposch Normal spacing in JSX tags: <tag /> and <tag>, not <tag/> or <tag >

The second example is already enforced by `standard`. Try it yourself:

```
$ echo '<t/>' | standard -
standard: Use JavaScript Standard Style (http://standardjs.com)
standard: Run `standard --fix` to automatically fix some problems.
  <text>:1:3: A space is required before closing bracket
```

We should enforce the first one too.

> @dcposch No lines over 100 characters

Sure, I'm fine with this. This probably won't ever be part of `standard` though.

> No vars, only const and let

This will probably become part of `standard` in a future version, then we can remove it from here.

feross:
Opened an issue for the JSX spacing issue: https://github.com/feross/eslint-config-standard-jsx/issues/14

PR #864, opened 2016-08-31
Fix Create Torrent file list

dcposch:
Trivial bug. When there are too many files in the Create Torrent screen, we currently say "+ -80 more" instead of "+ 80 more"


dcposch:
@mathiasvr @feross

mathiasvr:
LGTM

PR #869, opened 2016-09-01
Ensure that torrent file metadata is present

### LINKED ISSUE #854: Uncaught TypeError: path must be a string or Buffer
0.12.0 (0.96.3)
OS X 10.10.5
**What did you do?**
Closed the application with a torrent still loading.
When I turn on the application everything is unchanged.
**What actually happened?**
The window is completely blank and the console shows:
```
Uncaught TypeError: path must be a string or Buffer
```
And I can't add or remove torrents or interact with them

mathiasvr:
Fixes #854

When starting WebTorrent 'new' torrents usually don't have any `.files` yet, causing an error when calling fs.stat with a null parameter.

### DIFF
--- a/src/renderer/controllers/torrent-list-controller.js
+++ b/src/renderer/controllers/torrent-list-controller.js
@@ -84,8 +84,13 @@ module.exports = class TorrentListController {
       return start()
     }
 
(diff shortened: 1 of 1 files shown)


dcposch:
LGTM

PR #870, opened 2016-09-01
Early telemetry

mathiasvr:
To follow up on #854.

Telemetry was not initialized early enough to catch this error. I moved it from `delayedInit` and didn't experience a noticeable change in startup time.

Another problem was that the stack trace was `undefined` because `e` was used instead of `e.error`, so I swapped the if statements.

Also it might be a good idea to `console.log` the uncaught errors, instead of only logging them for telemetry?

@dcposch


dcposch:
> Also it might be a good idea to console.log the uncaught errors, instead of only logging them for telemetry?

I agree

LGTM! Thanks for doing this

Issue #839, opened 2016-08-27
Start playing "first" video in folder


nhuber:
I.e. Be smart about finding S01E01 and stream first in a folder ... Last night I DL'd a series and WT accidentally started me with S1E6 and I didn't realize til 40 mins in :( cc @feross

soapyslick:
Would be a helpful feature to let the user select which file they want to play from the torrent, had the same issue earlier today.

bertofer:
@soapyslick You can select the file you want to reproduce, by clicking on the torrent and expanding it, then clicking on the play icon of the file you want to reproduce.

Thing is, which file is reproduced when clicking on the 'play' button of the entire torrent itself. As I said, I think should be the first in the folder.

soapyslick:
@bertofer Oh sweet, didn't notice that. I've been using single torrents all this time, will start using grouped ones, thanks!

dcposch:
Makes sense. I reopened a closely related feature request, #123.

@soapyslick I agree the torrent list view should be more clear. Right now a lot of people aren't seeing:
1. That they can expand and collapse torrents
2. That they can turn torrents on and off (eg to stop downloading/seeding) via the blue button

We'll try to improve the design there soon. See #652

mathiasvr:
@dcposch any thoughts about this PR #751?

PR #642, opened 2016-06-15
[WIP] Playlists

Goldob:
Add playlists, as proposed in #123.

image

Features:
-
- All torrents are now treated as playlists (for one-file torrents, that is a one-file playlist)
- Playback initially starts at the first file (so `defaultPlayFileIndex` has to go), then we store index of the most recently played file, so that we can pick up where we left off later on
- Control playback through on-screen controls, menu or dedicated media keys
- _Shuffle_ and _repeat_ options

Known issues:
-
- Error thrown when returning from preferences view to the player. See #653

```
rendering error: Cannot read property 'startPiece' of undefined
	TypeError: Cannot read property 'startPiece' of undefined
    at renderLoadingBar (D:\Workspace\webtorrent-desktop\renderer\views\player.js:605:24)
    at renderPlayerControls (D:\Workspace\webtorrent-desktop\renderer\views\player.js:347:9)
    at Object.Player (D:\Workspace\webtorrent-desktop\renderer\views\player.js:22:9)
    at getView (D:\Workspace\webtorrent-desktop\renderer\views\app.js:81:20)
    at App (D:\Workspace\webtorrent-desktop\renderer\views\app.js:44:30)
    at render (file:///D:/Workspace/webtorrent-desktop/renderer/main.js:122:12)
    at redraw (D:\Workspace\webtorrent-desktop\node_modules\main-loop\index.js:65:23)
    at D:\Workspace\webtorrent-desktop\node_modules\main-loop\node_modules\raf\index.js:72:10
```

Issue #123, opened 2016-03-10
Turn torrent into 'playlist': automatically start playing next item


iadj:
There should be a playlist & shuffle option in `webtorrent-app` so you can automatically play the next videos in the list after the current video has finished playing. Also the option to start downloads while items are playing, and a previous/next option.

Goldob:
I've started work on the feature and will make a pull request once I have some results. The current idea is to have separate tabs for torrents and playlists in the home screen (also, all playlists are to be stored in `state.saved.playlists`).

The player would need `playlistId` and `trackId` for reference.

Files could be added individually or in mass, by torrent (e.g. Right click > Add all to playlist).

Any further thoughts/suggestions would be much appreciated.

dcposch:
dunno about shuffle, but i agree we should at least have the ability to play thru all the files in a single torrent without stopping.

lots of audiobook torrents are a collection of sound files, one per chapter. currently, webtorrent desktop doesn't handle them well: you have to click on each chapter one at a time

the reason i'm reluctant to add shuffle support or autoplay across multiple torrents, is that i don't want to go down the road of becoming a media library. once we do that, people will want a way to reorder the torrent list, group torrents with labels or folders or playlists, etc.

i want to keep the app minimalist

dcposch:
i like the idea in #642 : autoplay, but only within torrents, not between them

dcposch:
@feross thoughts?

feross:
@dcposch Agree with what you said. Autoplaying the next file in the torrent is sufficient.

dcposch:
So to recap:
* Play in order only, no shuffle
* Play within a torrent only, not across torrents
* Play files in the order they appear in the torrent

Goldob:
I can get back to work on #642, now that I'm back. There will probably be quite a lot refractoring, but still - most of the work is done. I will let you know about the progress soon.

dcposch:
Done in #871, released in v0.14

Thanks @Goldob !

Issue #599, opened 2016-05-27
Sound doesn't stop on returning to menu


furstenheim:
**What version of WebTorrent Desktop?** (See the 'About WebTorrent' menu)
Version 0.6.1 (0.92.0)
**What operating system and version?**
Linux
**What did you do?**
As I was playing a video I clicked on the back button to return to the main menu.
**What did you expect to happen?**
Return to menu and stop the sound.
**What actually happened?**
, it returned but the sound didn't stop. It's happened couple of times but I haven't been able to reproduce consciously. There are three erros on the console:

> undefined:1 Uncaught (in promise) DOMException: The play() request was interrupted by a call to pause().
> 0:1 GET http://localhost:36347/0 net::ERR_CONTENT_LENGTH_MISMATCH
> 0:1 GET http://localhost:36347/0 net::ERR_CONNECTION_REFUSED

feross:
What Linux distro and version are you using?

feross:
I can NOT reproduce this on Ubuntu 64-bit 16.04.

furstenheim:
Ubuntu 64bit 14.04. It is something that almost never happens. It happened once during normal execution and then 4 more times trying to reproduce it. 
What I did in order to reproduce was repeatedly playing either of two videos (Sintel and Big Buck Bunny) and then going back to the menu. As I said most of the time everything work out fine. But these five times the sound kept on going and the three same lines were logged to the console.

feross:
My guess is that is either related to this Chrome issue: https://bugs.chromium.org/p/chromium/issues/detail?id=593273

Or, it could be caused by an earlier exception. Let's close this for now, and re-open if we see it after the v0.7.0 release.

feross:
I just witnessed this issue, so it's real. I was about to open a new issue, but remembered that this one already exists.

**What version of WebTorrent Desktop?** (See the 'About WebTorrent' menu)

v0.7.2

**What operating system and version?**

OS X 10.11.5

**What did you do?**

I watched @jhiesey click the AirPlay icon repeatedly as fast as he could.

**What did you expect to happen?**

The app should remain in a reasonable state.

**What actually happened?**

Multiple copies of the audio play at the same time, even though there are no video tags in the DOM. This makes no sense of course, but I did confirm that `document.querySelectorAll('video')` returned no elements.

[...]

PR #871, opened 2016-09-01
Add playlists feature

Goldob:
Open multi-file torrents as playlists. Basically #642 after rebase.
Solves #123 and #839.

image


src/renderer/lib/cast.js
```diff
@@ -96,7 +96,7 @@ function chromecastPlayer () {
 
   function open () {
     var torrentSummary = state.saved.torrents.find((x) => x.infoHash === state.playing.infoHash)
-    ret.device.play(state.server.networkURL, {
+    ret.device.play(state.server.networkURL + '/' + state.playing.fileIndex, {
```
dcposch: I'd put this into its own function to reduce code duplication

src/renderer/pages/PlayerPage.js
```diff
@@ -109,7 +109,7 @@ function renderMedia (state) {
   var MediaTagName = state.playing.type
   var mediaTag = (
     <MediaTagName
-      src={state.server.localURL}
+      src={state.server.localURL + '/' + state.playing.fileIndex}
```
dcposch: How was this working before, given that you can click Play on an individual file in a torrent? (Did `localURL` previously contain the file index, but it doesn't anymore?)

dcposch:
@Goldob thanks for doing this! Looks great

I think it can be simplified a bit
* There's a bit of code duplication. Nothing serious
* I don't know if we need `state.playlist`. The state object is already pretty enormous, and `state.playing` already contains the current infohash + file index we're playing. I think that's enough state to implement skip-forward, skip-back, and automatically playing the next file after one finishes.
* I don't know if we want to add Repeat and Shuffle.

Here's why:
* Autoplay is great for listening to an album or audiobook, but Shuffle inside a single torrent isn't that useful, right? Usually people want to shuffle across a whole collection.
* We want to keep the app minimalist
* We want users (and reviewers) to think of us as a great torrent app, not as an OK media library. Adding a Shuffle and Repeat button starts to feel like ITunes / Media Player / etc.

The new UI (menu buttons plus player icons for skipping forward and back) looks great to me.

dcposch:
Tempted to merge this now and make a follow up PR to avoid conflicts with #863 

CC @feross @mathiasvr

src/renderer/lib/playlist.js
```diff
@@ -0,0 +1,161 @@
...
+  this._infoHash = torrentSummary.infoHash
+  this._position = 0
+  this._tracks = extractTracks(torrentSummary)
+  this._order = range(0, this._tracks.length)
+
+  this._repeat = false
+  this._shuffled = false
+}
+
+// =============================================================================
```
dcposch: I wouldn't add ascii banners like this. We don't have them anywhere else in the code

mathiasvr:
Just tested it, and I get this error when it auto skips to next track:
```
undefined:1 Uncaught (in promise) DOMException: The play() request was interrupted by a call to pause().
```
This is probably a Chrome issue as discussed in #599. Not sure if it's serious though.

src/renderer/lib/playlist.js
```diff
@@ -0,0 +1,161 @@
...
+// Utility fuctions
+// =============================================================================
+
+function extractTracks (torrentSummary) {
+  return torrentSummary.files.map((file, index) => ({ file, index }))
+    .filter((object) => TorrentPlayer.isPlayable(object.file))
+    .sort(function (a, b) {
+      if (a.file.name < b.file.name) return -1
+      if (b.file.name < a.file.name) return 1
```
mathiasvr: Do we want to sort the files by name? I think preserving the original order makes sense.

Goldob: Files in [torrent list](https://github.com/feross/webtorrent-desktop/blob/d806fd5/src/renderer/pages/TorrentListPage.js#L276) are displayed alpabetically, so I think it's reasonable to keep that order. Also, when we stick to file indices, the sequence appears totally random.

src/renderer/pages/PlayerPage.js
Goldob: Exactly like you say. Refer to [this line](https://github.com/feross/webtorrent-desktop/pull/871/files#diff-1f9e4aabf46e7be3ad02d963e0a2fc07L317).

src/renderer/lib/playlist.js
mathiasvr: Do you have an example? All torrents I've checked is ordered reasonably.

Goldob:
Thank you very much for the feedback! I'll introduce the suggested changes.

The `shuffle` and `repeat` options are here only because I implemented them in the original branch (before rebase). I'll revert the commit and you'll be able to just skip it when merging.

 @mathiasvr, similar error appears sometimes if you change tracks too fast. It is caused by [the Chrome issue](https://bugs.chromium.org/p/chromium/issues/detail?id=593273), just like you said. In case of autoplaying my guess is that it tries to replay the file after reaching the end, but before it is able to do that, new media object replaces it.

src/renderer/lib/playlist.js
Goldob: I just checked and you are right. I had problem with one particular torrent (although I won't post it here, as we don't do piracy) and that's why I assumed it would be the case everywhere.

That said, getting rid of file sorting would let for yet another simplification of the code. On the other hand, I still think it could be safer to stick with it, as most torrents follow _S01S01_, _S01S02_, _S01E03_ or just _01_, _02_, _03_ schema. It's not my decision to make, however  (@dcposch?).

Anyway, for the sake of consistency, I think the alphabetical order should go either both in playlists and torrent list or none of them.

dcposch: @Goldob @mathiasvr I agree, I think keeping the original order is probably a good idea.

@feross what was the reason for sorting torrents alphabetically?

mathiasvr: I guess sort by name is okay.
I agree that it should be consistent, maybe it would be easier to sort the files initially and stick with that order everywhere else?

dcposch:
@Goldob looks great. I have a few nitpicks still but they're really small and easy to fix later. 🚢 🚢 🚢

Goldob:
Awesome to see that merged. One remark from me, before actually releasing this it would be nice to add a line to `renderer/lib/migrations.js` to remove `torrentSummary.defaultPlayFileIndex`, because we don't use it anymore.

dcposch:
@Goldob fixed https://github.com/feross/webtorrent-desktop/pull/880/commits/b8effffa96d45132ccc01509f4a6cda91ea6b16a

PR #877, opened 2016-09-02
Restore playback state when reopening player

Goldob:
This is a follow up for PR #871.

The intended behavior was that when opening player, playing some files and then navigating backward/forward in the app, the playback state is restored. In one particular case, when we open a specific file instead of a torrent and then do the above, the playback starts again at the same file it started at instead.

This PR fixes that.

### DIFF
--- a/src/renderer/controllers/playback-controller.js
+++ b/src/renderer/controllers/playback-controller.js
@@ -34,12 +34,20 @@ module.exports = class PlaybackController {
         else this.play()
       })
     } else {
+      let initialized = false
       state.location.go({
         url: 'player',
         setup: (cb) => {
-          this.play()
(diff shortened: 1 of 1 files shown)


src/renderer/controllers/playback-controller.js
```diff
@@ -34,12 +34,20 @@ module.exports = class PlaybackController {
         else this.play()
       })
     } else {
+      let initialized = false
```
mathiasvr: I had an idea to set `torrentSummary.mostRecentFileIndex` instead of using the `initialized` variable and remove the index param from `openPlayer`, but I don't think it's the best solution after all.

mathiasvr:
LGTM and tested 👍

@dcposch Can we include this in #885?

dcposch:
@mathias yeah ill rebase #885

PR #882, opened 2016-09-03
Rename App.js -> app.js

### LINKED ISSUE #875: Uncaught Error: Cannot find module './pages/app'
Version 0.13.1 (0.96.4) - b93f41f564d2838fa1354d94e55f99aede6cc068

Arch Linux x64

**What did you do?**
Seeded some low-peer torrents and restarted app.

See the list of torrents.

**What actually happened?**
screenshot_2016-09-02_01-36-38
screenshot_2016-09-02_01-36-39

I was experiencing #854 so I updated to the latest commit in hopes to fix the issue, however I am presented with the same issue but with a new exception.

Here is my [config.json](https://gist.github.com/anoadragon453/1ac06d27ecaddb3d9ed9c0d4168336c6).

Thanks!

mathiasvr:
Fixes #875.

This issue was introduced in #863 because of renaming a `require` path, but not the file.
This is not a problem on Mac (and Windows?) since the filesystem is case-insensitive, and probably why it went undetected.


dcposch:
LGTM. I thought we lowercased all the file names earlier... guess not. My bad.

PR #886, opened 2016-09-04
Fix 0.14.0

dcposch:
@feross @mathiasvr 

Torrent progress bar was actually added in 0.14, not 0.13. I've updated the GH release notes


noamokman:
:+1:

dcposch:
So it turns out the `Heading.js` bug was real, I just couldn't reproduce it because I am on a Mac and Macs have a case insensitive file system.

Even with `git config core.ignorecase`, it is still broken.

I went down the rabbit hole of learning how and why Macs behave this way.

http://apple.stackexchange.com/questions/71357/how-to-check-if-my-hd-is-case-sensitive-or-not

> You can use terminal to read the File System Personality:
> 
> If the file system is case sensitive, you will see Case-sensitive Journaled HFS in the first pasted line and Mac OS Extended (Case-sensitive, Journaled) in the third.

This is capital R Retarded.

dcposch:
Fixed. I am not making `0.14.1

feross:
Sorry about this. This was my fault. I was naming all new React component files with capital letters to match the name of the class that each file exports, since this is what the Material UI library does:

```js
const RaisedButton = require('material-ui/RaisedButton').default
```

Because I didn't want this kind of inconsistency:

```js
const RaisedButton = require('material-ui/RaisedButton').default
const Heading = require('../components/heading')
```

But I should have been more careful about the testing the casing across OSs -- I've actually run into this issue with my Study Notes site before.

PR #893, opened 2016-09-04
Disable sound option in settings

avatsaev:


src/renderer/lib/prefs.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
```
dcposch: 99% sure we don't need this file

The state object already contains prefs

dcposch:
Otherwise, looks reasonable to me

@feross, what do you think? Maybe we can just make the sound effects a bit more subtle. Adding a checkbox to Preferences just for disabling sound effects doesn't feel very minimalist

avatsaev:
To be honest with you, from UX stand point, 99% of the sound effects in the app are not necessary, usually sounds are used to make user pay attention to an important event, like a notification for example, but start up sound? Play sound? They are more of an annoyance than convinience

feross:
-1 from me. Let's just make the right decision with regards to sounds. Adding a preference is a cop-out IMO.

avatsaev:
The right decision would be to remove them all together

feross:
You're the first person to express this opinion, so I disagree that it's clearly the right decision. Thanks for sharing your thoughts nonetheless.

avatsaev:
https://github.com/feross/webtorrent-desktop/issues/858

Grogs:
+1 for no sounds.

feross:
I just sent a PR to make the UI sounds subtler. See https://github.com/feross/webtorrent-desktop/pull/945 I hope you give it a try by cloning `master` or waiting until the next release comes out. Would be great to get your feedback.

I'm going to close this PR now.

PR #894, opened 2016-09-04
Fix some of the common errors

dcposch:
From telemetry.

@feross @mathiasvr

### DIFF
--- a/src/renderer/lib/cast.js
+++ b/src/renderer/lib/cast.js
@@ -126,13 +126,7 @@ function chromecastPlayer () {
   }
 
   function status () {
-    ret.device.status(function (err, status) {
-      if (err) return console.log('error getting %s status: %o', state.playing.location, err)
-      state.playing.isPaused = status.playerState === 'PAUSED'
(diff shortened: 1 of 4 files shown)


mathiasvr:
LGTM

src/renderer/lib/cast.js
```diff
@@ -328,6 +316,18 @@ function dlnaPlayer (player) {
   }
 }
 
+function handleStatus (err, status) {
+  if (err || !status) {
+    return console.log('error getting %s status: %o',
```
mathiasvr: Is there a reason for logging instead of throwing here?

dcposch: * We were logging before
* Looks like it's common for at least one of the cast libraries to call back with a null error **and** a null status object. I want to ignore when that happens
* Getting an error is also common. For example, if someone's using Chromecast, and then turns off the TV (powering off the Chromecast), then I think WebTorrent will get an error on the next status callback. That's expected.

PR #898, opened 2016-09-05
Move error definitions to `errors.js`

Goldob:
* Moved most of the error definitions to dedicated `errors.js` file, which was hanging unused until now
* Created separate classes for different types of errors to increase verbosity and get rid of vague errors
* Fixed a typo in error name inside `playback-controller.js` possibly resulting in a `TypeError`

### DIFF
--- a/src/renderer/lib/errors.js
+++ b/src/renderer/lib/errors.js
@@ -1,15 +1,54 @@
+const ExtendableError = require('es6-error')
+
+/* Generic errors */
+
+class CastingError extends ExtendableError {}
+class PlaybackError extends ExtendableError {}
+class SoundError extends ExtendableError {}
+class TorrentError extends ExtendableError {}
+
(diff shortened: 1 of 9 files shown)


src/renderer/lib/errors.js
```diff
@@ -1,15 +1,54 @@
...
 
-UnplayableTorrentError.prototype = Error
-UnplayableFileError.prototype = Error
+class InvalidTorrentError extends TorrentError {}
+
+/* Miscellaneous */
+
+class IllegalArgumentError extends ExtendableError {}
+
+module.exports = {
```
Goldob: Note on consistency. The `exports` statement can't be on the top like in the other files due to class declarations being non-hosted in ES6.

dcposch:
LGTM. @feross  ?

feross:
LGTM

Issue #765, opened 2016-08-06
Add basic integration test


feross:
The Electron team created Spectron to test Electron apps:

http://electron.atom.io/spectron/
https://github.com/electron/spectron

It looks like @dcposch already started integrating this on a branch: https://github.com/feross/webtorrent-desktop/tree/dc/test

I think we can keep things simple and just add a single integration test to ensure the app starts up without crashing.

dcposch:
Done. We still have lots more integration tests to add, but we have a nice system now. We have good tests for the torrent list.

Next up: testing audio and video playback, casting, adding existing torrents, creating new torrents, and the prefs page.

feross:
Thanks for tackling this. I think this will really us to not ship broken software.

PR #910, opened 2016-09-07
Integration testing

dcposch:
**I've added a system to run WebTorrent Desktop, click around, wait for the UI to update, take screenshots, and make sure those screenshots match what the UI should look like.**

It's fun to watch the tests run...

This is the first step to fixing #765

#### Linting

`npm test`

#### Integration testing

`npm run integration-test`

#### Creating an integration test

See `test/index.js` for two examples.

If you add a new `setup.screenshotCreateOrCompare`, the first time you run tests afterward, missing screenshots will be created under `test/screenshots`. After that, existing screenshots are compared.

### I've added two basic tests to start out

- [x] Run the app. Make sure it starts and that the default torrents look right
- [x] Run the app with the Downloads folder missing. Check for the error message, then click on the Preferences link in the error message.

We need many more tests.

Time to automate the testing we used to do by hand: https://github.com/feross/webtorrent-desktop/pull/865#issuecomment-243949030


dcposch:
@feross @mathiasvr @Flet @noamokman

Goldob:
Wouldn't it be better to check contents with WebdriverIO instead of taking screenshots? These tests will fail each time a small change in UI appears. At least maybe the 'faulty' screenshot should be saved? This way we could use it for feedback or even replace the 'right' one with it if the change is intentional (this workflow still doesn't seem right thou).

test/index.js
```diff
@@ -0,0 +1,43 @@
...
+    .then(() => setup.wait())
+    .then(() => setup.screenshotCreateOrCompare(app, t, 'torrent-list-basic'))
+    .then(() => setup.endTest(app, t),
+          (err) => setup.endTest(app, t, err || 'error'))
+})
+
+test('show download path missing', function (t) {
+  fs.removeSync(DOWNLOAD_DIR)
+
+  t.timeoutAfter(10e3)
```
Goldob: Maybe it's just me, but this not quite enough for app initialization and all the tests.

dcposch:
@Goldob I don't think so.
* Ad-hoc checking makes the tests a lot more work to write
* Even diffing the whole HTML is not as thorough as screenshot diffing. For example, it wouldn't catch an bug where hitting ESC from a video doesn't correctly restore window size.
* Chrome's own integration tests use screenshot diffing iirc
* Small UI changes will break a few tests, but the fix is as easy as deleting the offending screenshots and running the tests, which will recreate them with the new look.
* The resulting Github PR will then show, pixel by pixel, the exact UI changes that were made! This is pretty sweet IMO. https://github.com/blog/817-behold-image-view-modes

Goldob:
@dcposch  Okay, now I'm convinced. Thanks for the explanation.

It makes a lot more sense with this GitHub feature.

test/index.js
```diff
@@ -0,0 +1,43 @@
...
+const DOWNLOAD_DIR = path.join(setup.TEST_DATA_DIR, 'Downloads')
+fs.mkdirpSync(DOWNLOAD_DIR)
+
+test.onFinish(function () {
+  console.log('Removing test dir...')
+  fs.removeSync(setup.TEST_DATA_DIR)
+})
+
+test('app runs', function (t) {
+  t.timeoutAfter(10e3)
```
Flet: Huh, today I learned scientific notation is supported :)
10e3 = 10000

test/setup.js
```diff
@@ -0,0 +1,74 @@
...
+    path: path.join(__dirname, '..', 'node_modules', '.bin',
+      'electron' + (process.platform === 'win32' ? '.cmd' : '')),
+    args: [path.join(__dirname, '..')],
+    env: {NODE_ENV: 'test'}
+  })
+}
+
+// Starts the app, waits for it to load, returns a promise
+function waitForLoad (app, t) {
+  return app.start().then(function () {
```
Flet: Should the saved preferences be reset at every load for consistency?

Flet:
Very cool @dcposch! 👍

src/config.js
```diff
@@ -89,22 +100,31 @@ function getConfigPath () {
...
+    return getPath('downloads')
   }
+}
 
-  const electron = require('electron')
+function getPath (key) {
+  if (process.type === 'renderer') {
+    return electron.remote.app.getPath(key)
+  } else {
+    electron.app.getPath(key)
```
noamokman: @dcposch missing return here

test/index.js
dcposch: Yeah. 10e3 milliseconds = 10 seconds

test/setup.js
dcposch: I added `wipeTestDataDir` to delete everything. Not just saved prefs but all of config.json + torrents + torrent data. Clean slate.

src/config.js
dcposch: fixed

dcposch:
> Or at least maybe the 'faulty' screenshot should be saved? 
@Goldob yeah, good call. Done

feross:
LGTM!

PR #913, opened 2016-09-09
Integration tests

dcposch:
@feross @mathiasvr


dcposch:
```
$ time npm run integration-test

> webtorrent-desktop@0.14.0 integration-test /Users/dc/code/webtorrent-desktop
> npm run build && node ./test

> webtorrent-desktop@0.14.0 build /Users/dc/code/webtorrent-desktop
> babel --quiet src --out-dir build

TAP version 13
# app runs
ok 1 html title
ok 2 screenshot comparison app-basic
# torrent-list: start, stop, and delete torrents
ok 3 html title
ok 4 screenshot comparison torrent-list-hover
ok 5 screenshot comparison torrent-list-start-download
ok 6 screenshot comparison torrent-list-hover-download
ok 7 screenshot comparison torrent-list-delete-prompt
ok 8 screenshot comparison torrent-list-hover
ok 9 screenshot comparison torrent-list-delete-prompt
ok 10 screenshot comparison torrent-list-deleted
# torrent-list: expand torrent, unselect file
ok 11 html title
ok 12 screenshot comparison torrent-list-cosmos-hover
ok 13 screenshot comparison torrent-list-cosmos-expand
ok 14 screenshot comparison torrent-list-cosmos-expand-deselect
ok 15 screenshot comparison torrent-list-cosmos-expand-start
ok 16 download folder contents: CosmosLaundromatFirstCycle
ok 17 screenshot comparison torrent-list-cosmos-delete-data
ok 18 screenshot comparison torrent-list-cosmos-deleted
ok 19 download folder missing: CosmosLaundromatFirstCycle
# add-torrent
ok 20 html title
ok 21 screenshot comparison add-torrent-0-percent
ok 22 screenshot comparison add-torrent-100-percent
# create-torrent
ok 23 html title
ok 24 screenshot comparison create-torrent-simple
ok 25 screenshot comparison create-torrent-advanced
ok 26 screenshot comparison create-torrent-100-percent
ok 27 torrent contents: /Users/dc/code/webtorrent-desktop/test/tempTestData/Desktop/saved.torrent
# video-streaming
ok 28 html title
ok 29 screenshot comparison play-torrent-bbb
ok 30 screenshot comparison play-torrent-return
# audio-streaming
ok 31 html title
ok 32 screenshot comparison play-torrent-wired
ok 33 screenshot comparison play-torrent-wired-2
ok 34 screenshot comparison play-torrent-wired-3
ok 35 screenshot comparison play-torrent-wired-fullscreen
ok 36 screenshot comparison play-torrent-wired-4
ok 37 screenshot comparison play-torrent-wired-list
ok 38 screenshot comparison play-torrent-wired-5

1..38
# tests 38
# pass  38

# ok

real	2m40.345s
user	1m49.353s
sys	0m32.132s
```

src/renderer/pages/player-page.js
```diff
@@ -585,6 +586,8 @@ function renderPlayerControls (state) {
 // Renders the loading bar. Shows which parts of the torrent are loaded, which
 // can be 'spongey' / non-contiguous
 function renderLoadingBar (state) {
+  if (config.IS_TEST) return // Don't integration test the loading bar. Screenshots won't match.
```
feross: So, is this now the preferred approach vs. creating transparent regions in the test images?

src/renderer/webtorrent.js
```diff
@@ -54,11 +54,16 @@ const VERSION_STR = VERSION.match(/([0-9]+)/g)
  */
 const VERSION_PREFIX = '-WD' + VERSION_STR + '-'
 
+/**
+ * Generate an ephemeral peer ID each time.
+ * TODO: once there are around 2^24 = ~8 million WebTorrent Desktops online at the same time,
+ *       ID collisions will start happening. Birthday paradox. Can we use more than six bytes?
```
feross: The peer id must be 20 bytes. By convention, 8 of those bytes are used to identify the client and version (for example, `-WD0100-` for WebTorrent Desktop v1.0). That leaves 12 bytes for the rest of the id.

There's no requirement that the ID use printable characters, though that's really nice to have for debugging. That's why we're doing 6 bytes, but printing them as hex (which uses 12 bytes).

We could add entropy and still keep this printable by changing it to base64:

```js
crypto.randomBytes(9).toString('base64')
```

Thoughts?

feross: If we change this here, let's also change it in `webtorrent` too.

feross: Btw, peer ID collisions aren't that bad... A malicious peer can already join a swarm, copy the peer id of an existing peer and start connecting to people. Worst thing that can happen is a mild DOS if peers refuse to connect to two peers with the same ID. But the attacker would actually have to be sending real torrent data to peers in order to remain connected and continue the DOS, so it's a pretty weak DOS, IMO.

test/setup.js
```diff
@@ -91,21 +90,81 @@ function screenshotCreateOrCompare (app, t, name) {
   })
 }
 
+// Compares two PNGs, ignoring any transparent regions in bufExpected.
+// Returns true if they match.
+function compareIgnoringTransparency (bufActual, bufExpected) {
```
feross: If we're doing this comparison while ignoring transparent regions, why is there an if-statement to not render the progress bar if `config.IS_TEST` is true in another file?

test/setup.js
```diff
@@ -91,21 +90,81 @@ function screenshotCreateOrCompare (app, t, name) {
...
 // Resets the test directory, containing config.json, torrents, downloads, etc
-function wipeTestDataDir () {
-  fs.removeSync(TEST_DATA_DIR)
-  fs.mkdirpSync(TEST_DOWNLOAD_DIR) // Downloads/ is inside of TEST_DATA_DIR
+function resetTestDataDir () {
+  fs.removeSync(config.TEST_DIR)
+  // Create TEST_DIR as well as /Downloads and /Desktop
+  fs.mkdirpSync(config.TEST_DIR_DOWNLOAD)
+  fs.mkdirpSync(config.TEST_DIR_DESKTOP)
+}
```
feross: Could this be simplified by just creating a randomly named folder in `/tmp` and letting the OS deal with cleanup?

test/setup.js
```diff
@@ -91,21 +90,81 @@ function screenshotCreateOrCompare (app, t, name) {
...
+function compareTorrentFiles (t, pathActual, pathExpected) {
+  const bufActual = fs.readFileSync(pathActual)
+  const bufExpected = fs.readFileSync(pathExpected)
+  const fieldsActual = extractImportantFields(parseTorrent(bufActual))
+  const fieldsExpected = extractImportantFields(parseTorrent(bufExpected))
+  t.deepEqual(fieldsActual, fieldsExpected, 'torrent contents: ' + pathActual)
+}
+
+function extractImportantFields (parsedTorrent) {
+  const { infoHash, name, announce, urlList, comment } = parsedTorrent
```
feross: `name` is encapsulated within `infoHash`, but doesn't hurt to check it I suppose. The others do not affect the info hash, so good call on checking them.

test/setup.js
```diff
@@ -91,21 +90,81 @@ function screenshotCreateOrCompare (app, t, name) {
...
+  const priv = parsedTorrent.private
```
feross: Add a comment? `// 'private' is a reserved word`

feross:
@dcposch I noticed that the `play-torrent*` images aren't checked into this PR. Can you add them?

src/renderer/webtorrent.js
dcposch: Yeah this is fine. I'll just change the comment to say that peer ID collisions aren't that bad.

test/setup.js
dcposch: I did it that way initially, but I like this better because it means we can use transparency a lot less. That makes it easier to update the screenshots when we change the CSS.

Also, I was first trying to avoid `config.IS_TEST` completely, but then I ended up needing it anyway to mock `cast.js`

test/setup.js
dcposch: We can put it in `/tmp` instead of here but we should still clean it up ourselves. Otherwise, a single run of the integration tests will produce 10+ different app folders, each with their own `config.json`, Chrome cache, torrent and poster files, etc

I think it's fine

test/setup.js
dcposch: Sure

src/renderer/pages/player-page.js
dcposch: IMO we want to use `config.IS_TEST` as little as possible.

In this case, it saves us from having to make transparent screenshots for every shot of the player page. Feels better b/c regular screenshots are easier to update when the UI changes

dcposch:
@feross yeah, added them back. Ran the integration tests a bunch of times in a row to find and fix flaky parts.

src/renderer/webtorrent.js
feross: Just sent a PR to add more entropy since I think it's a good idea

PR #919, opened 2016-09-11
Update mouse moved time on header hover

thepelkus-too:
Previously, moving the mouse into the player window from the sides or
bottom would bring up the HUD, but moving the mouse in from the top
would not. With this commit, moving the mouse in from the top of the
window will also bring up the HUD.

Fixes feross/webtorrent-desktop#241


mathiasvr:
LGTM

dcposch:
LGTM. Thanks @thepelkus-too !

PR #922, opened 2016-09-13
Use raised button for inline button

feross:
Before:

screen shot 2016-09-13 at 8 16 20 am

After:

screen shot 2016-09-13 at 8 16 13 am

### DIFF
--- a/src/renderer/components/show-more.js
+++ b/src/renderer/components/show-more.js
@@ -1,6 +1,6 @@
 const React = require('react')
 
-const FlatButton = require('material-ui/FlatButton').default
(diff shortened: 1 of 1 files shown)


mathiasvr:
LGTM, but I think the `Cancel` and `Create Torrent` should have the same style, either raised or flat.

dcposch:
Needs more vertical space imo

feross:
@mathiasvr I didn't change the `Cancel` button style. That's just an artifact of where my mouse was when I took the screenshot ;)

feross:
@dcposch You're right, let's do in future PR

PR #926, opened 2016-09-16
Electron 1.4.0

feross:
The changes that are relevant to us are:

- Upgrade to Chrome 53.
- Upgrade to Node 6.5.0.
- Add support for titleBarStyle: 'hidden-inset' on OS X 10.9.

The last item should improve our UI on OS X 10.9 since we use 'hidden-inset'. I don't know what it looked like before.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "deep-equal": "^1.0.1",
     "dlnacasts": "^0.1.0",
     "drag-drop": "^2.12.1",
-    "electron": "1.3.5",
+    "electron": "1.4.0",
     "es6-error": "^3.0.1",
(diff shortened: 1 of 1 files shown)


feross:
Thanks @Flet!

dcposch:
I think this broke the console in a weird way, but only when running from Spectron.

1. Run the integration tests, but add a `wait(100e3)` somewhere to give yourself time
2. Open DevTools. Go to the Console
3. **Notice the Console is empty.** If you type anything and hit enter, it just disappears. The console is broken.
4. `npm install electron@1.3.5`
5. Repeat steps 1-3. The console works this time.

???

PR #928, opened 2016-09-17
Detect system architecture; send in update/telemetry

feross:
Detect the actual operating system CPU architecture. This is different
than `process.arch` which returns the architecture the binary was
compiled for.

This is just good info to have in the telemetry, but we're also sending
it in the update check so that eventually we can upgrade Windows 32-bit
apps to 64-bit, like Slack does.

Context: https://github.com/feross/webtorrent-desktop/issues/873#issuecomment-247722023

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -81,6 +81,8 @@ module.exports = {
   IS_PRODUCTION: IS_PRODUCTION,
   IS_TEST: IS_TEST,
 
+  OS_SYSARCH: is64BitOperatingSystem() ? 'x64' : 'ia32',
+
   POSTER_PATH: path.join(getConfigPath(), 'Posters'),
   ROOT_PATH: path.join(__dirname, '..'),
   STATIC_PATH: path.join(__dirname, '..', 'static'),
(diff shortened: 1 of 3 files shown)


feross:
Can I get a review of this? @dcposch @mathiasvr @ungoldman

src/config.js
```diff
@@ -149,3 +151,32 @@ function isProduction () {
...
+
+/**
+ * Returns the operating system's CPU architecture. This is different than
+ * `process.arch` which returns the architecture the binary was compiled for.
+ *
+ * On Windows, the most reliable way to detect a 64-bit OS from within a 32-bit
+ * app is based on the presence of a WOW64 file: %SystemRoot%\SysNative.
+ *
+ * Background: https://twitter.com/feross/status/776949077208510464
+ */
```
ungoldman: WOW, that's weird

dcposch: Stands for Windows On Windows 64. It's the 32-bit backward compatibility layer

Flet:
Could this be built into https://www.npmjs.com/package/arch ?

feross:
@Flet Not a bad idea. That would actually make the package more consistent. Right now it detects the architecture of the Node.js binary when running in Node.js, but the OS architecture when running in the browser.

If we added this code to `arch`, then it would be consistent and more useful: it would always tell you the OS architecture.

ungoldman:
Makes sense to me!

feross:
@ungoldman @Flet PR sent! https://github.com/feross/arch/pull/2

Issue #873, opened 2016-09-02
Common error: Array buffer allocation failed


dcposch:
See https://github.com/feross/webtorrent/issues/895 for details.

Can we package a 64-bit version in addition to the 32-bit Windows app? 

**This bug only happens on Windows.** It has happened 100s of times in the telemetry, always on Windows. Our Windows packager creates a 32-bit executable, so we only get 4GB of address space, even tho the vast majority of our users are using 64 bit hardware. This is probably part of the problem.

@grunjol @feross

feross:
It would be really nice to just go 64-bit only. Maybe the 32-bit usage will be low enough in a year? Right now, this is the breakdown according to [Steam's stats](http://store.steampowered.com/hwsurvey/):

<img width="366" alt="screen shot 2016-09-14 at 4 48 07 pm" src="https://cloud.githubusercontent.com/assets/121766/18534090/d77471de-7a9c-11e6-9cf9-eac059a1d8de.png">

The most important stats are:

- ~6% of users on Windows 7 (32-bit)
- ~1.5% on Windows 10 (32-bit).

Would be interesting to see what the breakdown of our users is -- more or less than Steam?

The problem with offering two versions is that it's really confusing for users to pick the right one. We can get away with that for Linux users because they're usually pretty technically saavy.

So, if we offer 32-bit as the default, and let users select 64-bit if they want (like we do for Linux, to err on the safe side), then most users will still experience the "Array buffer allocation failed" issue. Not ideal.

For comparison: Atom ships 32-bit only. Slack ships a 32-bit, but detects if the user is on 64-bit and sets a different auto-update URL so they'll get updated to 64-bit eventually (this is too complex IMO).

feross:
Here's an approach I like: Make a 64-bit Windows build. Offer 32-bit vs 64-bit installer based on the platform indicated by the browser. I just played around with this a bit and it looks like it can be pretty reliable:

https://github.com/feross/arch

In the browser, there is no spec that defines where this information lives, so we check all known locations including `navigator.userAgent`, `navigator.platform`, and `navigator.cpuClass` to make a best guess.

If there is no affirmative indication that the architecture is 64-bit, then 32-bit will be assumed. This is perfect for determining what installer executable to offer to desktop app users. If there is ambiguity, then the user will get the 32-bit installer, which will work fine even for a user with a 64-bit OS.

feross:
[...]

PR #931, opened 2016-09-18
Add 64-bit Windows build

feross:
Right now all Windows users are running a 32-bit app, even if their OS
is 64-bit. This means that the app is limited to a 4GB address space,
which may be contributing to the "Array buffer allocation failed" errors
we are seeing in the wild (#873).

If users on Windows 64-bit (90%+ of users) could run a 64-bit app,
they're less likely to run into out-of-memory errors.

Here's the plan to improve things:

1. We release a 64-bit installer, in addition to the 32-bit installer.

2. We auto-detect in the browser when a visitor is on a 32-bit vs.
64-bit OS and try to offer them the right installer. When in doubt, we
give them the 32-bit installer since that's safest.

3. The auto-updater will return the right binaries for the architecture
the user is on. This means that all our existing users who have 64-bit
OSs but are running the 32-bit app will get updated to the 64-bit app
on the next update. Also, 64-bit users who accidentally download the
32-bit installer will also get the 64-bit app on the next update.

---

Other notes:

- We don't generate 32-bit delta files. See reasoning inline.
- The package script now deletes extraneous Squirrel files (i.e.
*.nupkg delta files for older versions of the app) which should make
uploading the right files to GitHub easier. :)

The binary file naming works like this:

- Most users are on 64-bit systems, so they get nice clean file names
that don't specify an architecture (WebTorrentSetup-v1.0.0.exe). The
32-bit build files have the same naming but contain the string "-ia32"
at the end. In a few years, we will be able to just stop producing the
32-bit build files entirely.

- This means that the "WebTorrent-v0.15.0-linux-x64.zip" linux build
file is changing to "WebTorrent-v0.15.0-linux.zip" to match the Windows
naming convention. The .deb installer files must contain to
architecture in order to install correctly, so those do not change.

- Mac is 100% 64-bit, so it does not change.

### DIFF
--- a/bin/package.js
+++ b/bin/package.js
@@ -462,8 +462,8 @@ function buildWin32 (cb) {
 
           // WebTorrent-vX.X.X-full.nupkg -> WebTorrent-vX.X.X-ia32-full.nupkg
           fs.renameSync(
-            path.join(DIST_PATH, `${config.BUILD_NAME}-full.nupkg`),
-            path.join(DIST_PATH, `${config.BUILD_NAME}-ia32-full.nupkg`)
+            path.join(DIST_PATH, `${BUILD_NAME}-full.nupkg`),
+            path.join(DIST_PATH, `${BUILD_NAME}-ia32-full.nupkg`)
           )
 
           // Change file name inside RELEASES-ia32 to match renamed file
--- a/bin/package.js
+++ b/bin/package.js
@@ -468,8 +468,13 @@ function buildWin32 (cb) {
 
           // Change file name inside RELEASES-ia32 to match renamed file
           var relContent = fs.readFileSync(relPath, 'utf8')
-          relContent = relContent.replace(/full\.nupkg$/, '-ia32-full.nupkg')
-          fs.writeFileSync(relPath, relContent)
+          var relContent32 = relContent.replace(/full\.nupkg$/, '-ia32-full.nupkg')
+          fs.writeFileSync(relPath, relContent32)
+
+          if (relContent === relContent32) {
+            // Sanity check
+            throw new Error('Fixing RELEASE-ia32 failed. Replacement did not modify the file.')
+          }
 
           console.log('Windows: Renamed ia32 installer files.')
         }
--- a/bin/package.js
+++ b/bin/package.js
@@ -112,7 +112,7 @@ var darwin = {
   // Build for Mac
   platform: 'darwin',
 
-  // Build 64 bit binaries only.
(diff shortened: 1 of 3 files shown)


feross:
This is my test plan:

- [ ] Make two Windows releases using `master` with this PR: v0.15.1 and v0.15.99.
- [ ] Set the auto-update URL to a local version of `webtorrent-www` with supporting changes. (will send PR later)
- [ ] Upload the v0.15.99 release to GitHub.

Test 3 cases:

- [ ] On Windows 64-bit, install v0.15.1 (32-bit). Confirm that it updates to v0.15.99 (64-bit).
- [ ] On Windows 64-bit, install v0.15.1 (64-bit). Confirm that it updates to v0.15.99 (64-bit).
- [ ] On Windows 32-bit, install v0.15.1 (32-bit). Confirm that it updates to v0.15.99 (32-bit).

And, confirm that current app v0.15.0 (32-bit) can also update to new version:

- [ ] On Windows 64-bit, install v0.15.0 (32-bit). Confirm that it updates to v0.15.99 (32-bit).
  - Note: it cannot update straight to 64-bit, because this version is not sending the `sysarch` query parameter to the update server, so the server has to return the 32-bit binaries. This user will get onto the 64-bit version one release later.
- [ ] On Windows 32-bit, install v0.15.0 (32-bit). Confirm that it updates to v0.15.99 (32-bit).

dcposch:
Test plan sounds good to me

feross:
Can I get an LGTM so I can ship a new version right now?

bin/package.js
```diff
@@ -410,32 +416,82 @@ function buildWin32 (cb) {
...
+
+          // WebTorrent-vX.X.X-full.nupkg -> WebTorrent-vX.X.X-ia32-full.nupkg
+          fs.renameSync(
+            path.join(DIST_PATH, `${config.BUILD_NAME}-full.nupkg`),
+            path.join(DIST_PATH, `${config.BUILD_NAME}-ia32-full.nupkg`)
+          )
+
+          // Change file name inside RELEASES-ia32 to match renamed file
+          var relContent = fs.readFileSync(relPath, 'utf8')
+          relContent = relContent.replace(/full\.nupkg$/, '-ia32-full.nupkg')
```
dcposch: Might want to do something like this, just in case:
```js
var relContent32 = relContent.replace(...)
if (!relContent32.length || relContent32 === relContent) {
  throw new Error(...)
}
```

feross: Done

PR #933, opened 2016-09-18
Integration test: use /tmp

dcposch:
@feross


src/config.js
```diff
@@ -9,7 +10,7 @@ const APP_VERSION = require('../package.json').version
 
 const IS_TEST = isTest()
 const PORTABLE_PATH = IS_TEST
-  ? path.join(__dirname, '../test/tempTestData')
+  ? path.join(process.platform === 'windows' ? os.tmpdir() : '/tmp', 'WebTorrentTest')
```
feross: `windows` should be `win32`

test/config.js
```diff
@@ -1,6 +1,7 @@
 const path = require('path')
+const os = require('os')
 
-const TEST_DIR = path.join(__dirname, 'tempTestData')
+const TEST_DIR = path.join(process.platform === 'windows' ? os.tmpdir() : '/tmp', 'WebTorrentTest')
```
feross: Same comment as above

src/config.js
mathiasvr: Why not use `os.tmpdir()` on all platforms?

feross: The Mac `os.tmpdir()` is different with each run of the app (you can test this!), which causes the screenshots to be different, which won't work for our integration tests

PR #934, opened 2016-09-19
Integration test reliability

dcposch:
@feross


dcposch:
@Flet yeah. Electron 1.4.0 has two regressions:
1. The DevTools console is broken when using Spectron. It's empty and ignores all input. https://github.com/electron/spectron/issues/121
2. This error. https://github.com/electron/spectron/issues/121

dcposch:
Reliability is getting pretty decent. I just ran it 10 times in a row on Windows:

```
> npm run integration test && npm run integration test && npm run integration test && npm run integration test && npm run integration test && npm run integration test && npm run integration test && npm run integration test && npm run integration test && npm run integration test && echo "hooray"

<30+ minutes later>

"hooray"
```

package.json
```diff
@@ -24,7 +24,7 @@
     "deep-equal": "^1.0.1",
     "dlnacasts": "^0.1.0",
     "drag-drop": "^2.12.1",
-    "electron": "1.4.0",
+    "electron": "^1.3.5",
```
feross: We have always specified the Electron version exactly, never with a range. Can you fix this?

src/renderer/lib/torrent-summary.js
```diff
@@ -53,5 +53,6 @@ function getByKey (state, torrentKey) {
 function getFileOrFolder (torrentSummary) {
   const ts = torrentSummary
   if (!ts.path || !ts.files || ts.files.length === 0) return null
-  return path.join(ts.path, ts.files[0].path.split('/')[0])
+  const dirname = ts.files[0].path.split(path.sep)[0]
```
feross: Good catch!

test/setup.js
```diff
@@ -108,10 +112,21 @@ function compareIgnoringTransparency (bufActual, bufExpected) {
...
       const i = (y * w + x) * 4
       if (de[i + 3] === 0) continue // Skip transparent pixels
-      if (da[i] !== de[i] || da[i + 1] !== de[i + 1] || da[i + 2] !== de[i + 2]) return false
+      const ca = (da[i] << 16) | (da[i + 1] << 8) | da[i + 2]
+      const ce = (de[i] << 16) | (de[i + 1] << 8) | de[i + 2]
+      if (ca === ce) continue
+
+      // Add pixel diff to running sum
+      // This is necessary on Windows, where rendering apparently isn't quite deterministic
+      // and a few pixels in the screenshot will sometimes be off by 1. (Visually identical.)
```
feross: Nice!

test/setup.js
```diff
@@ -168,3 +186,13 @@ function extractImportantFields (parsedTorrent) {
...
   return { infoHash, name, announce, urlList, comment, 'private': priv }
 }
+
+function copy (pathFrom, pathTo) {
+  try {
+    fs.copySync(pathFrom, pathTo)
+  } catch (e) {
+    // Windows sometimes gives us an EPERM error even tho the copy happened...
+    if (process.platform !== 'win32' || e.code !== 'EPERM') throw e
+    console.log('ignoring windows copy EPERM error', e)
```
feross: This is janky. Are we sure there's not an open file handle at the path we are copying to? That could cause this error. See https://github.com/nodejs/node-v0.x-archive/issues/6599#issuecomment-30089704

test/test-torrent-list.js
```diff
@@ -28,35 +28,29 @@ test('torrent-list: start, stop, and delete torrents', function (t) {
   setup.resetTestDataDir()
 
   const app = setup.createApp()
-  setup.waitForLoad(app, t, {offline: true})
+  setup.waitForLoad(app, t)
     .then(() => app.client.waitUntilTextExists('.torrent-list', 'Big Buck Bunny'))
     // Mouse over the first torrent
     .then(() => app.client.moveToObject('.torrent'))
-    .then(() => setup.wait())
```
feross: Great that we're getting rid of all these `wait()` calls.

package.json
feross: I think we used to use this field in a script in a way where the `^` would break things, but I just checked and we don't do that anymore.

But I just realized a bigger issue: `^1.3.5` will install anything between 1.3.5 and 2.0.0 (exclusive). So, this still installs 1.4.0 👊

test/setup.js
dcposch: Turns out the copy is actually **not** working. This pretty much just skips that one part of the test on Windows.

There is a bug either in node or `fs-extra`. Windows lets the test process create files and folders underneath `C:\Windows\Temp`, but when you try `fs.copySync` to that location you get this error.

The path that we're copying doesn't exist yet, so definitely no open file handles. This is also the path that `fs.copySync` is trying to open under the hood.

package.json
dcposch: Fixed

feross: The line still looks unchanged to me. 😕

test/setup.js
feross: So weird. Can you add a TODO comment about this before merging?

package.json
dcposch: ya forgot to push

test/setup.js
dcposch: done

PR #944, opened 2016-09-20
Replace babel with bublé

feross:
Fixes https://github.com/feross/webtorrent-desktop/issues/940

Pros of [`buble`](https://buble.surge.sh/) over `babel`:

- No configuration (a la standard)
- Runs twice as fast, for quicker development
- Converts everything to ES5 (which is likely to be faster than ES6,
but this is untested)
- Easy to swap Babel back in -- low commitment

Cons:
- Less battle-tested than Babel, but recommended by React core
developer so probably not too bad
- No babel plugin support, but we're not using that right now anyway.
Can switch back to babel if we need that later

BEFORE:

$ time npm run build

> webtorrent-desktop@0.16.0 build /Users/feross/code/webtorrent-desktop
> babel --quiet src --out-dir build

npm run build  3.07s user 0.27s system 115% cpu 2.902 total

AFTER:

$ time npm run build

> webtorrent-desktop@0.16.0 build /Users/feross/code/webtorrent-desktop
> buble src --output build

npm run build  1.38s user 0.16s system 114% cpu 1.343 total

### DIFF
--- a/bin/package.js
+++ b/bin/package.js
@@ -39,9 +39,9 @@ function build () {
   rimraf.sync(DIST_PATH)
   rimraf.sync(BUILD_PATH)
 
-  console.log('Babel: Building JSX...')
+  console.log('Build: Transpiling to ES5...')
   cp.execSync('npm run build', { NODE_ENV: 'production', stdio: 'inherit' })
-  console.log('Babel: Built JSX.')
+  console.log('Build: Transpiled to ES5.')
 
   var platform = argv._[0]
   if (platform === 'darwin') {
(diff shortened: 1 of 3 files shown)


feross:
Fixes https://github.com/feross/webtorrent-desktop/issues/940

Flet:
Its probably negligible, but is there a big difference in time running integration tests?

feross:
> is there a big difference in time running integration tests?

I think the `npm run build` happens only once in the integration tests. But if ES5 is faster than ES6, then maybe we'll see that in the time it takes integration tests to complete...

Flet:
>ES5 is faster than ES6, then maybe we'll see that in the time it takes integration tests to complete

Yeah, that's what I'm curious about :)

PR #945, opened 2016-09-21
Use subtler UI sounds

### LINKED ISSUE #858: Need Option to disable sound on Mac
0.12.0 (0.96.3)
macOS Sierra 
**What did you do?**
Use the app
Have an option to disable sound
**What actually happened?**
Dont see an option to disable sound

feross:
Fixes #858


src/renderer/lib/sound.js
```diff
@@ -19,19 +19,19 @@ const sounds = {
   },
   DELETE: {
     url: 'file://' + path.join(config.STATIC_PATH, 'sound', 'delete.wav'),
-    volume: VOLUME
+    volume: VOLUME * 0.8
```
dcposch: Doesn't that mean these sounds are now louder than before? 0.4 * 0.8 > 0.15

Flet: The sounds themselves are also different @dcposch :)

feross: @dcposch You're right. The delete sound is unchanged, so it is now louder than before. Will fix.

feross:
I just tweaked the sound levels again. Can I get a review?

PR #949, opened 2016-09-21
State cleanup, rename, & tweaks

feross:
- Make State.save() always throttle calls -- since that's the common
case.

- Immediate saves are now the exception, with State.saveImmediate().

- The function is called State.save(), so the dispatch event should be
'stateSave'.

### DIFF
--- a/src/renderer/lib/state.js
+++ b/src/renderer/lib/state.js
@@ -1,15 +1,18 @@
 const appConfig = require('application-config')('WebTorrent')
+const debounce = require('debounce')
 const path = require('path')
 const {EventEmitter} = require('events')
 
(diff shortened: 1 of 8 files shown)


src/renderer/lib/state.js
```diff
@@ -98,7 +102,7 @@ function getDefaultPlayState () {
 }
 
 /* If the saved state file doesn't exist yet, here's what we use instead */
-function setupSavedState (cb) {
+function setupstateSaved (cb) {
```
dcposch: setupStateSaved?

feross: Good catch.

PR #950, opened 2016-09-21
Make config static, not dynamic

feross:
I don't think we should dynamically generate the config object -- it
makes it harder to figure out what's going on.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -14,11 +14,14 @@ const PORTABLE_PATH = IS_TEST
 const IS_PORTABLE = isPortable()
 const IS_PRODUCTION = isProduction()
 
+const UI_HEADER_HEIGHT = 38
+const UI_TORRENT_HEIGHT = 100
+
 console.log('Production: %s portable: %s test: %s',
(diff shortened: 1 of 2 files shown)


src/config.js
```diff
@@ -92,20 +95,17 @@ let cfg = {
   WINDOW_MAIN: 'file://' + path.join(__dirname, '..', 'static', 'main.html'),
   WINDOW_WEBTORRENT: 'file://' + path.join(__dirname, '..', 'static', 'webtorrent.html'),
 
+  WINDOW_INITIAL_BOUNDS: {
+    width: 500,
+    height: UI_HEADER_HEIGHT + (UI_TORRENT_HEIGHT * 6) // header + 6 torrents
```
Flet: We're losing the ability to make this dynamically match the length of the default torrent list, but it is easier to understand.

feross: Yeah, you're right. But I don't think the window should get taller than this, even if we add more default torrents. Because then it will be taller than some screens!

feross:
Thanks for the review, @Flet!

PR #951, opened 2016-09-21
Another round of integration test improvements

dcposch:
@feross want to check if this works on MacOS Sierra?


feross:
Tests don't pass on MacOS Sierra. The font change is too extreme. Honestly, not sure I like it. The letter spacing seems too wide to me.

El Capitan:

screen shot 2016-09-21 at 5 04 01 pm


Sierra:
screen shot 2016-09-21 at 5 03 04 pm

This could be an Electron or Chromium bug. Investigating this...

feross:
Just to followup, the font difference is probably a Chrome bug on macOS Sierra. I left a comment on this existing issue: https://bugs.chromium.org/p/chromium/issues/detail?id=627609#c20

PR #952, opened 2016-09-22
main: Start loading state before app is ready

feross:
As mentioned in
https://github.com/feross/webtorrent-desktop/pull/827#discussion_r799219
59

We should load the state outside the app.on('ready') handler so there's
a chance it's ready by the time 'ready' fires.

This improves startup time by roughly 50ms on my Macbook 12".

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -4,6 +4,8 @@ const electron = require('electron')
 const app = electron.app
 const ipcMain = electron.ipcMain
 
+const parallel = require('run-parallel')
+
 const announcement = require('./announcement')
 const config = require('../config')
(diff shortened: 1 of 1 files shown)


src/main/index.js
```diff
@@ -4,6 +4,8 @@ const electron = require('electron')
 const app = electron.app
 const ipcMain = electron.ipcMain
 
+const parallel = require('run-parallel')
```
dcposch: This works, but I think it's cleaner to use `Promise.all`

feross: Feel free to PR that, but then also remove other places where `run-parallel` is used so it can be removed from `package.json`.

PR #954, opened 2016-09-22
use arch to determine OS_SYSARCH

omrilitov:
Hi,
I replaced some of the code with the package arch, which was exactly the same 😄


Flet:
The circle is complete! ⚪

PR #957, opened 2016-09-22
startup perf: Reduce require() calls

feross:
Every require() that we do before the users sees UI reduces startup
time.

I used the following code (added to index.js) to log every require()
call in the main process:

```js
var Module = require('module')
var required = {}
Module.prototype.require = function (orig) {
  return function (id) {
    if (!required[id]) {
      required[id] = true
      console.log(`${id}   (from ${this.filename})`)
    }
    return orig.apply(this, arguments)
  }
}(Module.prototype.require)
```

From this I was able to learn that lots of modules were being required
that aren't actually used until later.

I also sent this related PR to eliminate another few require()s:
https://github.com/LinusU/node-application-config/pull/4

This increases startup time by 50ms.

We'll probably realize much bigger gains by following this same
procedure for the renderer process.

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -2,23 +2,15 @@ console.time('init')
 
 const electron = require('electron')
 const app = electron.app
-const ipcMain = electron.ipcMain
 
 const parallel = require('run-parallel')
 
-const announcement = require('./announcement')
 const config = require('../config')
 const crashReporter = require('../crash-reporter')
-const dialog = require('./dialog')
-const dock = require('./dock')
 const ipc = require('./ipc')
 const log = require('./log')
 const menu = require('./menu')
-const squirrelWin32 = require('./squirrel-win32')
 const State = require('../renderer/lib/state')
-const tray = require('./tray')
-const updater = require('./updater')
-const userTasks = require('./user-tasks')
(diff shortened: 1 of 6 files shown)


dcposch:
You mean decreases startup time by 50ms, right?

src/main/index.js
```diff
@@ -169,12 +170,16 @@ function sliceArgv (argv) {
 function processArgv (argv) {
   let torrentIds = []
   argv.forEach(function (arg) {
-    if (arg === '-n') {
-      dialog.openSeedDirectory()
-    } else if (arg === '-o') {
-      dialog.openTorrentFile()
-    } else if (arg === '-u') {
-      dialog.openTorrentAddress()
+    if (arg === '-n' || arg === '-o' || arg === '-u') {
```
dcposch: You can also write `['-n', ...].includes(arg)`

feross:
> You mean decreases startup time by 50ms, right?

Yep

src/main/index.js
feross: Right!

Flet:
...that snippet could be a nice little utility package on its own! 👍

PR #960, opened 2016-09-23
Add more peer ID entropy

feross:
This still keeps the peer ID printable as ascii, while adding lots more entropy


feross:
Teamwork! 👏 

screen shot 2016-09-23 at 6 10 54 pm

PR #962, opened 2016-09-23
Integration test fixes

dcposch:
@feross


Flet:
unrelated UX/UI side note: I like the simplicity of the checkbox for starting/stopping, but it was not immediately clear to me what it did.

mathiasvr:
The rows of pause checkboxes looks unnecessary to me. I think the checkbox should be hidden when not mousing over so you only see the title on idle torrents like before. Looks good otherwise.

dcposch:
CC @feross 
I agree having the default torrents all say "Paused" the first time the user opens the app doesn't seem very clear.

We wanted checkboxes on every torrent so you can see at a glance which ones are active. Currently, we use the blue circle next to the red download button, which has two problems:
1. You can only see whether a torrent is active by mousing over it
2. At least some of our users couldn't tell what the button does, so they filed issues like "add a button to stop seeding"

mathiasvr:
> 1. You can only see whether a torrent is active by mousing over it
You can still tell if they are active because only active torrents show progress info, while inactive ones displays the title. This is the part from the old design i liked better.

Flet:
Cool. I think the checkbox style will work with a little more iteration. I'll save further feedback/opinion for future issues/PRs pertaining to UX/style :)

dcposch:
Updated to the include UI improvements. See #963 #967 #966

dcposch:
I ran the integration tests three times in a row each on Mac and Windows. They should be reliable now.

The tests reflect how the app looks on master right now and what's shipping in v0.17.0

To run tests, you still have to use the following webtorrent.js branch via `npm link`

https://github.com/feross/webtorrent/pull/928

feross:
Sorry -- will get webtorrent PR merged now!

PR #964, opened 2016-09-23
avoid saving window bounds when on player

### LINKED ISSUE #961: App saves the wrong window bounds when quitting while in the player
### Steps to reproduce
- Start playing a video. The window resizes to fit.
- Quit the app (Cmd+Q)
- Restart the app

### Expected result
The torrent list (home page) should open at the size it was before, as if you had hit Esc to quit out of the video.

### Actual result
The window opens at the same huge size as the video.

We can fix this by only saving window bounds on resize when *not* in the player.

CC @Flet https://github.com/feross/webtorrent-desktop/pull/827

Flet:
fixes #961

### DIFF
--- a/src/renderer/main.js
+++ b/src/renderer/main.js
@@ -466,8 +466,10 @@ function onFullscreenChanged (e, isFullScreen) {
 }
 
 function onWindowBoundsChanged (e, newBounds) {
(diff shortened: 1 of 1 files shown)


dcposch:
Fixes #961

PR #963, opened 2016-09-23
WebTorrent Desktop v0.17.0

dcposch:
@feross @Flet @grunjol @mathiasvr

PR #966, opened 2016-09-23
set status to Idle if paused

Flet:
"Paused" status is a bit confusing to show on the torrent list, especially since it means something different if on the player view.

This PR changes it to say "Idle" instead, which helps convey the torrent state better and avoids using the same word for two things.

...just floating this PR out as an idea. If its not something we should do go ahead and close it :)

Screenshot:
image

### DIFF
--- a/src/renderer/pages/torrent-list-page.js
+++ b/src/renderer/pages/torrent-list-page.js
@@ -174,7 +174,10 @@ module.exports = class TorrentList extends React.Component {
     }
 
     function renderTorrentStatus () {
-      return (<span>{capitalize(torrentSummary.status)}</span>)
+      let status = torrentSummary.status === 'paused'
+        ? 'Idle'
+        : capitalize(torrentSummary.status)
+      return (<span>{status}</span>)
     }
   }
 


Flet:
Also apologies in advance for breaking integration tests :trollface:

dcposch:
How about

[ ] Start

dcposch:
And np the integration tests are already borked :)

I'll update #962 once we decide on a ui

feross:
> How about [ ] Start

The text next to the checkbox is the current status of the torrent. So, Downloading vs. Idle makes sense. Downloading vs. Start doesn't.

Actions like "Start" belong on buttons (or links), IMO

dcposch:
"Idle" is OK for torrents that have never been started, but doesn't make sense for paused downloads 

How about

#### New torrent, like the default torrents when you first install the app

No label, just the checkbox

[ ]

#### Downloading

[x] Downloading [===   ] 60% . ...

#### Paused, not 100% done

[ ] Paused [===   ] 60% . ...

#### Seeding

No progress bar:

[x] Seeding 100% . ...

#### Paused, 100% done

[ ] Not seeding 100% . ...

Flet:
It's a little more complex to implement but I like it @dcposch! Much clearer for the user.

dcposch:
PR here: #963

Flet:
Thanks @dcposch :)

PR #967, opened 2016-09-24
use a progress bar from material-ui

noamokman:
Hi there!

This replaces the built in progress bar for chrome with the progress bar from material-ui.
This helps make a unified experience across platforms.

image

Thanks!

### DIFF
--- a/src/renderer/pages/torrent-list-page.js
+++ b/src/renderer/pages/torrent-list-page.js
@@ -1,6 +1,7 @@
 const React = require('react')
 const prettyBytes = require('prettier-bytes')
 const Checkbox = require('material-ui/Checkbox').default
+const LinearProgress = require('material-ui/LinearProgress').default
 
 const TorrentSummary = require('../lib/torrent-summary')
 const TorrentPlayer = require('../lib/torrent-player')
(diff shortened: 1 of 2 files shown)


src/renderer/pages/torrent-list-page.js
```diff
@@ -1,6 +1,7 @@
 const React = require('react')
 const prettyBytes = require('prettier-bytes')
 const Checkbox = require('material-ui/Checkbox').default
+const LinearProgress = require('material-ui/LinearProgress').default
```
feross: I've been trying to keep the material-ui requires separated from the rest of the requires by a newline

src/renderer/pages/torrent-list-page.js
```diff
@@ -120,7 +121,21 @@ module.exports = class TorrentList extends React.Component {
 
     function renderProgressBar () {
       const progress = Math.floor(100 * prog.progress)
-      return (<progress value={progress} max='100'>{progress}%</progress>)
+      const styles = {
+        wrapper: {
+          display: 'inline-block',
+          marginRight: '8px'
```
feross: It's not necessary to specify `px` here. You can just use a number.

feross: Sent PR to fix

Issue #947, opened 2016-09-21
Update auto-launch to 5.0.0


feross:
`auto-launch` 5.0.0 was just released: https://github.com/Teamwork/node-auto-launch/releases/tag/5.0.0

Let's remove the macOS workaround added in https://github.com/feross/webtorrent-desktop/pull/923 if we can.

@dcposch @noamokman

noamokman:
PR is ready 👍

feross:
Closing re @dcposch's comment here: https://github.com/feross/webtorrent-desktop/pull/968#issuecomment-249455373

PR #968, opened 2016-09-24
update auto-launch

noamokman:
Hi,
This is about #947,
Tested on windows 😄 

Need help testing on mac.

Thanks!


dcposch:
i just tested this on mac.

testing is a kind of a pain: you have to package the app, which is slow, then change the prefs, restart, change the prefs, restart again

i don't think it works

we should just stay on auto-launch 0.4.1. if it ain't broke, don't fix it.

@feross

PR #973, opened 2016-09-27
Pixel values should use numbers, not strings

feross:
React converts to string and adds 'px' automatically


feross:
Thanks for the all the reviews, @demoneaux and @mathiasvr!

PR #976, opened 2016-09-27
Switch fs.statSync to fs.accessSync, which is faster

feross:
With `fs.accessSync`, we check that the folder is readable/writable by us.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -147,12 +147,14 @@ function isPortable () {
     return false
   }
 
-  let portablePathExists = false
   try {
-    portablePathExists = !!fs.statSync(PORTABLE_PATH)
-  } catch (err) {}
-
-  return portablePathExists
+    // This line throws if the "Portable Settings" folder does not exist, and does
+    // nothing otherwise.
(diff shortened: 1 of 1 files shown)


src/config.js
```diff
@@ -147,12 +147,12 @@ function isPortable () {
     return false
   }
 
-  let portablePathExists = false
+  let pathExists = false
   try {
-    portablePathExists = !!fs.statSync(PORTABLE_PATH)
+    pathExists = !!fs.accessSync(PORTABLE_PATH, fs.constants.R_OK | fs.constants.W_OK)
```
mathiasvr: `accessSync` does not return a value, so pathExists will be `undefined`.

feross: Looks like I need to take a deep breath and slow down today... 😊 Will fix.

feross: Okay -- it's updated. Thanks for the thorough review, @mathiasvr!

mathiasvr: I was honestly in doubt myself when reading the docs. It's so strange that the purpose of this method is to throw an error or just do nothing?? 😄

feross:
Just tested on Windows, and everything looks good!

PR #978, opened 2016-09-27
Fix "Download path missing" for Portable App first run

feross:
Create "Portable Settings\Downloads" folder to prevent "Download path
missing" warning, on Windows.

screen shot 2016-09-27 at 1 02 37 pm


feross:

PR #981, opened 2016-09-27
Make Portable App also a Silent App

feross:
Fixes two portable app bugs, to make the app fully "silent", not just
"portable". This means that not only are all data files stored in the
"Portable Settings" folder, but the app should leave no trace on the
host system.

- Disable Electron's single instance mode so no lock file is created in
"%APPDATA%\Roaming\WebTorrent".

- Put Electron crash files, and other electron files into the "Portable
Settings\Temp" folder instead of "%APPDATA%\Temp".

### DIFF
--- a/bin/package.js
+++ b/bin/package.js
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 
 /**
- * Builds app binaries for Mac, Linux, and Windows.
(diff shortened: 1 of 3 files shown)


feross:

Would also be nice to get a post-merge review, @dcposch.

PR #983, opened 2016-09-28
Remove `.nodemonignore` file.

demoneaux:
Adding configuration files for every tool used will clutter up the repository, especially for a configuration as simple as this.


feross:
LGTM!

PR #988, opened 2016-09-28
Style: Use let/const instead of var

feross:


feross:

PR #991, opened 2016-09-30
external player: Only run special .app logic on darwin

feross:


dcposch:
This would only matter if someone has a Windows or Linux media player that ends in ".app" somehow, right?

feross:
@dcposch Yep, that's right. Just being extra careful.

PR #992, opened 2016-09-30
sound: Reduce volume even further

feross:


dcposch:
good call removing the preload. since the files are local we don't need that

PR #995, opened 2016-10-01
Make app startup in less than 1 second (~400ms improvement)

feross:
## Holy require!

Let's look at the number of times that `require()` is called before UI is shown. Ignore duplicate calls to require the same module, since those are instant.

### Before this PR

Main process: 36 calls
Renderer process: 557 calls (~1340ms) 

### After this PR

Main process: 31 calls
Renderer process: 282 calls (~960ms)

After this, time to init is basically around 940 - 1070ms on various runs, on my 12" Macbook, with the default torrents, all disabled at startup time.


package.json
```diff
@@ -18,25 +18,29 @@
...
     "dlnacasts": "^0.1.0",
     "drag-drop": "^2.12.1",
     "electron": "1.4.1",
     "es6-error": "^3.0.1",
-    "fs-extra": "^0.30.0",
+    "fn-getter": "^1.0.0",
     "iso-639-1": "^1.2.1",
     "languagedetect": "^1.1.1",
     "location-history": "^1.0.0",
+    "lodash.merge": "^4.6.0",
```
feross: We don't actually use this. It's there so we can use it instead of `lodash/merge` (see comments at top of `renderer/main.js` for explanation)

src/main/handlers.js
```diff
@@ -326,24 +326,28 @@ function installLinux () {
...
+
     const iconFilePath = path.join(
       os.homedir(),
       '.local',
       'share',
       'icons',
       'webtorrent-desktop.png'
     )
-    fs.mkdirp(path.dirname(iconFilePath))
-    fs.writeFile(iconFilePath, iconFile, function (err) {
```
feross: There was a race condition here where `iconFilePath` might not be created before `fs.writeFile` tries to write into it. Fixed.

src/main/handlers.js
```diff
@@ -352,7 +356,7 @@ function uninstallLinux () {
     'applications',
     'webtorrent-desktop.desktop'
   )
-  fs.removeSync(desktopFilePath)
+  rimraf(desktopFilePath)
```
feross: These calls don't need to be `sync` since none of the uninstall functions take callbacks. The caller is just assuming that the handlers will be uninstalled eventually.

src/main/index.js
```diff
@@ -122,17 +122,22 @@ function init () {
 }
 
 function delayedInit () {
+  if (app.isQuitting) return
```
feross: Fixes exception sometimes seen in terminal during development, where `delayedInit` gets called while app is quitting

src/main/index.js
```diff
@@ -122,17 +122,22 @@ function init () {
...
   const userTasks = require('./user-tasks')
 
   announcement.init()
   dock.init()
-  tray.init()
   updater.init()
   userTasks.init()
+
+  if (process.platform !== 'darwin') {
+    const tray = require('./tray')
```
feross: Doesn't hurt to require `./tray` on Mac, but it's faster if we don't

src/renderer/lib/telemetry.js
```diff
@@ -55,41 +73,6 @@ function reset () {
   }
 }
 
-function postToServer () {
```
feross: Simplified this by just using `simple-get`'s friendly options. We're already requiring `simple-get` so it costs nothing.

src/renderer/lib/telemetry.js
```diff
@@ -2,27 +2,31 @@
...
-const https = require('https')
-const os = require('os')
-const url = require('url')
 
 const config = require('../../config')
 
 let telemetry
 
 function init (state) {
   telemetry = state.saved.telemetry
```
feross: Posting to the server used to happen in `init()` which did a lot of work gathering stats. Now we do that at the last minute in `send()` which is called after a delay.

src/renderer/lib/telemetry.js
```diff
@@ -32,15 +36,29 @@ function init (state) {
   telemetry.torrentStats = getTorrentStats(state)
   telemetry.approxNumTorrents = telemetry.torrentStats.approxCount
 
-  if (config.IS_PRODUCTION) {
-    postToServer()
-    // If the user keeps WebTorrent running for a long time, post every 12h
-    setInterval(postToServer, 12 * 3600 * 1000)
```
feross: Interval moved to `renderer/main.js` so that `send()` just does an instant send.

src/renderer/lib/telemetry.js
```diff
@@ -223,7 +207,7 @@ function getElemString (elem) {
   let ret = elem.tagName
   try {
     ret += '.' + Array.from(elem.classList).join('.')
-  } catch (e) {}
+  } catch (err) {}
```
feross: `e` is for events. `err` for errors.

src/renderer/main.js
```diff
@@ -1,8 +1,34 @@
...
+ *   `darkBaseTheme.userAgent = false`. Return a fake object.
+ */
+let Module = require('module')
+const _require = Module.prototype.require
+Module.prototype.require = function (id) {
+  if (id === 'lodash/merge') id = 'lodash.merge'
+  if (id === 'inline-style-prefixer') return {}
+  return _require.apply(this, arguments)
+}
+
```
feross: This is quite the hack, but it's pretty safe since we're just intercepting the exact two modules listed there. And it improved startup time a pretty decent amount.

src/renderer/main.js
```diff
@@ -1,8 +1,34 @@
...
 console.time('init')
 
 const crashReporter = require('../crash-reporter')
 crashReporter.init()
 
+// Perf optimization: Start asynchronously read on config file before all the
+// blocking require() calls below.
+
+const State = require('./lib/state')
+State.load(onState)
```
feross: Moving this here was a pretty huge win too.

src/renderer/main.js
```diff
@@ -50,32 +69,52 @@ let state
...
   // Create controllers
   controllers = {
-    media: new MediaController(state),
-    update: new UpdateController(state),
-    prefs: new PrefsController(state, config),
-    playback: new PlaybackController(state, config, update),
-    subtitles: new SubtitlesController(state),
-    torrentList: new TorrentListController(state),
-    torrent: new TorrentController(state)
+    media: createGetter(() => {
```
feross: Lazy loading these controllers means that we don't need to inline all the `require()` calls inside the controllers. We just don't load the controllers that aren't used until later when they're needed.

src/renderer/pages/app.js
```diff
@@ -1,32 +1,36 @@
...
 
 const Views = {
-  'home': require('./torrent-list-page'),
-  'player': require('./player-page'),
-  'create-torrent': require('./create-torrent-page'),
-  'preferences': require('./preferences-page')
+  'home': createGetter(() => TorrentListPage),
+  'player': createGetter(() => require('./player-page')),
+  'create-torrent': createGetter(() => require('./create-torrent-page')),
+  'preferences': createGetter(() => require('./preferences-page'))
```
feross: Same as controllers, lazy loading the views means that we don't need to inline all the require calls within each view by hand.

src/renderer/webtorrent.js
```diff
@@ -225,23 +229,14 @@ function saveTorrentFile (torrentKey) {
   })
 }
 
-// Checks whether we've already resolved a given infohash to a torrent file
-// Calls back with (torrentPath, exists). Logs, does not call back on error
-function checkIfTorrentFileExists (infoHash, cb) {
-  const torrentPath = path.join(config.TORRENT_PATH, infoHash + '.torrent')
-  fs.exists(torrentPath, function (exists) {
```
feross: `fs.exists` is deprecated, so I swapped it for `fs.access`

test/setup.js
```diff
@@ -75,8 +78,13 @@ function endTest (app, t, err) {
 function screenshotCreateOrCompare (app, t, name) {
   const ssDir = path.join(__dirname, 'screenshots', process.platform)
   const ssPath = path.join(ssDir, name + '.png')
-  fs.ensureFileSync(ssPath)
```
feross: This `fs-extra` method is easy to replicate with a try-catch around `fs.readFileSync`

test/setup.js
```diff
@@ -200,14 +208,14 @@ function extractImportantFields (parsedTorrent) {
 }
 
 function copy (pathFrom, pathTo) {
+  const cpFile = require('cp-file')
   try {
-    fs.copySync(pathFrom, pathTo)
```
feross: `fs.copySync` from `fs-extra` recursively copies entire directories, while `cp-file` only handles individual files, but we never need to copy directories, so it's fine.

src/main/index.js
demoneaux: If you’re doing that, maybe you might want to only load `./user-tasks.js` on Windows only?

src/main/windows/main.js
```diff
@@ -45,18 +45,26 @@ function init (state, options) {
     y: initialBounds.y
   })
 
-  win.once('ready-to-show', function () {
+  win.loadURL(config.WINDOW_MAIN)
```
demoneaux: Does it make a difference where `loadURL` is? IMO it would be safer to have the `ready-to-show` event listener declared first, in case `loadURL` finishes loading very quickly before the listener is added (quite unlikely though).

src/main/index.js
feross: Good idea. Will update PR

src/main/windows/main.js
feross: Calling `win.loadURL()` does a `fs.readFile` internally to read the file from disk. That's async, so there's no way that it could fire `'ready-to-show'` on the same tick of the event loop. Since our event handler is registered on the same tick, and the event will be fired on at least the next tick, this should be safe.

Also, there are a bunch of other event handlers below this one that are registered after calling `win.loadURL`, so this better be the way things work!

src/renderer/lib/telemetry.js
dcposch: calling `get({method: 'POST', ...})` feels jank

src/renderer/pages/app.js
```diff
@@ -1,39 +1,48 @@
...
 
 const darkBaseTheme = require('material-ui/styles/baseThemes/darkBaseTheme').default
-const lightBaseTheme = require('material-ui/styles/baseThemes/lightBaseTheme').default
 const getMuiTheme = require('material-ui/styles/getMuiTheme').default
 const MuiThemeProvider = require('material-ui/styles/MuiThemeProvider').default
 
 const Header = require('../components/header')
 
+// Perf optimization: Needed immediately, so do not lazy load it below
+const TorrentListPage = require('./torrent-list-page')
```
dcposch: Wouldn't it require later in the same tick anyway?

test/setup.js
dcposch: Currently we don't.

I don't think we should do lazy requires in the integration tests. Startup time isn't an issue here.

src/renderer/lib/state.js
```diff
@@ -205,7 +210,13 @@ function load (cb) {
     if (err) return cb(err)
     const state = getDefaultState()
     state.saved = saved
-    migrations.run(state)
+
+    if (process.type === 'renderer') {
+      // Perf optimization: Save require() calls in the main process
```
demoneaux: Ah,  I didn't see this @dcposch. Maybe after this gets merged I'll re-look #997.

test/setup.js
feross: Sure thing. Did it out of habit. Will fix.

src/renderer/lib/telemetry.js
feross: I can require the module with a different name, or use the `.post()` helper method. 

`get.post()` is equivalent to `get({method: 'POST'})`

test/setup.js
feross: Will fix. Did it out of habit.

src/renderer/pages/app.js
feross: You're right, but putting here at the top is still faster in my testing. It's better to do requires (which block the event loop) while async I/O is happening that we're waiting on before we can continue the startup process. In this case, we're still waiting on the `state.load()` call in `renderer/main.js` while this require happens. It's better to it while that's happening, then to do it later when there's no I/O happening.

Would be useful to have a proper waterfall graph for this to prove this rationale is correct, but this is what my testing seemed to point to.

feross:
Thanks for the feedback everyone. Merged!

PR #996, opened 2016-10-02
Remove weird outline from “Create Torrent” button.

### LINKED ISSUE #989: Create Torrent button has odd outline

0.17.0

Mac

**What did you do?**

Drop file on app, which loads Create Torrent page.

Submit button should look normal.

**What actually happened?**

It has an odd outline. See pic:

screen shot 2016-09-29 at 11 53 37 am

I thought that this was just an artifact of the integration testing, but it's actually the way the button looks. Let's fix it.

demoneaux:
The outline was caused by the wrong class being applied to the button.

Closes #989.

### DIFF
--- a/src/renderer/pages/create-torrent-page.js
+++ b/src/renderer/pages/create-torrent-page.js
@@ -106,7 +106,7 @@ class CreateTorrentPage extends React.Component {
             }}
             onClick={dispatcher('cancel')} />
           <RaisedButton
-            className='control create-torrent'
+            className='control create-torrent-button'
             label='Create Torrent'
             primary
             onClick={this.handleSubmit} />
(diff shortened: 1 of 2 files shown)


demoneaux:
@feross do you think it's a good idea to prefix the class names of pages with something like `page-`? eg. `page-create-torrent`, `page-preferences`, etc. to provide a distinction?

dcposch:
This is my bad. The problem is that the `create-torrent` CSS class is used twice, once on the outer div for the whole Create Torrent page and again on the Create button.

There are some CSS rules meant for the whole page that are mistakenly applying to the button.

I would change the button class to `create-torrent-button` and modify the integration test accordingly.

demoneaux:
@dcposch updated!

dcposch:
How do you mark "requested changes" as done?

Github has complexified itself :(

demoneaux:
Yeah, not too sure. Maybe @feross can change his review or something?

feross:
Oh, I think if I "request changes" then I'm blocking the merge until I approve. I'll use "comment" instead of "request changes" in the future 👍

feross:
@demoneaux I like your idea of prefixing page classes with 'page-' to distinguish them.

Issue #911, opened 2016-09-07
Clicking on Magnet links doesn't do anything after 0.10.0


djgenesis:
**What version of WebTorrent Desktop?** (See the 'About WebTorrent' menu)
0.11.0 and above
**What operating system and version?**
Windows 10 Pro
**What did you do?**
Click a magnet link in any website.
**What did you expect to happen?**
Open it in Webtorrent Desktop
**What actually happened?**
Nothing

Until version 0.10.0 I clicked on magnet links and it opened just fine. I have tried:
-ALL versions after 0.10.0 -> 11.12.13.13.1 & 14
-All 3 browsers (Firefox, Chrome, MEdge) 
-both without making it the default torrent app and with making it
-3 devices. 2 laptops and a media box. All with windows 10
and none of these combinations opens a magnet link. I am no programmer but since 0.10.0 still works fine clicking on any magnet link I am guessing that there is something going on with 0.11.0 new feature of "New Preference to "Set WebTorrent as default handler for torrents and magnet links"". I am looking through the issues and there is no mention on this. Hey... I cant be the only one :)

Hope you figure this out cause I am stuck to v0.10.0 for ever.

djgenesis:
The only solution I found so far is through Firefox to set the default handler for magnet links inside Firefox settings to Webtorrent Desktop.

Goldob:
This is intentional. Some users found it intrusive for WebTorrent to automatically set itself as default magnet and torrent file handler. Try setting WebTorrent as default torrent app in `Edit > Preferences` menu.  See #771.

djgenesis:
As I mentioned above, I already did. I tried every combination with settings, browsers and WTD versions. Any version above 0.10.0 doesnt grab the magnet links. I have also checked all the issues. Am I the only one with this problem? Strange cause I tried the same things on 2 laptops, 1 tablet and one Media Box all on windows 10. If there is any more info logs etc I can provide, let me know.

anonymlol:
Same issue as #791 I assume.

djgenesis:
Yes. I also confirm that the problem exists in all versions after 0.10.0. Both fresh installs (after uninstallation) and upgrades from previous versions.

dcposch:
@djgenesis thanks for being so thorough. We'll try to fix this in the next release

djgenesis:
Happy to help guys. Keep up the great work.

djgenesis:
[...]

Issue #773, opened 2016-08-11
Improve torrent handler registration


feross:
On macOS, the .torrent association is automatic (defined in Info.plist). So if there's no other handler, WebTorrent will be default. If there's another default, then the user can still use "Open With..." to select WebTorrent.

On Windows / Linux, the current situation (after PR https://github.com/feross/webtorrent-desktop/pull/771) is that WebTorrent does nothing until the preference is toggled. This is not ideal.

Now the `.desktop` file will not get installed on Linux systems. But that file is needed for the app icon to show up correctly in the Unity launcher, and to support pinning etc. And there's no way to do "Open With..." on Windows, until the user toggles the preference option.

I think we need to change `handlers.js` to distinguish between installing handlers and making WebTorrent the default.
- install: associate with .torrent (but not as the default program), i.e. install .desktop file (linux), add to registry (windows)
- makeDefault: become default .torrent handler, register magnet handler (there can only be one)
- removeDefault: give up .torrent handler, unregister magnet handler
- uninstall: only used by the Windows program uninstaller to completely clean up registry entries, never called on Mac or Linux

@dcposch @mathiasvr Does that seem reasonable to you?

mathiasvr:
Looks good, but is the removeDefault necessary? I'm used to just overriding the default extension handler from the OS or another program.
I don't feel WebTorrent should handle setting the default `.torrent` handler but i guess it's okay. I would still separate this from magnet handlers, what if a user wants to have WebTorrent be the default magnet (or stream-magnet) handler but use another program for `.torrent` files?

feross:
> @mathiasvr is the removeDefault necessary? I'm used to just overriding the default extension handler from the OS or another program.

`removeDefault` isn't strictly necessary, but we want the setting to be a toggle with on/off instead of a button that you press with no change in the UI. I think that is clearer.

> I don't feel WebTorrent should handle setting the default .torrent handler but i guess it's okay

Interesting point. On macOS and Linux, we don't currently have code to make WebTorrent the default .torrent file handler. On Windows, we do. What do other apps do? I think uTorrent asks if you want to make it the default handler for torrent files and magnet links on startup.

> what if a user wants to have WebTorrent be the default magnet (or stream-magnet) handler but use another program for .torrent files?

[...]

PR #997, opened 2016-10-02
Ensure that default file/protocol handlers are re-installed after updating.

### LINKED ISSUE #791: torrent handler registration broken? (windows)
v 0.11.0 (0.96.1)

Windows 10 Pro x64

**What did you do?**
Click on magnet link (tried with and without using the toggle)

Expected the browser to ask to open with webtorrent-desktop (at least show it as an option)

**What actually happened?**
Webtorrent-desktop wasn't in the "open with" app list (for both magnet and stream-magnet). Torrent files (.torrent) weren't registered properly either (still defaulted to my other torrent client).

bnjmnt4n:
Previously, they were only installed when the preference was changed. This caused the handlers to point to non-existing files after updates occurred and older versions were removed by Squirrel.

Closes #791, #911.

### DIFF
--- a/src/renderer/lib/migrations.js
+++ b/src/renderer/lib/migrations.js
@@ -29,8 +29,21 @@ function run (state) {
   if (semver.lt(version, '0.17.0')) migrate_0_17_0(saved)
   if (semver.lt(version, '0.17.2')) migrate_0_17_2(saved)
 
+  if (semver.lt(version, config.APP_VERSION)) {
+    installHandlers(state.saved)
+  }
+
   // Config is now on the new version
   state.saved.version = config.APP_VERSION
+}
+
+// Whenever the app is updated,  re-install default handlers if the user has
+// enabled them.
+function installHandlers (saved) {
+  if (saved.prefs.isFileHandler) {
+    const ipcRenderer = require('electron').ipcRenderer
+    ipcRenderer.send('setDefaultFileHandler', true)
+  }
 }
 
 function migrate_0_7_0 (saved) {


demoneaux:
Also, it seems weird that `state.js` is in the `renderer` folder, when it is called by both the `renderer` and `main` processes. Also, I think `migrations.js` belongs in the `main` folder, as it is typically only called to migrate updates by the main process. Maybe we should check which process it is in to decide which code to call?

src/renderer/lib/migrations.js
```diff
@@ -43,10 +43,23 @@ function run (state) {
...
+
   // Config is now on the new version
   state.saved.version = config.APP_VERSION
 }
 
+// Whenever the app is updated,  re-install default handlers if the user has
+// enabled them.
+function installHandlers (saved) {
+  if (saved.prefs.isFileHandler) {
+    const handlers = require('../../main/handlers')
```
dcposch: I don't think you can do this from the renderer process.

You have to use `ipcRenderer` to tell the main process to reinstall the handlers.

demoneaux:
@dcposch I believe that currently `state.js` is run on both renderer and main processes, as it is loaded in the main process in order to restore window bounds.

feross:
> I believe that currently state.js is run on both renderer and main processes, as it is loaded in the main process in order to restore window bounds.

You're right. `state.js` is required in both processes at the moment. That came in with the bound restoration change. I added a change in #995 that makes migrations only run once in the renderer (for perf, and bug prevention). `state.save()` should never be called from the main process, or we'll have serious problems. Not sure what the cleanest solution is here.

demoneaux:
@feross I'll try to rebase on `master` and change the code to account for #773, which I hadn't seen before doing this PR.

demoneaux:
@feross for now I've just set up the IPC renderer to install the handlers after every update. I'll look into #773 in the future.

demoneaux:
Any updates?

DiegoRBaquero (approved):

PR #1005, opened 2016-10-03
package: Explicitly specify architectures to build for

feross:
Electron now offers a "armv7l" build on Linux. Because we were
specifying "all", we were generating an armv7l binary which was being
included in "WebTorrent-vX.X.X-linux.zip" along with the ia32 build,
which explains why it was 90MB+ in the last release.


feross:
Self-merging to get v0.17.1 out. Review welcome 👍

PR #1008, opened 2016-10-03
Update Mac integration test for Sierra 10.12

feross:
The font changed slightly on the next version of Mac OS. Let's update
the screenshots to match, since I already updated. (@dcposch - you'll
need to update too if you want the integration tests to pass on your
machine)


feross:
FYI: The font changed for the worse, because dynamic letter spacing no longer works. Chromium issue about the problem: https://bugs.chromium.org/p/chromium/issues/detail?id=627609

dcposch:
1. The diff is subtle, but if you use the Swipe feature you can see all they did was suckify the kerning. Lol
2. This makes me feel great about the integration tests.
3. LGTM

feross:
Fixed

PR #1013, opened 2016-10-05
Windows: Fix impossible to delete Wired CD torrent

feross:
Fixes: https://github.com/feross/webtorrent-desktop/issues/905

Remove the trailing dot (.) from the Wired CD torrent name, since folders/files that end in a trailing dot (.) or space are not deletable from Windows Explorer.


feross:
Alright! Here's a PR to remove the trailing `.` from the WIRED torrent's name. I also added a migration so existing users with an un-deletable WIRED torrent folder in their Downloads folder will have it renamed. Now either WebTorrent or Windows Explorer can delete it for them, if they want to do that.

The real fix (preventing torrents from ending with a `.`) is harder to make since it changes the low-level `parse-torrent-file` and we currently have assumptions all throughout `webtorrent` and `webtorrent-desktop` that the torrent name and folder name on disk will match. I've never seen a torrent in the wild that ends with a `.`, so this should make this error basically go away. And since this is really a Windows Explorer bug, I think this solution is 💯.

Issue #1015, opened 2016-10-05
TypeError: Cannot read property 'startPiece' of undefined


feross:
Fourth most common error from today (80 occurrences):

```
Processes: window, platforms: darwin linux win32, versions: 0.17.0 0.17.1
TypeError: Cannot read property 'startPiece' of undefined
    at renderLoadingBar (...\build\renderer\pages\player-page.js:572:24)
    at renderPlayerControls (...\build\renderer\pages\player-page.js:392:7)
    at Player.render (...\build\renderer\pages\player-page.js:33:24)
    at ReactCompositeComponentWrapper._renderValidatedComponentWithoutOwnerOrContext (...\node_modules\react\lib\ReactCompositeComponent.js:796:32)
    at ReactCompositeComponentWrapper._renderValidatedComponent (...\node_modules\react\lib\ReactCompositeComponent.js:819:34)
    at ReactCompositeComponentWrapper._updateRenderedComponent (...\node_modules\react\lib\ReactCompositeComponent.js:743:36)
    at ReactCompositeComponentWrapper._performComponentUpdate (...\node_modules\react\lib\ReactCompositeComponent.js:721:10)
    at ReactCompositeComponentWrapper.updateComponent (...\node_modules\react\lib\ReactCompositeComponent.js:642:12)
    at ReactCompositeComponentWrapper.receiveComponent (...\node_modules\react\li
```

Opening an issue to track this, in case the PR I'm about to send doesn't actually fix it.

PR #1016, opened 2016-10-05
Attempt to fix "TypeError: Cannot read property 'startPiece' of undefined"

feross:
Fixes: #1015

It wasn't clear what was causing this error, and I couldn't reproduce
it.

This PR attempts to resolve the issue by just guarding against the
exception.


feross:
Okay, using my admin privileges to merge this 😈 because I want to do a release tonight. There are a couple pretty nasty bugs in the current version that put React into a bad state and gray-screen the app.

Can I get a post-merge review from someone?

PR #1020, opened 2016-10-06
Cleanup torrent list context menu

bnjmnt4n:


src/renderer/controllers/torrent-list-controller.js
```diff
@@ -182,47 +182,51 @@ module.exports = class TorrentListController {
...
-      click: () => dispatch('confirmDeleteTorrent', torrentSummary.infoHash, true)
-    }))
-
-    menu.append(new electron.remote.MenuItem({
-      type: 'separator'
-    }))
-
-    if (torrentSummary.files) {
-      menu.append(new electron.remote.MenuItem({
+    const torrentStarted = !!(torrentSummary.files && torrentSummary.progress)
```
demoneaux: Hmmm, I realise this wouldn't work as `progress` isn't saved in the state. Will try to figure out something else.

Flet: Any more thoughts on this @demoneaux?

Issue #906, opened 2016-09-06
VLC Playback rework


Goldob:
We need some big and small changes to VLC Playback, especially after #871
### Playlists handling

Suppose we have VLC Playback option enabled in preferences. When opening a torrent,  open all files in VLC. This can be achieved by passing a sequence of `[local-url]`, `:meta-title=[title]` arguments. Return to list after playback.
When specifying a single file, play only that file and return to list.

The tricky part is when we don't the option enabled, but one of our files causes `unsupported-media-modal`. Suppose that you play through a torrent and somewhere in the middle you encounter a troublesome file. The modal appears and at the moment you can either return to list or open that one file in VLC and... return to list. This is a UX no-no.
First of all, we should definetely play next file after returning from VLC, but there should also be an option to skip that file. Remember that we can traverse playlist either left or right direction, so we should either change UI to let that happen when modal appears or just add an option to ignore all unplayable files.

Also, I'm not sure what do do about other external players, i.e. the ones specified in preferences by the user. Can we make an assumption that passing multiple arguments will open files in playlists and not result in error? Maybe stay with current behavior and just pass one file? Do we even need it (see _Telemetry_ section)?
### Playback controls

Disable Playback menu and Windows thumbnail icons while in external player. We don't utilize any remote controlling, so these are misleading at least. I also confess to introducing a bug that causes many instances of VLC to spawn when pressing Skip Next/Skip Previous repeatedly.
### Telemetry

This is quite a lot of work and it would be nice to know how do we need to prioritize it.  I'm curious how many users actually utilize Play in VLC feature and how. I think it would be nice to add simple information about that in telemetry, e.g. percantage of files opened in external player and how often did it happen due to option in preferences and how often due to error.

M0Rf30:
Why not use webchimera technology? I ask as a newbie

Goldob:
@M0Rf30 see issue #938

PR #1026, opened 2016-10-11
Playback telemetry

Goldob:
Relates to #906.

Log play attempt for each item in playlist separately, instead of doing it only once on player close. Include information about external playbacks.

### DIFF
--- a/src/renderer/controllers/media-controller.js
+++ b/src/renderer/controllers/media-controller.js
@@ -1,7 +1,5 @@
-const electron = require('electron')
-
-const ipcRenderer = electron.ipcRenderer
-
+const {ipcRenderer} = require('electron')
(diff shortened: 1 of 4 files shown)


feross:
Will defer to @dcposch to review this.

src/renderer/controllers/media-controller.js
```diff
@@ -12,7 +10,7 @@ module.exports = class MediaController {
   }
 
   mediaSuccess () {
-    this.state.playing.result = 'success'
```
dcposch: Yes, I think `state.playing.result` can be deleted

feross:
@dcposch @Flet Is this ready to land? If it's ready, merge it!

PR #1042, opened 2016-10-14
save preferences immediately when changed

### LINKED ISSUE #805: Preferences should save instantly
I'm not sure what the whole `state.unsaved` vs. `state.saved` is about.

@dcposch Any ideas?

It creates weird situations where I can change a preference, quit the app without leaving the preferences page first, start the app again, preference was not saved.

Let's just do what every Mac app does and persist settings immediately when they are changed.

Flet:
Closes #805

Sorry @feross just noticed you're assigned to that issue.. hope this is not duplicated effort :/
- I took out the nested object path logic from `update` because there are not any preferences that use nested objects currently.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,53 +1,246 @@
 {
   "name": "webtorrent-desktop",
-  "version": "0.18.0",
+  "version": "0.20.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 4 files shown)


mathiasvr:
Hi, I took the liberty to rebase this PR on the latest master and remove new `unsaved` items.

PR #1043, opened 2016-10-14
Add a new Transfers menu to allow pause all and resume all torrents

### LINKED ISSUE #1027: Pause All button
0.17.1
Mac ,Linux ,Windows
**What did you do?**
try to pause all the torrents 
Pause all button 
**What actually happened?**
Button is absent

Can we add pause all button ? It will really help the users. 

thankyou!

karanjthakkar:
Fixes #1027

### DIFF
--- a/src/renderer/controllers/torrent-list-controller.js
+++ b/src/renderer/controllers/torrent-list-controller.js
@@ -133,9 +133,9 @@ module.exports = class TorrentListController {
       if (torrentSummary.status === 'downloading') {
         torrentSummary.status = 'paused'
         ipcRenderer.send('wt-stop-torrenting', torrentSummary.infoHash)
-        sound.play('DISABLE')
       }
     })
+    sound.play('DISABLE')
   }
 
   resumeAllTorrents () {
@@ -143,9 +143,9 @@ module.exports = class TorrentListController {
       if (torrentSummary.status === 'paused') {
         torrentSummary.status = 'downloading'
         this.startTorrentingSummary(torrentSummary.torrentKey)
-        sound.play('ENABLE')
       }
     })
+    sound.play('ENABLE')
   }
 
   toggleTorrentFile (infoHash, index) {
--- a/src/renderer/controllers/torrent-list-controller.js
+++ b/src/renderer/controllers/torrent-list-controller.js
@@ -130,7 +130,8 @@ module.exports = class TorrentListController {
 
   pauseAllTorrents () {
     this.state.saved.torrents.forEach((torrentSummary) => {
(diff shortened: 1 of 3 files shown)


karanjthakkar:
@feross Let me know what you think of this :)

src/renderer/controllers/torrent-list-controller.js
```diff
@@ -128,6 +128,26 @@ module.exports = class TorrentListController {
     }
   }
 
+  pauseAllTorrents () {
+    this.state.saved.torrents.forEach((torrentSummary) => {
+      if (torrentSummary.status === 'downloading') {
+        torrentSummary.status = 'paused'
+        ipcRenderer.send('wt-stop-torrenting', torrentSummary.infoHash)
+        sound.play('DISABLE')
```
Flet: Could `sound.play` be moved outside the `forEach` loop? This could get funky if there is a large number of torrents.

karanjthakkar: Sure. I can do that.

karanjthakkar:
@Flet Done!

karanjthakkar:
@Flet Bump! Is there anything stopping this from getting merged? Would love to do it so that this gets merged :)

src/renderer/controllers/torrent-list-controller.js
```diff
@@ -128,6 +128,26 @@ module.exports = class TorrentListController {
     }
   }
 
+  pauseAllTorrents () {
+    this.state.saved.torrents.forEach((torrentSummary) => {
+      if (torrentSummary.status === 'downloading') {
```
Flet: I noticed that if a torrent is seeding it is not paused... Does it make sense to paused torrents that are seeding too?

Flet:
I think @feross and @dcposch are taking a little break from this repo :)

I think updating to pause all including "Seeding" torrents makes sense. 

Happy to review and merge once this is done!

karanjthakkar:
@Flet Oh, I understand. I'll update the PR to include `seeding` torrents too.

karanjthakkar:
I've added the condition for checking `seeding` torrents while pausing. However since the old torrent state wasn't saved, while resuming, I have to set all the torrents to a `downloading` state. They're then moved to a `seeding` state. Is that cool?

Flet:
Yeah, I think its OK. 
It the spirit of keeping things simple, its a straightforward way to do it without adding separate "pause" states (and more menu items).

We can always make it more articulate later if folks feel pain.

feross:
LGTM!

PR #1068, opened 2016-11-08
Added the option to activate or deactivate sound notifications

kaotisk-hund:
I figured out how to make the option go in preferences menu but I can't find where I have to edit so the sound notifications go on/off. Any help?

### DIFF
--- a/src/renderer/lib/sound.js
+++ b/src/renderer/lib/sound.js
@@ -46,7 +46,7 @@ const sounds = {
   }
 }
 
-function play (name) {
+function play (name, state) {
   let audio = cache[name]
   if (!audio) {
     const sound = sounds[name]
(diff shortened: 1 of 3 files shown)


kaotisk-hund:
travis-ci says "state is not defined" but it runs ok and I got no error while running it. Any ideas on how to approach this better?

kaotisk-hund:
Now the checks are passing but it doesn't work...

kaotisk-hund:
How to gain access to saved configuration of webtorrent-desktop into lib/sound.js?

mathiasvr:
You could create an init function as in [lib/cast.js](https://github.com/feross/webtorrent-desktop/blob/4bb2056bc9c1a421815b97d03ffed512575dfde0/src/renderer/lib/cast.js#L33) to store the app state.

feross:
Thanks for the PR, but I don't want a preference for every trivial option that a user could want. I think we should just do the right thing, which is probably to remove the sounds entirely it seems.

PR #1136, opened 2017-02-15
material-ui@0.17

feross:
Tested and everything seems to work alright with the new version.


codealchemist:
**Win7** on VirtualBox:
image

image

image

**macOS Sierra** 10.12.2
image

image

codealchemist:
@feross Should we change the package version to `0.19.0`?
The rest looks good :)
Did you test it on Linux?

Cheers!

feross:
Thanks for the testing :) I already tested Mac, but a second pair of eyes doesn't hurt. 👍

> Should we change the package version to 0.19.0?

Isn't the latest 0.17.0? At least that's what I see! https://www.npmjs.com/package/material-ui

codealchemist:
I was referring to the [version](https://github.com/feross/webtorrent-desktop/blob/master/package.json) of **WebTorrent Desktop**.

It's currently `0.18.0` and I was wondering if after this changes it might get a bump on the minor.
But maybe it's ok not to update the version at all:
```
MAJOR version when you make incompatible API changes,
MINOR version when you add functionality in a backwards-compatible manner, and
PATCH version when you make backwards-compatible bug fixes.
```

What are your thoughts on it?
Thanks!

codealchemist:
I'm also thinking about the **changelog**.
The plan is to include other PRs and then release version `0.19.0` right?
Thx!

feross:
> I was referring to the version of WebTorrent Desktop.

Ahh, yes. We usually bump the version in package.json right before doing the release. That's so that there's a single commit like https://github.com/feross/webtorrent-desktop/commit/0ee7e80fd2dffb8520e9bb552009b01edfa65369 where the version changes. Also, we add a git tag there too, so `git checkout v0.18.0` checks you out to the specified version. 👍 

> The plan is to include other PRs and then release version 0.19.0 right?

Yes, I'd love to catch up on all the outstanding PRs, and do a release with all those changes!

codealchemist:
Cool! Thanks for the explanation dude!

PR #1137, opened 2017-02-15
electron@1.6.0

feross:
This version of Electron is 3 versions of Chromium newer, so there are
lots of performance improvements.

There is one known issue with this update:
https://github.com/electron/electron/issues/8694 which will be fixed in the next version.

But I think we should merge this anyway so that collaborators can try
out the new Chromium and V8 before we do a release with it. We might
find other bugs.


feross:
Yeah, would love to catch up on PRs and to not let them languish like this in the future.

PR #1138, opened 2017-02-16
Add StartupWMClass on Linux desktop file

bilelmoussaoui:
StartupWMClass is used to identify the opened applications under Linux with their desktop files, which means using the right icon and so on... more information can be found here https://standards.freedesktop.org/desktop-entry-spec/latest/ar01s05.html

The value assigned on  StartupWMClass can be easily found using `xprop WM_CLASS` and click other application you would like to get the WM_CLASS.

What does this affect? Nothing at all, The only difference is that if an icon theme includes an icon for WebTorrent it will be used instead of using the default one. Also, the default one will look blurry on some Desktop environments. And the user won't be able to get to use the actions defined in the desktop file while running the application if it's not correctly identified. This is mostly needed for non-native applications

Without a WM_CLASS:
screenshot from 2017-02-16 18-53-18

With a WM_CLASS: 
screenshot from 2017-02-16 18-55-01

### DIFF
--- a/static/linux/share/applications/webtorrent-desktop.desktop
+++ b/static/linux/share/applications/webtorrent-desktop.desktop
@@ -12,6 +12,7 @@ Path=/opt/webtorrent-desktop
 Exec=/opt/webtorrent-desktop/WebTorrent %U
 TryExec=/opt/webtorrent-desktop/WebTorrent
 StartupNotify=false
+StartupWMClass=WebTorrent
 Categories=Network;FileTransfer;P2P;
 MimeType=application/x-bittorrent;x-scheme-handler/magnet;x-scheme-handler/stream-magnet;
 


bilelmoussaoui:
Any news?

hicom150:
LGTM 👍

feross:
@hicom150 Is this change incorporated into your PR https://github.com/webtorrent/webtorrent-desktop/pull/1682 ?

hicom150:
> Is this change incorporated into your PR #1682

Yes it is already in my PR, so we can close this one 😉

BTW, thank you @bilelmoussaoui for this PR 👍

feross:
Cool, closing this. Thanks @bilelmoussaoui!

Issue #1130, opened 2017-02-12
Add option to disable notifications


adzie95:
The notification that tells me when a torrent has finished downloading is annoying and useless since i am already streaming the torrent in the webtorrent app. Please add an option to disable it if not remove it.

codealchemist:
Hey @adzie95, what do you think about the discussion on #1141 ?

PR #1141, opened 2017-02-19
Disable notifications

codealchemist:
Added notifications section to preferences; added ability to disable notifications, when enabled no notification is shown when a torrent finishes downloading.
This is a response to this request #1130 .

If we plan to add more notifications maybe we can add a notifications controller and then handle this preference just there.

Cheers!

### DIFF
--- a/src/renderer/controllers/torrent-controller.js
+++ b/src/renderer/controllers/torrent-controller.js
@@ -92,7 +92,12 @@ module.exports = class TorrentController {
       if (!this.state.window.isFocused) {
         this.state.dock.badge += 1
       }
-      showDoneNotification(torrentSummary)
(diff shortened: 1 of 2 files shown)


bil-elmoussaoui:
Please don't do this. It's up to the system to disable notification per app (this is what's done on every Linux desktop out there)
screenshot from 2017-02-23 15-25-35

codealchemist:
Good point @bil-elmoussaoui, the same thing can be done on Windows and OSX.
We can just use the OS to disable notifications without adding any new features to WebTorrent Desktop.

codealchemist:
We're not doing this one.
The same feature already exists at OS level.

PR #1146, opened 2017-03-02
fixes for standard v10

feross:


src/renderer/controllers/torrent-list-controller.js
```diff
@@ -281,7 +281,7 @@ module.exports = class TorrentListController {
 
 // Recursively finds {name, path, size} for all files in a folder
 // Calls `cb` on success, calls `onError` on failure
-function findFilesRecursive (paths, cb) {
+function findFilesRecursive (paths, cb_) {
```
feross: Hack so I don't have to rewrite this right now

PR #1148, opened 2017-03-08
Toggle fullscreen on pinch to zoom

ebraminio:
This patch makes webtorrent to do fullscreen (maximize/unmaximize) when pinch to zoom event (two finger zoom in/out gesture) happened on macOS instead probably less meaningful zoom. Have a look [here](https://github.com/dimsemenov/PhotoSwipe/issues/897) for more info about the approach.

To test this, open webtorrent on mac and do two finger zoom on webtorrent window with and without this patch. The intended gesture matches VLC on pinch to zoom action.

I hope you find this useful. Thanks!

### DIFF
--- a/src/renderer/main.js
+++ b/src/renderer/main.js
@@ -24,6 +24,7 @@ const State = require('./lib/state')
 State.load(onState)
 
 const createGetter = require('fn-getter')
+const debounce = require('debounce')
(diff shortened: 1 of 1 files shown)


feross:
Thanks, looks good!

Issue #1149, opened 2017-03-11
Feature - Auto add torrent files


pallavmac:
Auto add torrent files to web torrent if it finds one in a folder specified by the user. So If I add a torrent file to the downloads folder, auto adding it to the list when I open web torrent is an immensely useful feature. especially when needing to add multiple torrent files.

codealchemist:
You can also leave one machine running WebTorrent Desktop, listening for file changes on a specific folder, then remotely add torrent files to that folder and have them load without interacting with the UI at all. Really useful for batch downloading too.
This is a powerful feature.

codealchemist:
Hey @pallavmac, the feature is implemented in this PR: #1154  
If you want to start playing with it you can clone the repo and switch to `watch-folder` branch.
Any feedback you can provide is welcome.

Cheers!

PR #1154, opened 2017-03-24
Watch folder

codealchemist:
In response to #1149, this feature adds a preference to watch a folder for new `.torrent` files.
When a torrent file is saved on the watched folder it is automatically added to WebTorrent Desktop.
The feature can be enabled / disabled from preferences.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "auto-launch": "^4.0.1",
     "bitfield": "^1.0.2",
     "capture-frame": "^1.0.0",
+    "chokidar": "^1.6.1",
     "chromecasts": "^1.8.0",
     "cp-file": "^4.0.1",
     "create-torrent": "^3.24.5",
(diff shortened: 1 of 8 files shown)


Flet:
I really like this idea!

I think the wording on the the preferences could be phrased better as its not totally clear to me (and it doesn't mention "watching" at all)

Maybe:
```
(remove the header)

✅ Watch for new .torrent files and add them immediately.
Folder to watch: __________________________________________
```

Also, as this feature feels a little beyond the "keep things simple" user experience of this app, so it may be best to have this turned off by default.

codealchemist:
Hey @Flet, thanks a lot for your feedback!
I like the rewording and simplification.
About the header, I think we still need it because there aren't preferences without headers. If we just remove the header the preference will look like being part of another section.

Maybe we can rename *Downloads* to *Folders* and add it under that section.
@feross, @dcposch, what do you guys think about this change?

Cheers!

codealchemist:
Preview with proposed adjustments:

image

codealchemist:
Hey @Flet, can you provide a review to move this one forward?
Thanks!

PR #1165, opened 2017-04-27
Add support for subtitles on Chromecast

### LINKED ISSUE #530: Subtitles on Chromecast
New subtitles feature works _really_ well, thank you very much for that!

You think it's technically possible to have them show in Cast mode, at least for Chromecast ? From what I see the underlying lib should support it, right? (https://github.com/mafintosh/chromecasts#playersubtitlestrack-cb)

janza:
Support subtitles on Chromecast by creating a new http server for selected subtitles before calling player.play()

Closes #530

### DIFF
--- a/src/renderer/lib/cast.js
+++ b/src/renderer/lib/cast.js
@@ -13,6 +13,8 @@ module.exports = {
   setRate
 }
 
+const http = require('http')
+
 const config = require('../../config')
 const {CastingError} = require('./errors')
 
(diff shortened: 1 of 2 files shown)


talarari:
any update on this? i love the player but i usually cast my videos and not having subtitles is kind of a deal breaker, so this would be awesome

src/renderer/lib/cast.js
```diff
@@ -130,22 +132,51 @@ function chromecastPlayer () {
...
   }
 
+  function serveSubtitles (cb) {
+    const subtitles = state.playing.subtitles
+    const selectedSubtitle = subtitles.tracks[subtitles.selectedIndex]
+    if (!selectedSubtitle) {
+      cb()
+    } else {
+      ret.subServer = http.createServer((req, res) => {
+        if (!selectedSubtitle) {
```
jonkoops: I don't think this code will ever be reached because `selectedSubtile` has already been checked for a value in the above `else` clause.

src/renderer/lib/cast.js
```diff
@@ -130,22 +132,51 @@ function chromecastPlayer () {
     })
   }
 
+  function serveSubtitles (cb) {
```
jonkoops: In all functions in this file callback arguments are called `callback`, probably best to keep this consistent.

src/renderer/lib/cast.js
```diff
@@ -130,22 +132,51 @@ function chromecastPlayer () {
...
-        state.playing.location = 'local'
-        state.errors.push({
-          time: new Date().getTime(),
-          message: 'Could not connect to Chromecast. ' + err.message
-        })
-      } else {
-        state.playing.location = 'chromecast'
-      }
-      update()
+    serveSubtitles(function (subtitles) {
```
jonkoops: Since the `subtitles` variable is not actually an array of subtitles it makes more sense to use the same name as used before in the `serveSubtitles` function - namely `subtitlesUrl`.

jonkoops: Since the `subtitles` variable is not actually an array of subtitles it makes more sense to use the same name as used before in the `serveSubtitles` function - namely `subtitlesUrl`.

paulogr:
Hi all! Would be amazing to have this PR merged!!!

feross:
This looks like a useful PR. Will find a Chromecast to test it on as soon as I can

gonzaloserrano:
Hey there, I built the master branch and cherry-picked https://github.com/janza/webtorrent-desktop/commit/25087016ae1a715730c0beeb3bbd2ca563f8aa59 and it works!

paulogr:
Just did the same here like @gonzaloserrano did, and it worked for me too!

Nice job @janza!

@feross Any change to merge this PR?

jonkoops:
I've tested this PR and it seems to work on both my Phillips TV's built-in Chromecast, a Chromecast Gen. 1, and the Chromecast Ultra without any problems.

Issue #1185, opened 2017-06-11
Feature request: Sort files alphanumerically


vitlav:
WebTorrent 0.18.0

There is no order in torrent file list, downloaded parts can be show like
part3.avi
part2.avi
part4.avi
part1.avi

I expected alphanumeric sort.

bnjmnt4n:
@feross mentioned in https://github.com/webtorrent/webtorrent-desktop/issues/1076#issuecomment-264748538:

> We used to sort file names alphabetically, but we don't do it anymore. Instead, we just show the files in the order that the torrent file lists them. It's up to the torrent file creator to put the files in a sensical order.

dsernst:
I want to kindly add the the current UI on this is not intuitive and makes handling large groups of files very difficult. This isn't exactly a fringe user experience. It's clearly come up a number of times — e.g. the linked issues marked as dupes.

In particular, I have a 10gb torrent w/ 222 files, but only want to download a single one of them. It's possible to `Pause All` and then re-select just the needed file, but very difficult to find in such a large list if sorted by filesize.

This has come up for me a handful of times. The current lack of support for **Sort by Name** requires keeping a different torrent program, exclusively for this use-case. Seems... excessive.

-----

I do see the linked `WONTFIX` comment from @feross. 

And these from PR https://github.com/webtorrent/webtorrent-desktop/pull/1188:

![image](https://user-images.githubusercontent.com/6340841/57163256-8af73900-6da5-11e9-87f7-cfba04279eda.png)

-----

I certainly admire & agree with the sentiment to avoid creating a bloated Settings page. Yuck.

But maybe there's a simpler UI to handle this? What about a single context menu option on the list of files: "Sort by Name".

No addition to the Settings page is needed.
The control for this would be right where someone uses it. 
And not visible to anyone who doesn't right-click the file list.

-----

Current context menu:

<img alt="Screen Shot 2019-05-03 at 1 58 17 PM" src="https://user-images.githubusercontent.com/6340841/57165348-a36a5200-6dab-11e9-8fe2-21eed5dc11e7.png">

Proposal:

```
Remove From List
Remove Data File
-----
Show in Finder
-----
Copy Magnet Link to Clipboard
Copy Instant.io Link to Clipboard
Save Torrent File As...
-----
Sort by Name
```

Selecting `Sort by Name` would mark it checked, similar to Finder.app:

[...]

Issue #1076, opened 2016-11-15
Sort files in alphabetical order


Icx4Lemon:
**What version of WebTorrent Desktop?** (See the 'About WebTorrent' menu)
0.17.2 (0.97.2)
**What operating system and version?**
Windows 10 x64
**What did you do?**
I loaded a torrent with many files 
**What did you expect to happen?**
The files should have been sorted correctly
**What actually happened?**
The files were in a random order

feross:
We used to sort file names alphabetically, but we don't do it anymore. Instead, we just show the files in the order that the torrent file lists them. It's up to the torrent file creator to put the files in a sensical order.

PR #1188, opened 2017-06-24
Sort files alphanumerically #1185

yurtaev:


AleksandarDev:
Is this linked to any issue/discussion?

This feature was already requested and declined multiple times.

https://github.com/webtorrent/webtorrent-desktop/issues/1076

DiegoRBaquero:
Thank you for your PR. We want to keep a simple UI, with not a lot of configuration. It would be cool to offer the possibility to extend config with extensions.

PR #1216, opened 2017-08-02
only checking for the libappindicator1 package on debian systems

AshlessCinder:
This check only works on debian systems. It fails no matter what everywhere else, so this PR is to prevent false negatives where the tray works just fine without libappindicator1 (like my system).

### DIFF
--- a/src/main/tray.js
+++ b/src/main/tray.js
@@ -49,19 +49,33 @@ function initWin32 () {
  * Check for libappindicator1 support before creating tray icon
  */
 function checkLinuxTraySupport (cb) {
-  const cp = require('child_process')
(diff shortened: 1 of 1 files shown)


DiegoRBaquero:
@feross  should we recover this?

feross:
Yeah. The bot was a bit overzealous. We need to start add the "bug", "enhancement", "accepted", or "meta" tag to keep issues and PRs open. I'll be going through all the issues and PRs that it closed today and re-opening the necessary ones. :)

For this specific PR, it appears that the repository that submitted the pull request was deleted, so I can't reopen the PR. I'll make an issue linking to this PR so we won't forget to look at it.

PR #1217, opened 2017-08-03
Display download info on cast screen

### LINKED ISSUE #1070: Display downloading info while playing in VLC
Can't see any info on torrents downloading while in VLC mode.

cezar-carneiro:
Now we're able to see information such as download progress, file size, download speed, ETA and number of peers on the cast screen. 
The displayed information refers to the **currently playing file** only. Many torrents are composed of a single file, but some torrents are composed of several playable files. Showing information of the whole torrent (all files) would confuse the user, I think.
Any thoughts?
How it looks:
download_info_cast_screen

Closes #1070

### DIFF
--- a/src/renderer/pages/player-page.js
+++ b/src/renderer/pages/player-page.js
@@ -309,24 +309,71 @@ function renderCastScreen (state) {
 
   const isStarting = state.playing.location.endsWith('-pending')
   const castName = state.playing.castName
+  const fileName = state.getPlayingFileSummary().name || ''
   let castStatus
   if (isCast && isStarting) castStatus = 'Connecting to ' + castName + '...'
   else if (isCast && !isStarting) castStatus = 'Connected to ' + castName
   else castStatus = ''
 
   const prog = state.getPlayingTorrentSummary().progress || {}
-  let fileProgress = 0
-  if (prog.files) {
-    const file = prog.files[state.playing.fileIndex]
-    fileProgress = Math.floor(100 * file.numPiecesPresent / file.numPieces)
-  }
-
-  const downloaded = prettyBytes(prog.downloaded)
-  const total = prettyBytes(prog.length || 0)
 
   // Show a nice title image, if possible
   const style = {
     backgroundImage: cssBackgroundImagePoster(state)
+  }
+
+  function renderDownloadProgress () {
+    if (!prog.files) return
+
+    const fileProg = prog.files[state.playing.fileIndex]
+    const fileProgress = fileProg.numPiecesPresent / fileProg.numPieces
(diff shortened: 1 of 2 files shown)


DiegoRBaquero:
I think it looks great (visually).

KingMaxer:
Currently, all software versions exist with a bug about the display of the file size.
Can you fix it ?
The file size is calculated in 1MB=1000KB now, not in accordance with 1MB=1024KB.

cezar-carneiro:
@DiegoRBaquero Thanks! Any suggestions for the code?
@KingMaxer I notice that, but I wouldn't say it's a bug. Today there are two standards for displaying sizes: SI which is base-10 and IEC which is base-2.The guys from WebTorrent Desktop apparently choose SI (base-10). [https://wiki.ubuntu.com/UnitsPolicy](https://wiki.ubuntu.com/UnitsPolicy)
I could change the standard if people would prefer IEC. Is there an issue for this?

Thanks for the comments, guys! :wink:

DiegoRBaquero:
@KingMaxer Where do you get that idea? We use pretty bytes which uses base-2.

cezar-carneiro:
Pardon me @DiegoRBaquero, but doesn't WebTorrent Desktop use Flet's prettier-bytes which is on top of pretty-bytes? As it says in their [readme](https://github.com/sindresorhus/pretty-bytes), they use base-10.

DiegoRBaquero:
@cezar-carneiro You are correct! We are using Flet's `prettier-bytes` which is base-10. It's fine imho.

cezar-carneiro:
yeah base-10 works for me too

KingMaxer:
@DiegoRBaquero Windows 10 use base-2 by default, while WebTorrent Desktop windows edition use base-10, so the file sizes shown by these two are different.

DiegoRBaquero:
@KingMaxer WebTorrent core lib and Desktop app will continue to use base-10.

feross:
This is a good change! I just wish we didn't duplicate the calculation code that already appears in other parts of the app.

remi2j:
> This is a good change! I just wish we didn't duplicate the calculation code that already appears in other parts of the app.

Is this what's preventing from merging this PR? If so I can have a go at it

subins2000:
@remi2j If you're gonna have a go at it, can you also see if you can show the red line progress as mentioned in #1814

hicom150:
Given that the original PR got more than 3 years, what do you think if we merge it and work on refactoring the duplicated calculation code in another PR? 😅

hicom150 (approved):

feross:
@hicom150 Okay, fair :) Can you make an issue to track that we need to refactor the duplicated calculation code?

PR #1233, opened 2017-08-28
Use tabular nums and fixed width for no jitter on time

emilbayes:
The width of the video time was making me uncomfortable 😅, so I fixed by having the numbers be tabular and formatting the time with consideration for the total time. 

I was unable to run the integration tests on my 2016 12" Macbook.

I guess using tabular nums could be applied a lot of other places in the app, such as download percentage, file sizes etc.

Before Image:

Before

After image (Added 1h to duration for test purpose):

After

### DIFF
--- a/src/renderer/pages/player-page.js
+++ b/src/renderer/pages/player-page.js
@@ -652,9 +652,10 @@ function formatTime (time, total) {
   }
 
   let totalHours = Math.floor(total / 3600)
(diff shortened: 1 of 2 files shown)


emilbayes:
Do you want to merge this before it starts conflicting with other changes? 😅

feross:
This is awesome, sorry it took so long to merge and release 👍

PR #1237, opened 2017-09-09
Release media keys when opening external player

Goldob:
Until now, media keys didn't work when playing a file in VLC (at least on Linux), because they were already taken by WebTorrent. This PR fixes that.


cezar-carneiro:
Excellent!

PR #1239, opened 2017-09-11
feat(delete-all-torrents): added remove all torrents capability to the Transfers menu

SharonGrossman:
Hey, I've been wanting a Remove All Torrents capability for a while now, so I decided to implement it via the Transfers menu instead of some fast action button.

What do you think?


mfurlend:
I think this is very important. I made some scripts that I run from the dev console to do this, but buttons sure would be better.
https://gist.github.com/mfurlend/e131fe9daaecc59b65af671be8d216db

mathiasvr:
Just rebased on the latest master, and made some changes.
I think it's reasonable to have this feature.

moshmage:
? What's holding up this PR to get into master? this QoL feature is a must have

PR #1262, opened 2017-10-29
Clean up `handlers.js`

michaelgeorgeattard:


src/main/handlers.js
```diff
@@ -7,26 +7,24 @@ const config = require('../config')
...
-  }
-  if (process.platform === 'linux') {
-    installLinux()
+  switch (process.platform) {
+    case 'darwin': installDarwin()
+      break
+    case 'win32': installWin32()
+      break
+    case 'linux': installLinux()
+      break
```
feross: Not a fan of `switch` statements. I'll take a small set of `if`s any day. Switches are prone to misbehaving if you forget to `break`.

src/main/handlers.js
```diff
@@ -7,26 +7,24 @@ const config = require('../config')
...
-  }
-  if (process.platform === 'linux') {
-    uninstallLinux()
+  switch (process.platform) {
+    case 'darwin': uninstallDarwin()
+      break
+    case 'win32': uninstallWin32()
+      break
+    case 'linux': uninstallLinux()
+      break
```
feross: Same.

src/main/handlers.js
```diff
@@ -42,9 +40,9 @@ function installDarwin () {
   // File handlers are defined in `Info.plist`.
 }
 
-function uninstallDarwin () {}
+function uninstallDarwin () { }
```
feross: Not a fan of random whitespace changes.

src/main/handlers.js
```diff
@@ -42,9 +40,9 @@ function installDarwin () {
...
 
-const EXEC_COMMAND = [ process.execPath ]
+const EXEC_COMMAND = [process.execPath]
```
feross: Same.

src/main/handlers.js
```diff
@@ -108,37 +106,37 @@ function installWin32 () {
...
     }
 
     function setIcon (err) {
-      if (err) log.error(err.message)
+      if (err) return log.error(err.message)
 
       const iconKey = new Registry({
         hive: Registry.HKCU,
-        key: '\\Software\\Classes\\' + protocol + '\\DefaultIcon'
+        key: `\\Software\\Classes\\${protocol}\\DefaultIcon`
```
feross: Have you tested this change? We don't have a test for this, so it should be manually tested before merging.

src/main/handlers.js
Borewit: Not the most elegant syntax inherited from C, but it is intended for this scenario. From a logical point of view the elses are missing. Adding them with all the curly brackets is not a very pretty view neither. 

Lets get these pull requests out of the way and focus on important stuff.

src/main/handlers.js
Borewit: Irrelevant indeed

feross:
Thanks for the PR @michaelgeorgeattard! This has been superseded by https://github.com/webtorrent/webtorrent-desktop/pull/1406

feross:
Re-opening, see: https://github.com/webtorrent/webtorrent-desktop/pull/1406#issuecomment-391973517

michaelgeorgeattard:
Went through comments.

PR #1266, opened 2017-10-30
Cleaned up `external-player.js`

michaelgeorgeattard:


src/main/external-player.js
```diff
@@ -40,17 +40,17 @@ function spawn (playerPath, url, title) {
...
   if (!proc) return
-  log('Killing external player, pid ' + proc.pid)
+  log(`Killing external player, pid ${proc.pid}`)
   proc.kill('SIGKILL') // kill -9
   proc = null
 }
 
 function spawnExternal (playerPath, args) {
-  log('Running external media player:', playerPath + ' ' + args.join(' '))
+  log('Running external media player:', `${playerPath} ${args.join(' ')}`)
```
feross: Can you merge this into a single string?

src/main/external-player.js
```diff
@@ -71,7 +71,7 @@ function spawnExternal (playerPath, args) {
     proc = null
   })
 
-  proc.on('error', function (e) {
+  proc.on('error', e => {
```
feross: Can you change `e` to `err`, which is the convention that we use for errors?

feross:
Thanks for the PR @michaelgeorgeattard! This has been superseded by https://github.com/webtorrent/webtorrent-desktop/pull/1406

feross:
Re-opening, see: https://github.com/webtorrent/webtorrent-desktop/pull/1406#issuecomment-391973517

PR #1286, opened 2017-12-11
Remove 'video-on-top' from VLC player launch options

terryhau:
This command line option just sets the 'Always on top' setting in VLC.
The player will pop up in the foreground when playing a video from webtorrent without this option.

Having VLC always on top is very annoying when multi tasking.
If the user wants VLC always on top, they can set it in VLC preferences.

### DIFF
--- a/src/main/external-player.js
+++ b/src/main/external-player.js
@@ -29,7 +29,6 @@ function spawn (playerPath, url, title) {
     if (err) return windows.main.dispatch('externalPlayerNotFound')
     const args = [
       '--play-and-exit',
-      '--video-on-top',
       '--quiet',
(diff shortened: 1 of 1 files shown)


Ropiel:
I totally agree

With such a minimalist and handy software that's a pretty annoying option,please make that the "Floating on Top" option if unchecked works on vlc too.

pastudan:
+1 This is a simple change. Any word on if this can be merged?

feross:
We added this option because the VLC window does not come to the front without it. We don't want the window to stay in the front, but there was no other way to focus it without the option.

All told, my opinion on this has changed and I think it's more annoying to have the option enabled so let's try disabling it and see how many complaints that generates :)

PR #1299, opened 2018-01-01
NCU package.json upgrades

farribeiro:
Bump to lastest major releases ALL dependencies


farribeiro:
I think if isn't needed

PR #1309, opened 2018-01-20
Fix the desktop file

bilelmoussaoui:
If you run `desktop-file-validate` on the desktop file you will get this list of errors
```
 desktop-file-validate /usr/share/applications/webtorrent-desktop.desktop                                                         master    
/usr/share/applications/webtorrent-desktop.desktop: warning: key "Encoding" in group "Desktop Entry" is deprecated
/usr/share/applications/webtorrent-desktop.desktop: error: file contains key "Path" in group "Desktop Action CreateNewTorrent", but keys extending the format should start with "X-"
/usr/share/applications/webtorrent-desktop.desktop: error: file contains key "Path" in group "Desktop Action OpenTorrentFile", but keys extending the format should start with "X-"
/usr/share/applications/webtorrent-desktop.desktop: error: file contains key "Path" in group "Desktop Action OpenTorrentAddress", but keys extending the format should start with "X-"
```

This PR does:
- Remove the encoding key as it's not needed anymore
- Remove the Path key as it's not required, the group actions inherit the properties of their "parent"

### DIFF
--- a/static/linux/webtorrent-desktop.desktop
+++ b/static/linux/webtorrent-desktop.desktop
@@ -4,7 +4,6 @@ Version=1.0
 GenericName=BitTorrent Client
 X-GNOME-FullName=$APP_NAME
 Comment=Download and share files over BitTorrent
-Encoding=UTF-8
 Type=Application
 Icon=webtorrent-desktop
 Terminal=false
@@ -20,14 +19,11 @@ Actions=CreateNewTorrent;OpenTorrentFile;OpenTorrentAddress;
 [Desktop Action CreateNewTorrent]
 Name=Create New Torrent...
 Exec=$EXEC_PATH -n
-Path=$APP_PATH
 
 [Desktop Action OpenTorrentFile]
 Name=Open Torrent File...
 Exec=$EXEC_PATH -o
-Path=$APP_PATH
 
 [Desktop Action OpenTorrentAddress]
 Name=Open Torrent Address...
 Exec=$EXEC_PATH -u
-Path=$APP_PATH


bilelmoussaoui:
Any news? I would like to work more on the linux standards in order to improve the usability of this application :)

mathiasvr:
@bilelmoussaoui If you don't mind that it can take a long time to get a PR merged, you are welcome to work on this! 👍 
However, since there are fewer linux users, it can be more difficult to get feedback on your work.

bilelmoussaoui:
Yeah, totally got that. Except that the idea is to make the software easily available on Linux and working out of the box to bring more Linux users. 
This should have been merged as I have mentioned the latest specs, the tool used to verify the desktop file, it can be tested within a docker container in a few minutes :)

mathiasvr:
The time it takes to test it is not the issue, webtorrent members have to actually do it, and this repository doesn't currently have much activity.
Also, I think many wouldn't simply accept a pull request based on running a tool on an OS they don't use regularly, so they also have to spend time looking into how this desktop file and it's attributes work.

Whether this should have been merged or not is not up to you, and that's why you should only work on this if you don't mind the long wait time.

Borewit:
This would help me #1502 to review on Fedora ;-)

bilelmoussaoui:
Just a quick update on this: here's the Desktop file specs: https://standards.freedesktop.org/desktop-entry-spec/latest/

bilelmoussaoui:
@Borewit Thanks for merging this!

Borewit:
Thanks for your effort @bilelmoussaoui.

feross:
@bilelmoussaoui Sorry for the long delay on this PR. If you'd like to continue improving Linux support, please do so! We'll try to be better about reviewing nice, small PRs like this one in a more timely manner. Thanks!

bilelmoussaoui:
@feross Thanks for reaching. I just wanted to mention that WebTorrent is now packaged as a flatpak package (those new cool container based packaging formats) here https://flathub.org/apps/details/io.webtorrent.WebTorrent we are shipping a few files to make the app looks nice on every Linux software center upstream; would you mind having it here instead? https://github.com/flathub/io.webtorrent.WebTorrent/blob/master/io.webtorrent.WebTorrent.appdata.xml also I would give you write access if you want to automate the upgrade somehow :)

feross:
@bilelmoussaoui A few questions:

- How do you build the Flatpack file?
- Where is the flatpack file currently hosted? On Flathub? (We host all our release files in GitHub releases, so it would be best for us to host them there for consistency.)
- How does Flathub learn that there's a new version available? (Ideally, it would periodically look at GitHub releases to check for new versions)

bilelmoussaoui:
The flatpak is built from a manifest, either a JSON or YAML file, https://github.com/flathub/io.webtorrent.WebTorrent/blob/master/io.webtorrent.WebTorrent.json

You can build it locally if you have `flatpak-builder` using 
`flatpak-builder --install app ./io.webtorrent.WebTorrent.json --user`

Flatpak uses runtimes, which is basically a very simple container that contains the basics libraries to compile a new application. We have a BaseApp which is a bunch of shared modules used by a lot of apps but not enough to be a runtime. WebTorrent uses `io.atom.electron.BaseApp` 

Currently, the package only extract the deb file & install it content. See the webtorrent module here https://github.com/flathub/io.webtorrent.WebTorrent/blob/master/io.webtorrent.WebTorrent.json#L20

I can make it build from the source code instead. 

Regarding the updates, we can't "for now" look and auto-update packages just like that, as some other apps might not work with the latest runtime available, libraries... But you just need to update the URL pointing to the deb packages & sha256. You can either push directly to the master repository if you're sure that the build is working or open a PR & a bot will automatically build the package, provide you with a testing URL to try the app before merging & publishing your app see this MR for example https://github.com/flathub/com.spotify.Client/pull/74.

Flathub is about stable, reproducible builds (but it has a beta branch support too). So the builds are built in Flathub structure build it for the 4 supported architectures & thanks to Flatpak the users of whatever distro can get the latest update no matter which distro they are using, if it supports Flatpak.  

Don't hesitate to ping me for further questions.

Borewit:
This is what you are looking @feross: flathub/flathub#188

See also: _[Someone else has put my app on Flathub—what do I do?](https://github.com/flathub/flathub/wiki/App-Submission#someone-else-has-put-my-app-on-flathubwhat-do-i-do)_

feross:
@bilelmoussaoui Thanks for answering my questions. So if we move this manifest json file to this repo, like you're suggesting, will we be able to ship new versions to flathub users without sending a PR to the flathub repo?

bilelmoussaoui:
@feross You can't host the file here and have updates on Flathub without updating the file on Flathub repository. I can give you write access to whoever you want

feross:
@bilelmoussaoui I think it would be best if you guys and the community can handle keeping this file up to date since I don't have an easy way to test this.

bilelmoussaoui:
Sure, do you mind if I send a PR to add the appdata file that we have downstream? It's not a Flatpak specific thing, it's used to display the application info on almost any App Center on Linux. https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html#sect-Quickstart-DesktopApps

feross:
Sure! This should fix the problem where WebTorrent Desktop shows up as "Proprietary" software in the App Center, right? PR welcome!

bilelmoussaoui:
Yes! Perfect, I will prepare a PR for that whenever I can.

PR #1333, opened 2018-03-04
Add NPM script target 'fix' to fix standard code style problems

Borewit:


feross:
You can just use the new `npx` command that comes with node now to run this:

```bash
npx standard --fix
```

💪

Issue #1332, opened 2018-03-04
Try to select the front cover as the poster for audio/music torrents


Borewit:
Improve the poster selection for audio/music based torrent, aiming to get the cover-art selected as the torrent poster. 
 
The poster selection function has an existing _ToDo_, indicating some more work needs to be done for audio based torrents:

https://github.com/webtorrent/webtorrent-desktop/blob/474654fd163f1bf8298b280937453c04b097749b/src/renderer/lib/torrent-poster.js#L22

Downloading the torrent from [the audio album: Parov Stellar - The Princess](https://archive.org/details/ParovPrincess/Parov+Stelar+-+The+Princess/) currently results in the following:
![parovprincess weird poster without music-metadata](https://user-images.githubusercontent.com/23255260/36944841-3462ccd0-1fa4-11e8-80f8-c591aae5c269.png)

Borewit:
It does now select the correct poster in the index:
![parovprincess index with poster](https://user-images.githubusercontent.com/23255260/36945797-78f32c14-1fb3-11e8-8b3d-d57f6e925d20.png)

Playing the same track with this merge:
![parov steller - milla s dream - with poster](https://user-images.githubusercontent.com/23255260/36945839-1c710550-1fb4-11e8-943b-c3a974e9718c.png)

Putting the [music-metadata merge](https://github.com/webtorrent/webtorrent-desktop/pull/1240) (#1240) on top of that would give:
![parov steller - milla s dream - with poster and music-metadata merge](https://user-images.githubusercontent.com/23255260/36945905-f656e67c-1fb4-11e8-902e-750c5b75130f.png)

codealchemist:
Closing after #1334.

PR #1334, opened 2018-03-04
Improve the poster selection for audio/music based torrent

Borewit:
Addressing issue #1332


Borewit:
@codealchemist: Alberto, can you please review this one as well?

codealchemist:
I'm getting this error:
image

I was able to reproduce with MP3 files only so far.
Good thing is this should get fixed by #1361 :)

codealchemist:
Latest merge on this branch fixed thee issues I saw before, yay!

Properly displaying poster before all metadata was downloaded:
image

Properly displaying poster after metadata was downloaded:
image

Content list background:
image

Torrent list poster looking good:
image

codealchemist:
I was wondering if we should also display the same background on the torrent details view (files list).
Anyway, this looks really good!
Nice enhancement @Borewit!

codealchemist:
Hey @feross, if you agree I think we should merge this one too and then I'll prepare the changelog on another PR.

Borewit:
That was probably not a good idea to commit on top of this pull request after the merge.
I created a new push request: #1368,

PR #1240, opened 2017-09-17
Fix playing back audio files & reading music metadata

Borewit:
Replaced musicmetadata with [music-metadata](https://github.com/Borewit/music-metadata) module. The original one appears to be no longer maintained.

Fixes bug #1320, playing back audio files broken: caused by stream incompatibility between musicmetadata & readable-stream.

Should be faster, more stable is able to read more metadata and format information.
Supports additional audio formats and metadata tag headers.

Added audio format & some release information to the lay screen.

You could also use the library to retrieve album art from the audio files.

PR #1361, opened 2018-04-19
Displaying filename while music metadata is being downloaded.

codealchemist:
Related to #1240.
Thanks @Borewit for all the hard work there!
Can you help testing this one?

### DIFF
--- a/src/renderer/pages/player-page.js
+++ b/src/renderer/pages/player-page.js
@@ -207,7 +207,7 @@ function renderOverlay (state) {
 function renderAudioMetadata (state) {
   const fileSummary = state.getPlayingFileSummary()
   if (!fileSummary.audioInfo) return
-  const common = fileSummary.audioInfo.common
+  const common = fileSummary.audioInfo.common || {}
 
   // Get audio track info
   const title = common.title ? common.title : fileSummary.name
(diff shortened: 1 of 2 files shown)


codealchemist:
Player page while downloading metadata:
image

Player page when metadata is available:
image

Borewit:
Off topic:, I wrote a spectrum analyzer component which looks like this:
spectrum
Possibly we could add that to the bottom of the music play page, would that fit, would that be of any added value you think?
Please be critical.

codealchemist:
Yeah, I just looked at the source on #1334, neat stuff getting the `bestCover` ;)
If we can merge this fix I can later merge `master` onto that branch and retest it.
That would be a good enhancement.

On the other hand, about the _spectrum analyzer_, while I think it's something pretty cool it might not make too much sense to have it integrated into *WebTorrent*.
I don't think it adds value to the app.
If you download an album and start listening to a song you might not stare at the spectrum while you listen, at least not for too long :D

I'd love to add a plugin system into WebTorrent, so all this features would be possible without affecting the core.

Borewit:
Thanks for your feedback!

> If we can merge this fix I can later merge master onto that branch and retest it.

I merged all (master, this one (#1361) into the #1334 branch.

Issue #884, opened 2016-09-03
Delete the webtorrent-desktop npm package


dcposch:
It looks like we actually have people using `npm install webtorrent-desktop`.

Version 0.13.0 was never served from webtorrent.io - we went directly from 0.12.0 to 0.13.1. The the auto updaters also skipped that version. Despite that, we got telemetry reports from quite a few users using version 0.13.0.

That means they must have either downloaded the version directly from the Github Releases page, or used `npm install`. I bet it's the latter.

* Does the auto updater actually work correctly after `npm install`? The initial install is not done thru Squirrel, so probably not...
* Either way, we already have tons of different ways to install WebTorrent. Windows exe, Windows zip, Mac dmg, Mac zip, Linux deb (64- and 32-bit), Linux zip (64- and 32-bit). I don't think we want to support even more ways than that. Complexity makes it harder to be robust.

dcposch:
If the auto updater *doesn't* work for our npm users, then that helps explain why ~15% of our users are still on versions older than 0.12:

<img width="870" alt="screen shot 2016-09-03 at 2 19 05 pm" src="https://cloud.githubusercontent.com/assets/169280/18227673/77a3fffe-71e1-11e6-9c33-ef73374c6f65.png">

That 15% also includes:
* People who ran the app today for the first time in weeks
* People who used the Portable ZIP install option

mathiasvr:
I agree, I don't think GUI applications fits well with `npm install`.
On Mac I think something like [Cask](https://caskroom.github.io/) would be the better option.

dcposch:
@mathiasvr I don't want to support cask either. For Mac, we have a single supported way to install the app: using the DMG.

mathiasvr:
@dcposch I was just mentioning it as an alternative to npm if people really wanted to install through the terminal. Regarding supporting it developers rarely have a choice, and I just checked - it's already [there](https://github.com/caskroom/homebrew-cask/blob/master/Casks/webtorrent.rb).

demoneaux:
For now maybe you could just deprecate it first to give users time to adjust, before fully deleting it.

dcposch:
@demoneaux sounds good

feross:
@dcposch I don't have a strong opinion about this, so I'll defer to you.

But I don't think `webtorrent-desktop` is the main reason that people are still on pre-0.12 versions. Check out the latest stats:

[...]

PR #1364, opened 2018-04-25
README.md: add cask instructions

vitorgalvao:
This adds instructions on how to install with homebrew-cask. I did find [this comment](https://github.com/webtorrent/webtorrent-desktop/issues/884#issuecomment-244575027), but I’ll try to put your mind at ease regarding that.

1. The cask already exists, and has existed for over two years.
2. You don’t need to support it. Most casks are not supported by their maintainers. We have a vast number of contributors that keep them up to date.

If you have any questions regarding the process, feel free to ask (I’m a Homebrew-Cask maintainer). But rest assured **we do not require any effort on your part to keep the cask going**.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,11 @@
 
 Download the latest version of WebTorrent Desktop from
 [the official website](https://webtorrent.io/desktop/) or the
-[GitHub releases](https://github.com/webtorrent/webtorrent-desktop/releases) page.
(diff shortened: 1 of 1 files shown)


vitorgalvao:
> Will our auto-updater function if WebTorrent is installed with Homebrew-Cask?

Do you mean if it’ll continue to auto-update with your included method? Yes.

> Does Homebrew-Cask modify the binary in any way?

No. We download the app from your official URL, unpack it and move the app bundle to `/Applications` (or whatever the user decides). No alteration is done to any part of the app.

feross:
Perfect. Thanks!

Issue #1340, opened 2018-03-10
Improve the music metadata reading and updating mechanism


Borewit:
Improve the way metadata is read during the torrent life cycle, starting with a partial download to a completed file.

codealchemist:
Maybe we can close this after #1361?
What do you think @Borewit ?

Borewit:
I think there is still room for improvement. One the things we can do is use the embedded album art (if that is available in a suitable resolution) of the current played song.

PR #1368, opened 2018-04-26
Prevent spectrogram being selected as the poster.

Borewit:
Related to:
- Improve poster selection: webtorrent/webtorrent-desktop#1332,
- pull request webtorrent/webtorrent-desktop#1334

### DIFF
--- a/src/renderer/lib/torrent-poster.js
+++ b/src/renderer/lib/torrent-poster.js
@@ -97,7 +97,8 @@ function scoreAudioCoverFile (imgFile) {
     folder: 80,
     album: 80,
     front: 80,
-    back: 20
+    back: 20,
+    spectrogram: -80
   }
 
   for (let keyword in relevanceScore) {


Borewit:
Test torrent found on this page: [Alvvays-Alvvays2014FLAC](https://archive.org/details/Alvvays-Alvvays2014FLAC)


Before:
spectrum

After this change:
fixed-spectrum

feross:
Nice. Why does this need to be -80? Would 0 be sufficient?

Borewit:
0 points would have no effect.
Prior to this change the '_spectrometer.jpg_' would have got 0 points. 0 points are assigned to those images not matching any of the listed keyword / scores. So the assumption is that 0 point images are more likely to better choice for a poster then the spectrogram. Note that from these images, the largest will be selected.

The `0.8 *` however is pointless, I will remove that one (was part of an unfinished attempt to balance a size score with a keyword score).

Borewit:
Taking the album name from the metadata, and inject this as a dynamic keyword, will definitely improve the poster selection in a lot of cases.

Relates to the metadata update lifecycle #1340.

Borewit:
Thanks @ocheriaf !

PR #1411, opened 2018-05-25
If torrent is not private, leave private flag unset

feross:
If torrent is not private, leave private flag unset. This ensures that the torrent info hash will match the result generated by other tools, including `webtorrent-cli`.

Fixes: https://github.com/webtorrent/webtorrent-hybrid/issues/79

---

More info:

WebTorrent Desktop uses the result of the "Private" checkbox at torrent creation time to set the torrent's `private` flag to `true` or `false`. Unfortunately, there's an issue with the way this `private` flag spec was designed. There are actually 3 possible values for the `private` flag:

- `private: true`
- `private: false`
- not specified (defaults to `false`)

The problem is that `private` flag is part of the `info` section in the .torrent file, which means that each of these three values for `private` will produce a different info hash.

So, what's happening is that `webtorrent-cli` and `webtorrent-hybrid` are not setting `private` at all, while WebTorrent Desktop is setting it to `false`.

We should either set it to `true` or leave it not specified to match what `webtorrent-cli`, `webtorrent-hybrid`, and other mainstream torrent clients do.


feross:
More info:

WebTorrent Desktop uses the result of the "Private" checkbox at torrent creation time to set the torrent's `private` flag to `true` or `false`. Unfortunately, there's an issue with the way this `private` flag spec was designed. There are actually 3 possible values for the `private` flag:

- `private: true`
- `private: false`
- not specified (defaults to `false`)

The problem is that `private` flag is part of the `info` section in the .torrent file, which means that each of these three values for `private` will produce a different info hash.

So, what's happening is that `webtorrent-cli` and `webtorrent-hybrid` are not setting `private` at all, while WebTorrent Desktop is setting it to `false`.

We should either set it to `true` or leave it not specified to match what `webtorrent-cli`, `webtorrent-hybrid`, and other mainstream torrent clients do.

feross:
Thanks for the review!

Issue #1438, opened 2018-07-21
[Low-priority] Minutes section of hour-long times formatted without leading zero


BenMusch:
**What version of WebTorrent Desktop?**

0.19.0

**What operating system and version?**

Fedora 27

**What did you do?**

Opened a torrent longer than 1 hour, started watching from the beginning

**What did you expect to happen?**

The remaining time would be formatted as `2:08:33`

**What actually happened?**

Remaining time was formatted as `2:8:33`

![webtorrent-issue](https://user-images.githubusercontent.com/10344453/43031088-a8d890c8-8c4f-11e8-9b8b-fa57b449f58a.png)

BenMusch:
Don't know this codebase very well, but it seems like this issue is [here?](https://github.com/webtorrent/webtorrent-desktop/blob/master/src/renderer/pages/player-page.js#L722) I think it should be `totalMinutes <= 9`?

Borewit:
Looks like you nailed the problem.

BenMusch:
Cool, I'll open a PR

BenMusch:
@Borewit actually I think that wasn't the problem -- that if condition _should_ be checking if there are under 9 minutes total, so that a 9-minute time is displayed as `9:00` rather than `09:00`.

That probably is still the source of the bug, but I'd need to look things over again before fixing it

PR #1439, opened 2018-07-21
Fix zero-padding logic on minutes

BenMusch:
Fixes https://github.com/webtorrent/webtorrent-desktop/issues/1438

The if statement was testing the opposite of the case it was supposed to test. Changing `>` to `<=` would fix the issue, but there's no reason to even have the conditional to begin with, since `zeroPad(2, 10) === '10'`


BenMusch:
Hmm nevermind, I think I misunderstood the logic here. Will take a second look

Issue #1395, opened 2018-05-15
Add YouTube-style keyboard shortcuts


feross:
Let's consider adding the basic YouTube keyboard shortcuts in addition to the VLC ones that we already support. I don't think they conflict and they're familiar to lots of people.

Left/Right arrows - seek back/forward 5 seconds
Up/Down arrows - volume up/down 5%
<kbd>j</kbd> or <kbd>l</kbd> - seek back/forward 10 seconds
<kbd>k</kbd> - play/pause
<kbd>Shift</kbd> + <kbd>></kbd> - Increase speed
<kbd>Shift</kbd> + <kbd><</kbd> - Decrease speed

sergiorojasa:
I would like to work on this issue.

mathiasvr:
IMO we should not have multiple shortcuts for the same action, but I think it's fine to change the shortcuts for seeking and volume to single arrow presses.

PR #1443, opened 2018-07-30
Add YouTube style keyboard shortcuts

hicom150:
refs:  #1395

Left/Right arrows - seek back/forward 5 seconds
Up/Down arrows - volume up/down 5%
j or l - seek back/forward 10 seconds
k - play/pause
Shift + > - Increase speed
Shift + < - Decrease speed

### DIFF
--- a/src/renderer/main.js
+++ b/src/renderer/main.js
@@ -513,8 +513,8 @@ function onPaste (e) {
 }
 
 function onKeyup (e) {
-  const code = e.code;
-  const shiftKey = e.shiftKey;
+  const code = e.code
+  const shiftKey = e.shiftKey
 
   if (code === 'ArrowLeft') {
     dispatch('skip', -5)
--- a/src/renderer/main.js
+++ b/src/renderer/main.js
@@ -150,6 +150,8 @@ function onState (err, _state) {
   // ...same thing if you paste a torrent
   document.addEventListener('paste', onPaste)
 
+  document.addEventListener('keyup', onKeyup)
+
   const debouncedFullscreenToggle = debounce(function () {
     dispatch('toggleFullScreen')
   }, 1000, true)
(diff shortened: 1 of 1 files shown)


sunshinecool:
Hi! Is this going to get merged anytime soon?

Borewit:
So please get rid of https://github.com/webtorrent/webtorrent-desktop/pull/1443/commits/c2468502cdc765a11bef1da802b528c60a5ce1c0 & https://github.com/webtorrent/webtorrent-desktop/pull/1443/commits/fd937419151dd74d9860395637a4b22124227cfd in this PR.

hicom150:
@Borewit I think I have successfully removed the requested commits. Sorry I'm still a newbie :sweat_smile:

Borewit:
Thanks @sunshinecool, you doing just fine.

mathiasvr:
As I mentioned in the issue, I don't really like to have multiple shortcuts for the same actions.
I'm also not a fan of using two different key handlers (electron menu accelerators, and addEventListener).
I find it confusing that `Cmd+Alt+Right` skips 10s and a single `Right` skips 5s, and would they ever overlap?

I think we should just update the accelerators for seeking and volume control, and keep the rest.

Thoughts?

Borewit:
I agree with @mathiasvr for consistency in the code it would be better to have [one way to assign keys](https://github.com/webtorrent/webtorrent-desktop/blob/master/src/main/menu.js).

I have in principle no problem with multiple assignents mapped to the same action, as long there is a justification for the existence of each of them. To be compliant with YouTube layout is in my opinion a good justification, but it is indeed a fair question, do we still need the other one. 

If you move your shortcuts to the _accelerators_, please mark them with comment so it is clear they are part of the YouTube like layout key-layout.

Happy with that consession @mathiasvr?

To have all the shortcuts documented somehere would be awesome.

Just a small advice on your next PR, for mostly your own benefit, don't commit changes to your master branch. Keep that one reserver for incoming changes from the original master. Some more info on this kind of conventions can be found [here](https://nvie.com/posts/a-successful-git-branching-model/).

Will you create another PR with the work you did on the Linux integration tests?

mathiasvr:
@Borewit The problem is that the menu accelerators don't support multiple shortcuts, so if we want to support different shortcuts (at the same time), we probably have to do it this way.

I don't know how we are going to decide this though?

Borewit:
By being pragmatic.

I believe to support the YouTube shortcuts is desired, @feross the way I read the underlying issue. This issue has been marked as a _good first contribution_, someone took the challenge, I think we should approve or give guidence how he can succeed on his objective.

If it doesn’t fit in the menu accelerators, it makes sense to have the mechanism defined elsewhere. It should not conflict indeed, if so that needs to be addressed.

src/renderer/main.js
```diff
@@ -150,6 +150,8 @@ function onState (err, _state) {
   // ...same thing if you paste a torrent
   document.addEventListener('paste', onPaste)
 
+  document.addEventListener('keyup', onKeyup)
```
Borewit: Why keydown and not keyup?

mathiasvr:
I guess I will let @hicom150 decide then.

You can update the menu accelerators for seeking and volume control, I think that's the best solution.

Or if you think we should go with the current solution, I would request the following changes:
- use keydown instead of keyup, i think instant feedback and repeated events is an improvement.
- use window.addEventListener instead of document.addEventListener.
- use e.key instead e.code, currently this PR only works with qwerty keyboard layouts.

Borewit:
Superseded by #1395 due the fact the underlying repo holding the PR is missing.

PR #1449, opened 2018-08-13
webtorrent/webtorrent-desktop#1340: Switch to async metadata updates.

Borewit:
Implementation of asynchronous tag update events (send each time [music-metadata](https://github.com/Borewit/music-metadata) found a new tag while it is parsing the downloaded stream build up from the torrent chunks).

Solution for #1340: _Improve the music metadata reading and updating mechanism_

- [x] Add feature which allow async parser subscription: 
  - Issue: Borewit/music-metadata#125
  - Pull request: Borewit/music-metadata#126
  - Branch: [feature/async-updates](https://github.com/Borewit/music-metadata/tree/feature/async-updates)
- [ ] Review async concept
- [ ] Release version music-metadata with async observer API included
- [ ] Ready for merging

### DIFF
--- a/src/renderer/pages/player-page.js
+++ b/src/renderer/pages/player-page.js
@@ -216,9 +216,7 @@ function renderAudioMetadata (state) {
   const elems = []
 
   // Audio metadata: artist(s)
-  const artist = common.albumartist || common.artist ||
-    (common.artists && common.artists.filter(function (a) { return a }).join(', ')) ||
-    '(Unknown Artist)'
+  const artist = common.artist || common.albumartist
   if (artist) {
     elems.push((
       <div key='artist' className='audio-artist'>
--- a/src/renderer/webtorrent.js
+++ b/src/renderer/webtorrent.js
@@ -345,7 +345,13 @@ function getAudioMetadata (infoHash, index) {
   const metadata = { title: file.name }
   ipc.send('wt-audio-metadata', infoHash, index, metadata)
 
-  const options = {native: false, skipCovers: true, fileSize: file.length}
+  const options = {native: false,
+    skipCovers: true,
+    fileSize: file.length,
+    observer: (event) => {
+      console.log(`async-audio-metadata-update: file='${file.name}', type=${event.tag.type}, tag-id=${event.tag.id}`)
+      ipc.send('wt-audio-metadata', infoHash, index, event.metadata)
+    }}
   const onMetaData = file.done
     // If completed; use direct file access
     ? mm.parseFile(path.join(torrent.path, file.path), options)
@@ -353,9 +359,8 @@ function getAudioMetadata (infoHash, index) {
     : mm.parseStream(file.createReadStream(), file.name, options)
 
   onMetaData
-    .then(function (metadata) {
-      console.log('got audio metadata for %s (length=%s): %o', file.name, file.length, metadata)
-      ipc.send('wt-audio-metadata', infoHash, index, metadata)
+    .then(function () {
+      console.log(`metadata for file='${file.name}' completed.`)
     }).catch(function (err) {
       return console.log('error getting audio metadata for ' + infoHash + ':' + index, err)
     })


Borewit:
@codealchemist Can you please review these changes?

Note that the dependency of music-metadata is not updated yet.
Since it a fairly large change to music-metadata, I would like have it reviewed prior to merging it. 

At this time you need to checkout [feature/async-updates branch of music-metadata](https://github.com/Borewit/music-metadata/tree/feature/async-updates), which is still a pending [pull request](https://github.com/Borewit/music-metadata/pull/126).

You can yarn to [link](https://yarnpkg.com/lang/en/docs/cli/link/) to your local music-metadata branch.

Borewit:
Partial music-metadata update:
async-update 01

Few seconds later:
async-update 02

codealchemist:
Oh, I can't play any files on first try.
But it doesn't throw any errors; strange.
Tomorrow I'll try again fresh, I'm too sleepy now.

Just wanted to say this is awesome stuff!
And I'm looking forward to testing it.
:)

codealchemist:
BTW, I installed `music-metadata` from your branch like this:
`npm i Borewit/music-metadata#feature/async-updates`

Borewit:
@codealchemist:

I am looking forward to your feedback!

> `npm i Borewit/music-metadata#feature/async-updates`

I guess that is not working, because the TypeScript in music-metadata is not getting compiled.

I use [yarn, to link](https://yarnpkg.com/lang/en/docs/cli/link/) the local checked out dependency. 

```bash
git clone https://github.com/Borewit/music-metadata.git
cd music-metadata
git checkout feature/async-updates
yarn install
yarn build
yarn link
cd ../webtorrent-desktop
yarn link music-metadata
```

src/renderer/webtorrent.js
```diff
@@ -345,17 +345,22 @@ function getAudioMetadata (infoHash, index) {
   const metadata = { title: file.name }
   ipc.send('wt-audio-metadata', infoHash, index, metadata)
 
-  const options = {native: false, skipCovers: true, fileSize: file.length}
+  const options = {native: false,
+    skipCovers: true,
+    fileSize: file.length,
+    observer: event => {
+      console.log(`async-audio-metadata-update: file='${file.name}', type=${event.tag.type}, tag-id=${event.tag.id}`)
```
codealchemist: Maybe we can remove this extra logging before merging.
:)

src/renderer/pages/player-page.js
```diff
@@ -216,9 +216,7 @@ function renderAudioMetadata (state) {
   const elems = []
 
   // Audio metadata: artist(s)
-  const artist = common.albumartist || common.artist ||
-    (common.artists && common.artists.filter(function (a) { return a }).join(', ')) ||
```
codealchemist: `common.artists` array is not available anymore?

Borewit: It is still available, but common.artist is already designed to summarize the artist(s) in a single string.
Either derived from common.artists or directly from the metadata.

In other words, the code removed is already done in music-metadata.

Borewit:
> I saw 13 async calls which initially looked like too many, but we might not be using all the data we get.

There maybe indeed calls now which do not effect any visual change because the metadata triggering the call is not used. I thought about this, and assumed these callback were not to expensive. But the source of all evil in the world are assumptions ;-). To keep things simple I decided not to filter the events, but that can be done if you think it is worth it.

codealchemist:
About having extra async calls I think it's ok.
If we want to do further optimizations in the future maybe we can tell `music-metadata` the fields we're interested about to only do the callback for those.
This is really good already!

codealchemist:
Sample OGG file properly getting async metadata:
image

src/renderer/lib/migrations.js
```diff
@@ -29,10 +29,23 @@ function run (state) {
...
+    installHandlers(state.saved)
+  }
+
   // Config is now on the new version
   state.saved.version = config.APP_VERSION
 }
 
+// Whenever the app is updated,  re-install default handlers if the user has
+// enabled them.
+function installHandlers (saved) {
```
codealchemist: Maybe this enhancement is worth its own PR.
But if the other folks agree we can add it here.
Really nice catch BTW :)

Borewit:
@mathiasvr this one should be good for merging as well.

mathiasvr:
Maybe it is not relevant to this PR, but doesn't the metadata streaming force webtorrent to download the file sequentially, even if the user wants to seek to a position later in the file?

Anyway, I think we can merge this and consider potential optimizations later.

A couple of things:
- Could you remove the commit from #997, I think it should be handled separately.
- We could maybe update music-metadata to the latest version?

Borewit:
> Maybe it is not relevant to this PR, but doesn't the metadata streaming force webtorrent to download the file sequentially, even if the user wants to seek to a position later in the file?

Not sure if the metadata is also driving the sequential download. But if metadata doesn't trigger it, the audio player will do for sure will, despite the fact that you want to start playing half way for multiple reasons:
1. The header information is stored at the beginning of the file.
2. The audio player will read the metadata chunks as well
3. The Node stream does not provide a mechanism to jump in the stream

Although it is depending on the file format and provided settings, music-metadata doesn't read necessary to the end of the stream, in many cases, a relative small portion of the beginning of the stream is sufficient.

I have the impression the sequential download is already prioritized at the moment you navigate to the audio track "page".

For sure this PR doesn't change the way the audio metadata is read, that's the same is as before. It starts to display metadata during parsing.

But if you see a way how we can do things smarter, I will look into it.

> Could you remove the commit from #997, I think it should be handled separately.

Regarding that commit, for sure I did not inject that on purpose. 
I checked my local branch and I cannot find that commit. I deleted my local branch, checked it out again, it is not there. Neither I can find it in the [commit history in the git repo](https://github.com/Borewit/webtorrent-desktop/commits/feature/async-update-music-metadata).

Neither I see it ending up the changes, so I suppose is not part of the PR.

Maybe I made mistake at some point and falsified the history of the branch. So sorry.

> We could maybe update music-metadata to the latest version?

Sounds good to me.

mathiasvr:
> Although it is depending on the file format and provided settings, music-metadata doesn't read necessary to the end of the stream, in many cases, a relative small portion of the beginning of the stream is sufficient.

Sounds good, in that case I do not see a problem. 👍 

>I checked my local branch and I cannot find that commit. I deleted my local branch, checked it out again, it is not there. Neither I can find it in the commit history on the git repo.

Yes that looks pretty weird? However I still see it in GitHub changes view, and also when manually pulling your branch to the master.

Can you try to rebase your branch? Otherwise if you check **Allow edits from maintainers** on this PR, I can try to do it.

Also, is it intentional that this PR removes this part from line 241 in `player-page.js`?:
```js
|| (common.artists && common.artists.filter(function (a) { return a }).join(', ')) || '(Unknown Artist)'
```

Borewit:
Oh I see what you mean, it part of the change-set, overlooked that.

Allow edits from maintainers is and was granted Mathias, be my guest.

Borewit:
Thanks Mathias!

PR #1454, opened 2018-08-15
Fix audio track number not displayed and adds audio disk (media) number

Borewit:
- Fixes audio track number not displayed of the total number of is not defined
- Adds the disk (media) number to audio track properties
- Changes the order of audio track properties

### DIFF
--- a/src/renderer/pages/player-page.js
+++ b/src/renderer/pages/player-page.js
@@ -245,6 +245,15 @@ function renderAudioMetadata (state) {
     ))
   }
 
+  // Audio metadata: disk & track-number
+  const count = ['track', 'disk']
+  count.forEach(key => {
+    const nrElem = renderTrack(common, key, key[0].toUpperCase() + key.substring(1))
+    if (nrElem) {
+      elems.push(nrElem)
+    }
+  })
+
   // Audio metadata: album
   if (common.album) {
     elems.push((
@@ -286,15 +295,6 @@ function renderAudioMetadata (state) {
       </div>
     ))
   }
-
-  // Audio metadata: disk & track-number
-  const count = ['track', 'disk']
-  count.forEach(key => {
-    const nrElem = renderTrack(common, key, key[0].toUpperCase() + key.substring(1))
-    if (nrElem) {
-      elems.push(nrElem)
-    }
-  })
 
   // Audio metadata: format
   const format = []
--- a/src/renderer/pages/player-page.js
+++ b/src/renderer/pages/player-page.js
@@ -288,7 +288,7 @@ function renderAudioMetadata (state) {
   }
 
   // Audio metadata: disk & track-number
-  const count = ['track']
+  const count = ['track', 'disk']
   count.forEach(key => {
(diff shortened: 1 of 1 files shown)


Borewit:
Disk number, total nr of disks, track nr & total tracks displayed:
track disk nr - parov stelar - the princess

In this track, only the track number is present:
track disk nr - robin grey - these days

src/renderer/pages/player-page.js
```diff
@@ -204,6 +204,24 @@ function renderOverlay (state) {
...
+ * @param label should be either 'Track' or 'Disk'
+ */
+function renderTrack (common, key, label) {
+  // Audio metadata: track-number
+  if (common[key].no) {
+    let str = `${common[key].no}`
+    if (common[key].of) {
+      str += ` of ${common[key].of}`
+    }
+    return React.createElement('div', { key, className: 'audio-' + key },
```
codealchemist: I'd suggest returning a JSX block, like the one we had before.

src/renderer/pages/player-page.js
```diff
@@ -227,6 +245,15 @@ function renderAudioMetadata (state) {
     ))
   }
 
+  // Audio metadata: disk & track-number
+  const count = ['track', 'disk']
+  count.forEach(key => {
+    const nrElem = renderTrack(common, key, key[0].toUpperCase() + key.substring(1))
```
codealchemist: Maybe we can use CSS to capitalize the label.
If we do that we would only pass `key` and avoid operating directly on the string.

On `renderTrack` we would be able to do:
```
    const style = { textTransform: 'capitalize' }
    return (
      <div className={`audio-${key}`}>
        <label style={style}>{key}</label> {str}
      </div>
    )
```

Borewit: I have no experience with React, I blindly followed your suggestion.

src/renderer/pages/player-page.js
```diff
@@ -227,6 +248,15 @@ function renderAudioMetadata (state) {
...
```
codealchemist: Here we can remove the third param, which is not needed anymore.
:)

Borewit: Oops, yeah sure.

codealchemist:
Found this error when `common[key]` is undefined on `renderTrack`:
image

Borewit:
Good catch, but strange, common.track: & common.disk should always exist...

````TypeScript
  public readonly common: ICommonTagsResult = {
    track: {no: null, of: null},
    disk: {no: null, of: null}
  };
````
[Ref](https://github.com/Borewit/music-metadata/blob/a207f498e89445326fc8dac1df2e525c6c124479/src/common/MetadataCollector.ts#L55-L58)

codealchemist:
On [player-page:232](https://github.com/webtorrent/webtorrent-desktop/blob/8e9184084dc1ffbec45b273b71a92816018bbb00/src/renderer/pages/player-page.js#L232) we default `common` to `{}` when falsy.
The first two calls to `renderTrack` were made with `common` not being set yet:
image

PR #1483, opened 2018-09-26
Update to material-ui 0.20.2

mathiasvr:
- removes `react-tap-event-plugin`, as it is not needed anymore.


src/renderer/main.js
```diff
@@ -39,9 +39,6 @@ const TorrentPlayer = require('./lib/torrent-player')
 // Perf optimization: Needed immediately, so do not lazy load it below
 const TorrentListController = require('./controllers/torrent-list-controller')
 
-// Required by Material UI -- adds `onTouchTap` event
-require('react-tap-event-plugin')()
```
Borewit: Module is deprecated ([react-tap-event-plugin](https://github.com/zilverline/react-tap-event-plugin)), removed as a dependency as well :heavy_check_mark:

mathiasvr:
Thanks for reviewing @Borewit! 😃

mathiasvr:
Thanks a lot @DiegoRBaquero! 😃

PR #1485, opened 2018-09-27
Update auto-launch

mathiasvr:
Removed Mac work around, as it is not needed anymore.
Tested on macOS 10.11.6.


mathiasvr:
I've just made the package-lock update less invasive (easier to merge), and synced the package.json version numbers.

This only updates dependencies within semver, which was also done when we introduced the package-lock, but this was lost in an earlier PR.

I also updated to the latest electron v1, because I haven't been able to run webtorrent desktop without it.

I know it is wierd of me to sneek all this stuff into the merge, but since we don't release that often I hope we can quickly fix any issues this may cause.

Also, I have tested webtorrent desktop in a similar state many times the past months, and honestly this is the only oportunity I've had to merge stuff in all that time 😉

PR #1498, opened 2018-10-10
Cleanup preferences page

mathiasvr:
- Moves path logic to PathSelector component
- Removes `displayValue`, and always shows path instead.

I think it makes sense to remove `displayValue`, because it's only used to display the name of the external player, which is inconsistent with how we display other paths.
Also, the player name is already displayed in a message above the path, so we don't lose that information.

### DIFF
--- a/src/renderer/components/path-selector.js
+++ b/src/renderer/components/path-selector.js
@@ -1,3 +1,5 @@
+const path = require('path')
+
 const colors = require('material-ui/styles/colors')
 const electron = require('electron')
 const React = require('react')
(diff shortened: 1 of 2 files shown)


mathiasvr:
Let's try this again 😏 
It's pretty low priority though, so don't waste too much time on it.

hicom150:
LGTM! 👍

If this PR gets merged, #1652 and #1653 could be closed 😉

feross:
@hicom150 Thanks for letting me know. I'll close those now :)

Thanks, @mathiasvr!

PR #1499, opened 2018-10-10
Add button to re-download torrent data if missing

mathiasvr:
Allows to re-download torrent data if it has been deleted.
Otherwise you would have to remove the torrent and add it again.

Not sure if the message or style is quite right?

### DIFF
--- a/src/renderer/pages/torrent-list-page.js
+++ b/src/renderer/pages/torrent-list-page.js
@@ -6,7 +6,7 @@ const LinearProgress = require('material-ui/LinearProgress').default
 
 const TorrentSummary = require('../lib/torrent-summary')
 const TorrentPlayer = require('../lib/torrent-player')
-const { dispatcher } = require('../lib/dispatcher')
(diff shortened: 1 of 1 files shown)


Borewit:
Should we not try to detect the state in which data does not correspond with the torrent.
I am starting shivering after thinking back to the occasians discovering files with partial 0 filled blocks most likely caused by redistributing partial downloads.

mathiasvr:
I think webtorrent checks this when starting a torrent?

src/renderer/pages/torrent-list-page.js
```diff
@@ -394,10 +394,20 @@ function getErrorMessage (torrentSummary) {
   if (err === 'path-missing') {
     return (
       <span>
-        Path missing.<br />
-        Fix and restart the app, or delete the torrent.
+        Torrent data missing.<br />
+        Fix or click to
+        <a onClick={(e) => {
```
Muhnad: it should be `button` not `a`

Issue #1500, opened 2018-10-12
Select All / Delete All torrents from list


fknop:
**What version of WebTorrent Desktop?** (See the 'About WebTorrent' menu)

0.20.0

**What operating system and version?**

Windows 7

**What did you do?**

Select All / Delete button from Edit menu do nothing.

**What did you expect to happen?**

I would like them to select and batch delete my torrents.

I would be glad to create a pull request if this is something the application needs. I always feel the need to batch delete my torrents instead of clicking and deleting them one by one.

mathiasvr:
Does #1239 do what you want?

fknop:
For the delete part it seems good.

Is the select all to pause or restart everything already implemented ?

PR #1512, opened 2018-11-11
Cleanup edit menu

fknop:
Related to #1500 and #1239.

I don't think those edit submenus serve any purpose. I believe it would be better to remove them.

### DIFF
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -121,29 +121,8 @@ function getMenuTemplate () {
       label: 'Edit',
       submenu: [
         {
-          role: 'undo'
-        },
-        {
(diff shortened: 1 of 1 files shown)


mathiasvr:
At least on mac, these menu items are standard for all applications, and they do work for text fields in webtorrent desktop, such as adding a torrent address.
However other applications are better at disabling (grey out) these items when they do not serve any purpose, so it's worth looking into how to do this.

fknop:
What use case do you have in mind for this particular application? The only one that currently works is pasting a torrent address on the main window. Greying them out is a good idea if they are indeed standard.

PR #1514, opened 2018-11-20
Greenkeeper/music metadata 3.3.0

Borewit:
1. Update music-metadata to 3.3.0
2. Update media audio exentsions:
   - Add aif/aiff, spx (Speex) & mpc (Musepack) as audio file extensions.
   - Remove clashing .mp4 extension (with video) from audio.


Borewit:
Thanks @mathiasvr

Borewit:
**Don't merge yet**, I would like to further address this nasty bug first: Borewit/music-metadata#174

DiegoRBaquero:
Any update @Borewit ?

Borewit:
Let me start with the background.

In certain cases, when music-metadata is synchronization on a byte sequence in the audio track (stream), to many async calls are handled in a recursive manner causing Node to run out memory, become slow or entirely crash.

To prevent that I implemented a kind of read-ahead buffer, which reduced the number of read operations when _byte hunting_ in an MPEG stream. This seemed to work fine, until issue Borewit/music-metadata#174 pointed out a similar bug was back in. Crashing node. Caused by to many recursions.

While investigate the error, I added a catch error some where in promise handling chain and the error was gone. I will repeat that, I slightly changed the code structure to   
Does this make any sense? Not to me. The slightest change anywhere in the code could apparently fix or respectively break music-metadata.

This little catch, is what I released as a workaround in [v3.2.3](https://github.com/Borewit/music-metadata/releases/tag/v3.2.3). I apologize for this, it is not my style of fixing issues. But it did the job so far, no new issues reported since then.

So what is the good way of addressing this issue?

Doing the "byte hunting" with synchronous reads? That would require drastic changes in [music-metadata](https://github.com/borewit/music-metadata), and the underlying modules including changes in the API's. In addition to that, I don't think that it is how JavaScript is intended to work. 

I went for another option, I started replacing the _then_ / recursive structures with "new" async/await which allows recursions to be replaced with loops. That is not efficient enough to get rid of the ahead buffers, but it seems to address the restriction on iteration.

Although I did most of the work already, I need to find some time to finalize it

DiegoRBaquero:
It's better that we start moving everything to await/async, yeah, that works.

I've removed the package-lock. It should be generated after merging or when building.

lmk when you are done.

mathiasvr:
Please consider keeping the package-lock. It makes installs and builds reproducible, and I think it is a huge improvement.

Any conflicts can be solved by running `npm install --package-lock-only`, or installing npm-merge-driver.

Deleting and remaking it usually causes uneccesarily large diffs.

DiegoRBaquero:
Ooops, didn't mean to delete it, meant to remove it from the PR.

Borewit:
I have updated the following:
1. music-metadata updated to [v3.3.0](https://github.com/Borewit/music-metadata/releases/tag/v3.3.0)
2. Restored NPM lock file
3. Fixed merge conflicts with master

@mathiasvr, can you please approve?

PR #1520, opened 2018-12-10
Compact view for torrent list

fknop:
Hi,

I implemented this small feature for a compacted list of torrents (reduce ~by half the space needed for one item in the list) via a button icon in the header.

I know this might not be a feature necessary to the core application so I'm adding this here and leaving it to discussion!

Tests / formatting may fail for now, I did not check those yet.


dsernst:
Can you post before & after screenshots?

feross:
Ping – can someone post screenshots of this?

In general, I'm against adding options for stuff like this. If the compact view is "better" then we should just change to using the compact view for everyone. In general, I agree that WebTorrent current takes up a bit too much space for each torrent list item, so this is probably a good PR to merge, minus the configuration button.

alxhotel:
Before and after:

wt_compare_compact

feross:
@alxhotel Thanks for posting that. I like the general idea.

Issues:
- The buttons on the right side or not properly vertically centered.
- Does the torrent meta-info show up in two lines when the torrent is expanded? Would be more readable than keeping it on one line, in that situation.

fknop:
I've managed to fix the vertical alignment locally, I think it should be well aligned now.

image

Right now the metadata is kept on one line when the torrent is expanded. I can try to fix this.

fknop:
The behaviour is now the same as before when the torrent is expanded.

feross:
@fknop Thanks for making those changes. It appears that there is a merge conflict. Can you try to resolve that?

fknop:
It should be working now. The app still has the two options, if you want to keep only one view I can probably remove the old one as well.
Or maybe adding a preference would be better than having a button on the main view?

PR #1532, opened 2019-01-02
Add preference option to sort files by name, size, or default

Arryan:
I know it has been requested numerous times to sort files alphanumerically because it can be very frustrating to stream a season of a show and you have to look very hard to find the episode you need. In this PR, I allow users to have the option to sort by file name or file size. The files are not sorted by default so it should not change the user experience unless they seek the option.

In response to [this](https://github.com/webtorrent/webtorrent-desktop/pull/1188#issuecomment-315634680): 
>Thank you for your PR. We want to keep a simple UI, with not a lot of configuration. It would be cool to offer the possibility to extend config with extensions.

This doesn't add any other dependencies and keeps the main UI completely the same.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1639,12 +1639,14 @@
             "balanced-match": {
               "version": "1.0.0",
               "resolved": false,
-              "integrity": "sha1-ibTRmasr7kneFk6gK4nORi1xt2c="
+              "integrity": "sha1-ibTRmasr7kneFk6gK4nORi1xt2c=",
+              "optional": true
             },
             "brace-expansion": {
               "version": "1.1.11",
               "resolved": false,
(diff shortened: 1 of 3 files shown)


mathiasvr:
@Arryan This is not targeted at you, but more as a general discussion of WebTorrent Desktop, which I think is needed before we can decide to merge this.

One of the main principles of WebTorrent Desktop is simplicity and not to offer a preference for every trivial option.

That being said, I'm not against a little customization and I think issues being raised often enough should probably be optional. I suppose I'm naive enough to believe this won't blow up the preference page.

Also, during beta we could consider measuring how users use each preference to decide which to keep for v1.0, as a way to figure out the right thing to do, when it isn't easy to decide.

Arryan:
Perhaps we could make a drop down for advanced options that would keep the preferences page clean while also giving more power to the user. I think your idea for the beta sounds good @mathiasvr👍

Issue #1067, opened 2016-11-07
Remove all sounds but the startup sound


yuriiz:
There is no way to turn off annoying sound notifications when interacting with application (adding torrent, removing torrent etc).

**What version of WebTorrent Desktop?** (See the 'About WebTorrent' menu)
Version 0.17.2 (0.97.2) (64-bit)

**What operating system and version?**
Xubuntu 16.04

**What did you do?**
Opened Edit->Preferences

**What did you expect to happen?**
Too see an option to turn off super-annoying blip-blips on every action.

**What actually happened?**
Didn't found it.

yuriiz:
Workaround under linux is to mute "Chromium" application under PulseAudio Volume Control but this mutes media player too and it is becoming usable only with external media player.

kaotisk-hund:
I started this [PR](https://github.com/feross/webtorrent-desktop/pull/1068) but I can't complete it.

feross:
Thanks for the issue, but I don't want a preference for every trivial option that a user could want. I think we should just do the right thing, which is probably to remove the sounds entirely it seems.

feross:
Re-opening to use this as the tracking issue.

mifi:
I think the sounds are cute

feross:
@mifi Me too, but they're pretty controversial

rjmill:
@feross I agree with you on not having preferences for trivial things, but I think a checkbox for muting the sounds would be completely appropriate.

I'd toggle the sounds all day if I could. I don't like them most of the time, but they're really convenient when I am actively waiting for a torrent to download. The fact that the sounds are controversial and polarizing is a good sign that each user should be able to decide for themselves what to do, IMO.

PR #1536, opened 2019-01-12
Toggle the sound notifications

adriantombu:
This PR is related to the #1067 issue

As the notificaiton sounds seem a bit controversial, I guess the best way to content everybody would be to allow people to chose what they want, hence this little checkbox to toggle the sounds 😃 

Personnally, I'm like @rjmill : sometime I need absolute silence, and sometime I need the notifications.

Anyways, I'm open for feedback 👍

### DIFF
--- a/src/renderer/pages/preferences-page.js
+++ b/src/renderer/pages/preferences-page.js
@@ -26,6 +26,9 @@ class PreferencesPage extends React.Component {
 
     this.handleStartupChange =
       this.handleStartupChange.bind(this)
+
+    this.handlesoundNotificationsChange =
+      this.handlesoundNotificationsChange.bind(this)
   }
 
   downloadPathSelector () {
@@ -205,6 +208,22 @@ class PreferencesPage extends React.Component {
     )
   }
 
+  soundNotificationsCheckbox () {
+    return (
+      <Preference>
+        <Checkbox
+          className='control'
+          checked={this.props.state.saved.prefs.soundNotifications}
+          label={'Activate the notification sounds'}
+          onCheck={this.handlesoundNotificationsChange} />
+      </Preference>
+    )
+  }
+
+  handlesoundNotificationsChange (e, isChecked) {
+    dispatch('updatePreferences', 'soundNotifications', isChecked)
+  }
+
   handleSetDefaultApp () {
     dispatch('updatePreferences', 'isFileHandler', true)
   }
@@ -231,6 +250,9 @@ class PreferencesPage extends React.Component {
           {this.setDefaultAppButton()}
         </PreferencesSection>
         {this.setStartupSection()}
+        <PreferencesSection title='Miscellaneous'>
+          {this.soundNotificationsCheckbox()}
+        </PreferencesSection>
       </div>
     )
   }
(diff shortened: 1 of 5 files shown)


src/renderer/pages/preferences-page.js
```diff
@@ -205,6 +208,22 @@ class PreferencesPage extends React.Component {
     )
   }
 
+  soundNotificationsCheckbox () {
+    return (
+      <Preference>
+        <Checkbox
+          className='control'
+          checked={this.props.state.saved.prefs.soundNotifications}
+          label={'Activate the notification sounds'}
```
mathiasvr: Nitpick: I would prefer 'Enable notification sounds'.

But since it is both interaction sounds and notification sounds, maybe it should simply be 'Enable sounds'?

src/renderer/pages/preferences-page.js
```diff
@@ -231,6 +250,9 @@ class PreferencesPage extends React.Component {
           {this.setDefaultAppButton()}
         </PreferencesSection>
         {this.setStartupSection()}
+        <PreferencesSection title='Miscellaneous'>
```
mathiasvr: I think we should group the startup and sound options under a 'General' section instead.

adriantombu: You're right, looks cleaner to me too 👍

mathiasvr:
@adriantombu Thanks for the changes! I took the liberty to add some migration handling to your branch as well.

Based on the previous feedback I also think it's fine to make this an option.

trujamal:
Code checks out, everything looks good, and works as intended with the latest build. Checklist noted below:

## General

- [x] Does the code work? 
- [x] Is all the code easily understood?
- [x] Does it conform to your agreed coding conventions?
- [ ] Is there any redundant or duplicate code?
- [x] Is the code as modular as possible?
- [ ] Can any global variables be replaced?
- [ ] Is there any commented out code?
- [x] Do loops have a set length and correct termination conditions?
- [x] Do the names used in the program convey intent?

## Performance
- [x] Are there any obvious optimizations that will improve performance?
- [x] Can any of the code be replaced with library or built-in functions?
- [x] Can any logging or debugging code be removed?

## Security
- [x] Are all data inputs checked (for the correct type, length, format, and range) and encoded?
- [x] Where third-party utilities are used, are returning errors being caught?
- [x] Are output values checked and encoded?
- [x] Are invalid parameter values handled?

## Documentation
- [x] Do comments exist and describe the intent of the code?
- [x] Are all functions commented?
- [x] Is any unusual behavior or edge-case handling described?
- [x] Is the use and function of third-party libraries documented?
- [x] Are data structures and units of measurement explained?
- [x] Is there any incomplete code? If so, should it be removed or flagged with a suitable marker like ‘TODO’?

## Testing
- [x] Is the code testable? The code should be structured so that it doesn’t add too many or hide dependencies, is unable to initialize objects, test frameworks can use methods etc.
- [x] Do tests exist, and are they comprehensive?
- [x] Do unit tests actually test that the code is performing the intended functionality?
- [x] Could any test code be replaced with the use of an existing API?

PR #1540, opened 2019-01-30
Recommended readings to start working in the app.

RecoX:
As there is no too much information in the readme.md in how is this made, I think put a link of Electron and React to have an idea of the technologies used here is a good starter.


RecoX:
After come back to the developing 3 months ago i re-need this https://material-ui.com/getting-started

so i recommend to merge it.

feross:
Sure, thanks for the PR :)

PR #1579, opened 2019-05-05
Add YouTube style hotkeys. Closes #1395

### LINKED ISSUE #1395: Add YouTube-style keyboard shortcuts

dsernst:
This is a continuation of PR https://github.com/webtorrent/webtorrent-desktop/pull/1443, which appears abandoned. 

Props to @hicom150 for starting that. I would have added commits on top of that, to preserve git history etc, but the underlying branch looks deleted 😕

This PR takes into account the code changes requested by @mathiasvr in https://github.com/webtorrent/webtorrent-desktop/pull/1443#issuecomment-429120108

> I would request the following changes:
> 
> * use keydown instead of keyup, i think instant feedback and repeated events is an improvement.
> * use window.addEventListener instead of document.addEventListener.
> * use e.key instead e.code, currently this PR only works with qwerty keyboard layouts.

All of these have been included. 

----

This closes the original feature request by @feross in https://github.com/webtorrent/webtorrent-desktop/issues/1395.

### DIFF
--- a/src/renderer/main.js
+++ b/src/renderer/main.js
@@ -147,6 +147,9 @@ function onState (err, _state) {
   // ...same thing if you paste a torrent
   document.addEventListener('paste', onPaste)
 
+  // Add YouTube style hotkey shortcuts
+  window.addEventListener('keydown', onKeydown)
+
   const debouncedFullscreenToggle = debounce(function () {
     dispatch('toggleFullScreen')
   }, 1000, true)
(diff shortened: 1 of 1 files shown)


dsernst:
There was also discussion in that PR about taking a different approach: replacing the existing hotkeys (in `Menu > Playback`) instead of adding these new ones

image


So I also made a branch to take this other approach:

https://github.com/webtorrent/webtorrent-desktop/compare/master...dsernst:playback-hotkeys

image

👍 This commit was tested to work locally on macOS 10.14.4.

----

Happy to send this in as a PR instead.
I have no strong preference either way — I only want to use the new hotkeys. 

Hope this won't create a bikeshed... 🙃

dsernst:
> Looks great. I'm just missing the key `F` to go fullscreen :)

Added (https://github.com/webtorrent/webtorrent-desktop/pull/1579/commits/07816f74705e03477cbe54928ac0769f53d333dd). Works on local macOS 10.14.4 testing.

dsernst:
> @mathiasvr:
>
> Thanks for picking this up and addressing all my annoying comments! 😄
> 
> I still prefer the solution on your second branch with youtube style shortcuts only.
> I think the menu looks cleaner and allows new users to quickly learn the IMO preferred shortcuts.
> 
> However, to avoid more stalling I will accept this and let another person decide whether to merge this or the other (better) option. 😏

Yes, I can see pros & cons each way.

My recommendation for productive forward momentum (similar to your `PR approval`) is to merge these in as additional hotkeys for now, so they can immediately start being used, albeit undocumented.

And then I will make a new PR (like https://github.com/webtorrent/webtorrent-desktop/compare/master...dsernst:playback-hotkeys), to sort out whether to remove the old hotkeys and show these new ones in the Menus. That seems like a bit more of a drawn out conversation, but doesn't need to block getting these new ones in.

Borewit:
Let's indeed keep the ball roling guys.

PR #1580, opened 2019-05-05
Ignore JetBrains IntelliJ IDEA project files

Borewit:


feross:
We could include these in `.gitignore` if you would like, but I usually prefer that files which are specific to one person's development environment are ignored in your own personal global `.gitignore` file. See this for instructions on how to set that up: https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore

Borewit:
There are cases where I do check in some the IDE project files with project wide settings, in my experience not a lot of added value, but recommended way of working according to JetBrains).

Move it to personal global scope would effect those as well.

Putting it in .ignore can also be seen as statement, this kind of project files are not desired to be part of the project, even if it does not apply to all of us, it does not harm anyone.

PR #1581, opened 2019-05-06
Remove old hotkeys in favor of new YouTube style

dsernst:
This is a follow on to the discussions in https://github.com/webtorrent/webtorrent-desktop/pull/1579 and https://github.com/webtorrent/webtorrent-desktop/pull/1443.

The unresolved question was **whether it's better to have both the old hotkeys as well as the new ones, or to simply replace the old with the new.** 

#1579 was merged in to add both. This PR would remove the old hotkeys to only have the new ones.

image

Currently these new hotkeys are undocumented.

-----

### DIFF
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -224,13 +224,13 @@ function getMenuTemplate () {
         },
         {
           label: 'Increase Volume',
-          accelerator: 'CmdOrCtrl+Up',
+          accelerator: 'Up',
           click: () => windows.main.dispatch('changeVolume', 0.1),
           enabled: false
         },
         {
           label: 'Decrease Volume',
-          accelerator: 'CmdOrCtrl+Down',
+          accelerator: 'Down',
           click: () => windows.main.dispatch('changeVolume', -0.1),
           enabled: false
         },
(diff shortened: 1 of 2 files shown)


dsernst:
Relevant previous comments on this:

image



-----

image

-----

image




Speaking personally, I think I lean a little bit towards replacing the old with the new (merging this PR). 

The biggest advantage is that it consolidates the documentation for the new hotkeys (the only ones I plan to use) into the menu. It's also nice that it removes the `onKeyDown` code that doesn't match other conventions.

The one big disadvantage is that some people may be depending upon the old hotkeys? I'm not sure, I have no sense of how widely they are used, or if it's a problem for people for people to switch over to these simpler ones.

(*cough [xkcd 1172](https://xkcd.com/1172/) cough*)

dsernst:
If the new ones are preferred, but it's still important to preserve the legacy hotkeys, another option is to move these new ones into the menu, but still leave the old ones available (undocumented) using the `onKeyDown` logic.

feross:
Hey all,

My take is that we should show the YouTube shortcuts in the menus and prefer those as the defaults since they're easier and probably more familiar to users but we should support the existing shortcuts as well. The existing shortcuts are used by VLC, which is our preferred video player when a video is unsupported, and I think it will cause unnecessary confusion for users like me who only use VLC when WebTorrent can't play back the video. I'm constantly playing videos in both places and I don't want to mentally switch between the two sets of shortcuts.

We can use https://github.com/parro-it/electron-localshortcut to register the alternative shortcuts (the ones that don't appear in the menu).

PR #1582, opened 2019-05-06
Context menu option to sort files by name

dsernst:
Copying in comments from #1185 

> I want to kindly add the the current UI on this is not intuitive and makes handling large groups of files very difficult. This isn't exactly a fringe user experience. It's clearly come up a number of times — e.g. the linked issues marked as dupes.
> 
> In particular, I have a 10gb torrent w/ 200+ files, but only want to download a single one of them. It's possible to `Pause All` and then re-select just the needed file, but very difficult to find in such a large list if sorted by its default: filesize.
> 
> This has come up for me a handful of times. The current lack of support for **Sort by Name** requires keeping a different torrent program, exclusively for this use-case. Seems... excessive.
> 
> -----
> 
> I do see the previously quoted `WONTFIX` comment from @feross. 
> 
> And these from PR https://github.com/webtorrent/webtorrent-desktop/pull/1188:
> 
> image
> 
> -----
> 
> I certainly admire & agree with the sentiment to avoid creating a bloated Settings page. Yuck.
> 
> But maybe there's a simpler UI to handle this? What about a single context menu option on the list of files: "Sort by Name".
> 
> No addition to the Settings page is needed.
> The control would be right where someone uses it. 
> And not visible to anyone who doesn't right-click the file list.
> 
> -----
> 
> Current context menu:
> 
> Screen Shot 2019-05-03 at 1 58 17 PM

New context menu:

sort-by-name

------


nikkoman:
I would totally love this feature. Just a minimalistic additional option for sorting by name.

RecoX:
+1 to this! :)

src/renderer/pages/torrent-list-page.js
```diff
@@ -1,5 +1,6 @@
 const React = require('react')
 const prettyBytes = require('prettier-bytes')
+const natsort = require('natsort')
```
hicom150: What do you think about using javascript's native `localeCompare` instead an external library?

src/renderer/pages/torrent-list-page.js
```diff
@@ -278,10 +279,17 @@ module.exports = class TorrentList extends React.Component {
       )
     } else {
       // We do know the files. List them and show download stats for each one
-      const fileRows = torrentSummary.files
+      const sortByName = this.props.state.saved.prefs.sortByName
```
hicom150: The code with javascript's native `localeCompare` could be something like this:
```
const sortByName = this.props.state.saved.prefs.sortByName
const collator = new Intl.Collator(undefined, { numeric: true, sensitivity: 'base' })
let fileRows = torrentSummary.files
    .filter((file) => !file.path.includes('/.____padding_file/'))
    .map((file, index) => ({ file, index }))

if (sortByName) {
    fileRows = fileRows.sort((a, b) => collator.compare(a.file.name, b.file.name))
}

fileRows = fileRows.map((obj) => this.renderFileRow(torrentSummary, obj.file, obj.index))
```

ldexterldesign:
Hi @dsernst,

Thanks for PR

Love to see this feature in the wild ASAP

Yours hopefully

src/renderer/pages/torrent-list-page.js
dsernst: Thanks. I had kept the same logic as the previous commit this PR was extended from, but agree this would be easier to keep deps from bloating unnecessarily.

alxhotel:
I agree with @hicom150, we should keep the dependency tree as minimal as possible. Other than that, this PR looks good.

dsernst:
I've pushed up another commit to remove the extra `natsort` dependency, based on @hicom150's `Intl.Collator` suggestion. Thx! 🙌

dsernst:
And now fixed the merge conflict.

✅ Confirmed that it all works locally.

AFAIK this is ready to merge 🙂

mathiasvr:
Thanks @dsernst for updating this!
However, I think we should just make this the default sorting and remove the menu to keep the UI minimal.
Webtorrent Desktop did sort files by name early on, but this was not always desired.
I think this natural sorting solves most of the issues and should be the single default.

dsernst:
> (Common) use case: if I have 25 active torrents then the newly added torrent gets buried

@ldexterldesign: Just to be clear, this PR only allows sorting the files *within* a single torrent, not among the list of different torrents.

dsernst:
> Thanks @dsernst for updating this!
However, I think we should just make this the default sorting and remove the menu to keep the UI minimal.
Webtorrent Desktop did sort files by name early on, but this was not always desired.
I think this natural sorting solves most of the issues and should be the single default.

I don't have strong feelings about enforcing this as the only sorting method vs having a relatively out-of-the-way context menu option. I suspect

dsernst:
> Thanks @dsernst for updating this!
However, I think we should just make this the default sorting and remove the menu to keep the UI minimal.
Webtorrent Desktop did sort files by name early on, but this was not always desired.
I think this natural sorting solves most of the issues and should be the single default.

I don't have strong feelings either way about enforcing this as the only sorting method vs having a relatively out-of-the-way context menu option.

My hunch is I will want to keep this natural sort turned on all the time, as you suggest. I can't recall a time when I really needed to see the list of files sorted by the order they were added to the torrent file... 🙃

dsernst:
Separately, can someone help make sense of the Travis CI error? It looks like it got stuck while configuring `nvm`?

ldexterldesign:
foo

Apologies all

PR #1583, opened 2019-05-08
Removed zero-fill

jimmywarting:
sets the version the same way webtorrent dose it

https://github.com/webtorrent/webtorrent/blob/f8923a66a8b1178a412b067bc913e267025b3d57/index.js#L27-L29

### DIFF
--- a/src/renderer/webtorrent.js
+++ b/src/renderer/webtorrent.js
@@ -12,7 +12,6 @@ const mm = require('music-metadata')
 const networkAddress = require('network-address')
 const path = require('path')
 const WebTorrent = require('webtorrent')
-const zeroFill = require('zero-fill')
 
 const crashReporter = require('../crash-reporter')
 const config = require('../config')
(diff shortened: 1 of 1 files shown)


dsernst:
Can you explain the motivation ? :)

jimmywarting:
Have none other then "doing it the same way webtorrent" 

And also possible removing zero-fill completely in the second last place 

https://github.com/webtorrent/webtorrent-desktop/blob/81e6d7f3a756862de89861b7f152727ebb0ed0cb/src/renderer/pages/player-page.js#L741-L744

...by replacing it with native [padStart](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart) to reduce the dependencies

feross:
@jimmywarting Can you send another PR to completely remove `zero-fill`, as you mentioned in your comment? Thanks!

jimmywarting:
> Can you send another PR to completely remove zero-fill, as you mentioned in your comment? Thanks!

sure, will do it soon

PR #1481, opened 2018-09-25
Update to Electron 3

mathiasvr:
This PR updates electron, electron-packager, and spectron.

I have tried to fix issues related to updating these dependencies, but have only tested it on macOS 10.11.6.

I have fixed and run the integration tests for macOS (the new electron version e.g. renders fonts differently).

It would be great if someone could try this on other platforms, and maybe run/fix the integration tests.

This is a work in progress, current issues:

- [ ]  Wait for electron to fix #1490 
- [x]  Wait for spectron to officially support electron 3
- [ ]  Run integration tests on macOS Mojave
- [ ]  Run/fix integration tests on Windows 10

PR #1590, opened 2019-05-15
Update to electron 4

Borewit:
Related to PR: #1549.

I have not tested these changes thoroughly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,11 +16,11 @@
     "arch": "^2.0.0",
     "auto-launch": "^5.0.5",
     "bitfield": "^1.0.2",
-    "capture-frame": "^2.0.0",
+    "capture-frame": "^3.0.0",
     "chokidar": "^2.0.4",
     "chromecasts": "^1.9.1",
     "cp-file": "^7.0.0",
-    "create-torrent": "^3.33.0",
+    "create-torrent": "^4.0.0",
     "debounce": "^1.0.0",
     "deep-equal": "^1.0.1",
     "dlnacasts": "^0.1.0",
@@ -34,7 +34,7 @@
     "mkdirp": "^0.5.1",
     "music-metadata": "^3.6.1",
     "network-address": "^1.1.0",
-    "parse-torrent": "^6.0.1",
+    "parse-torrent": "^7.0.0",
     "prettier-bytes": "^1.0.1",
     "prop-types": "^15.6.2",
     "react": "^16.5.2",
@@ -51,10 +51,10 @@
     "zero-fill": "^2.2.3"
   },
   "devDependencies": {
-    "babel-eslint": "^9.0.0",
+    "babel-eslint": "^10.0.2",
     "buble": "^0.19.6",
     "cross-zip": "^2.0.1",
-    "depcheck": "^0.7.2",
+    "depcheck": "^0.8.0",
     "electron": "^4.0.0",
     "electron-osx-sign": "^0.4.11",
     "electron-packager": "~8.5.1",
@@ -70,7 +70,7 @@
     "spectron": "^3.3.0",
     "standard": "*",
     "tape": "^4.9.1",
-    "walk-sync": "^1.1.3"
+    "walk-sync": "^2.0.2"
   },
   "engines": {
     "node": ">=4.0.0"
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "buble": "^0.19.6",
     "cross-zip": "^2.0.1",
     "depcheck": "^0.7.2",
-    "electron": "^1.8.8",
+    "electron": "^4.0.0",
     "electron-osx-sign": "^0.4.11",
     "electron-packager": "~8.5.1",
     "electron-winstaller": "^2.6.4",
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -167,7 +167,7 @@ function onOpen (e, torrentId) {
     // Electron issue: https://github.com/atom/electron/issues/4338
     setTimeout(() => windows.main.show(), 100)
 
-    processArgv([ torrentId ])
+    processArgv([torrentId])
   } else {
     argv.push(torrentId)
   }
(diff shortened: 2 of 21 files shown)


dsernst:
Seeing bugginess w/ a duplicated title bar on macOS:

image

dsernst:
Reviewing items from the Electron Breaking Changes doc: https://github.com/electron/electron/blob/master/docs/api/breaking-changes.md

**Breaking API Changes (2.0)**

- [ ] `BrowserWindow` - The single use is updated in PR https://github.com/webtorrent/webtorrent-desktop/pull/1596
- [x] `menu` - Only one use of `menu.popup`, but doesn't use changed positioning params.
- [x] `nativeImage` - No use of `.toPng` or `.toJpeg`
- [x] `process` - Two calls to `process.versions.electron`, but both are reads, not writes. No calls to `process.version.chrome`.
- [x] `webContents` - No calls to `webContents.setZoomLevelLimits`
- [x] `webFrame` - No calls to `webFrame.setZoomLevelLimits`
- [x] `<webview>` - No calls to `webview.setZoomLevelLimits`
- [ ] Duplicate ARM Assets - Not quite sure what to make of this one. Linux specific. Not seeing any references to `linux-arm` in the codebase.

dsernst:
Reviewing items from the Electron Breaking Changes doc: https://github.com/electron/electron/blob/master/docs/api/breaking-changes.md#breaking-api-changes-30

**Breaking API Changes (3.0)**

- [x] `app`
  - [x] No calls to `.getAppMemoryInfo()`
  - [x] No calls to `.getAppMetrics()`
- [x] `BrowserWindow`
  - [x] No references to `blinkFeatures`
  - [x] No references to `media-play_pause`
- [x] `clipboard`
  - [x] No calls to `.readRtf()`
  - [x] No calls to `.writeRtf()`
  - [x] No calls to `.readHtml()`
  - [x] No calls to `.writeHtml()`
- [x] `crashReporter`
  - [x] [`crashReporter` is called](https://github.com/dsernst/webtorrent-desktop/blob/master/src/crash-reporter.js#L9), but webtorrent doesn't use the deprecated `autoSubmit` option.
  - [x] All the options being used are still supported, per [the docs](https://electronjs.org/docs/api/crash-reporter).
- [x] `nativeImage` — No calls to `.createFromBuffer`
- [x] `process` — No calls to `.getProcessMemoryInfo()`
- [x] `screen` — No calls to `.getMenuBarHeight()`
- [x] `session` — No calls to `.setCertificateVerifyProc`
- [x] `Tray` — No calls to `.setHighlightMode`
- [ ] `webContents`
  - [ ] `openDevTools({ detach: true })` ***BREAKING CHANGE***
    - **Requires fix.** Will update once up.
  - [x] No calls to `webContents.setSize`
- [x] `webFrame`
  - [x] No calls to `.registerURLSchemeAsSecure`
  - [x] No calls to `.registerURLSchemeAsPrivileged`
- [x] `<webview>`
  - [x] No references to `disableguestresize`
  - [x] No references to `guestinstance`
  - [x] No references to webview, so don't think these keyboard handlers are being used.
- [x] Node Headers URL — Think this one is good. Not seeing any references in the codebase to `atom-shell`, or an `.npmrc` file. Everything seems to be building fine.

Borewit:
Thanks a lot for your review @dsernst, very useful.

Note that I may have broken something by dropping the depreciated ```onAppOpen``` event.

I guess the idea was, if the application is already running and called a second time, the argument, presumably a torrent file, would be passed and added to the instance running.

dsernst:
Reviewing items from the Electron Breaking Changes doc: https://github.com/electron/electron/blob/master/docs/api/breaking-changes.md#planned-breaking-api-changes-40

**Breaking API Changes (4.0)**

- [x] `app.makeSingleInstance` — No calls to `.makeSingleInstance`
- [x] `app.releaseSingleInstance` — No calls to `.releaseSingleInstance`
- [x] `app.getGPUInfo` — No calls to `.getGPUInfo`
- [x] `win_delay_load_hook` — No references to `win_delay_load_hook`

dsernst:
> Note that I may have broken something by dropping the depreciated `onAppOpen` event.
> 
> I guess the idea was, if the application is already running and called a second time, the argument, presumably a torrent file, would be passed and added to the instance running.

Ah. Yes, can confirm that I'm having trouble adding a torrent via magnet link when the app is already open.

dsernst:
I'm trying to test my own fix for this... but even on `master` (electron v1) I'm unable to add new torrents if the `npm start` version of the app is already open.

mathiasvr:
You may want to take a look at my Electron 3 PR which was not merged #1481.
This should handle the breaking changes for v2 and v3.

Borewit:
With the remaining fix @mathiasvr put in, it looks good to me.
Can you confirm @dsernst?
Would be nice to upgrade Electron as well as part of the new release (#1604) 😉

package.json
```diff
@@ -55,7 +55,7 @@
     "buble": "^0.19.6",
     "cross-zip": "^2.0.1",
     "depcheck": "^0.7.2",
-    "electron": "^1.8.8",
+    "electron": "^4.0.0",
```
feross: This is great!

src/main/index.js
```diff
@@ -31,19 +34,30 @@ if (process.platform === 'win32') {
   const squirrelWin32 = require('./squirrel-win32')
   shouldQuit = squirrelWin32.handleEvent(argv[0])
   argv = argv.filter((arg) => !arg.includes('--squirrel'))
+
+  if (shouldQuit) {
+    app.quit()
+  } else {
+    app.on('second-instance', (event, commandLine) => onAppOpen(commandLine))
+  }
```
feross: Why is this logic repeated below? It seems like this will cause the `'second-instance'` event to be registered twice. Is this intentional?

feross:
Also, can you resolve `package.json` conflicts?

Borewit:
@feross, can you please update the review?

feross:
Thanks for landing this, @Borewit!

DiegoRBaquero:
Beautiful :´)

PR #1596, opened 2019-06-04
Fix hiddenInset (Mac) titlebar style for electron 4

dsernst:
This fixes bugginess w/ a duplicated title bar on macOS, as part of the upgrade to Electron 4 (https://github.com/webtorrent/webtorrent-desktop/pull/1590):

### Before:

image

### After:

image

----

Docs about this param: https://electronjs.org/docs/api/frameless-window#hiddeninset 

Note about this breaking change introduced in Electron 2.0: https://github.com/electron/electron/blob/master/docs/api/breaking-changes.md#browserwindow-1

>


Borewit:
I very much appreciate your help @dsernst.

PR #1602, opened 2019-06-09
Show torrent controls item

Muhnad:
## Problem
to reach and use the action items like remove the torrent or open the video you need to hover on the item and as we know the hover action is only available by a mouse so while you use a keyboard you can't reach the buttons.

so I make the buttons visible without needing to hover and enhanced the UI layout.

Screen Shot 2019-06-09 at 8 02 19 PM

### DIFF
--- a/static/main.css
+++ b/static/main.css
@@ -341,12 +341,14 @@ textarea,
   border-bottom: 1px solid #282828;
 }
 
+.torrent__meta {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+}
+
 .torrent .metadata {
-  position: absolute;
-  top: 20px;
-  left: 15px;
-  right: 15px;
-  width: calc(100% - 30px);
+  padding: 20px 15px;
   text-shadow: rgba(0, 0, 0, 0.5) 0 0 4px;
 }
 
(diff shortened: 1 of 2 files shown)


static/main.css
```diff
@@ -472,7 +463,7 @@ body.drag .app::after {
 }
 
 .torrent-details {
-  padding: 8em 0 20px 0;
```
Muhnad: this was hack used to fix the positioning layout break, right now I removed the absolute position and used flex and enhanced the layout.

dsernst:
Hmm. From a clean UI point of view, this approach seems like a bit of a regression because it greatly multiplies the number of buttons to choose from at any one time.

But I like your thinking about keyboard-only accessibility.

What about being able to navigate through the list of torrents with up/down arrows, or j/k (like Gmail hotkeys)? And then maybe using Enter to Play, or Backspace to Delete?

Muhnad:
@dsernst Definitely we need to build a keyboard system and guideline inside the app so let's take a step forward and start making a new issue to determine exactly what hotkeys we need to use and where the keyboard guideline map we need to put inside the app and so on. you can start this and mention me + due to my military service so almost I'm not available to receive this kind of notification until I got a vacation but I'm excited to start doing this in my vacation time.

What I did in this PR not consider as a regression for a couple of reasons:
- it should be implemented from the first time accessible by using the right element to keep the app works for all users.
- so I just make this happen by this small enhancement until we make the big one.
consider this as hotfix until revamping your app.

PR #1635, opened 2019-08-04
Update changelog for v0.21.0

mathiasvr:
This is a list of changes in the form of PRs that have been merged since `v0.20`.

In addition, many changes and PRs have been made regarding: dependency updates, code refactoring, and style updates. I have not included them as it looks a bit overwhelming, let me know if you think we should include it somehow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,34 @@
 # WebTorrent Desktop Version History
 
+## v0.21.0 - 2019-08
+
+### Added
+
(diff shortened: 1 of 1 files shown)


feross:
> many changes and PRs have been made regarding: dependency updates, code refactoring, and style updates. I have not included them as it looks a bit overwhelming, let me know if you think we should include it somehow.

We should only include user-facing changes. Perhaps one exception is that we should also include which version of `webtorrent` is used, since that really helps folks who are building stuff on the web with WebTorrent :)

mathiasvr:
Sure that makes sense.
And BTW we should remember to update webtorrent to the latest version.

PR #1638, opened 2019-08-04
Refactor bound control

hicom150:


src/main/windows/main.js
```diff
@@ -135,43 +135,18 @@ function setAspectRatio (aspectRatio) {
...
  */
-function setBounds (bounds, maximize) {
-  // Do nothing in fullscreen
-  if (!main.win || main.win.isFullScreen()) {
-    log('setBounds: not setting bounds because already in full screen mode')
-    return
-  }
+function setBounds (bounds) {
+  // X and Y not specified? By default, center on current screen
+  if (bounds.x === null && bounds.y === null) {
```
alxhotel, line 140: `bounds` could still be null from the saved state of the user.

alxhotel (changes requested):

PR #1641, opened 2019-08-07
Update to Electron 6

feross:
~~DO NOT MERGE.~~ Ready to merge!

There seems to be an issue with the media codecs in this version of Electron. Sintel and the WIRED CD do not play back directly in Electron anymore. Instead, VLC is recommended. We need to dig into this problem to figure out if the problem is on our end or on Electron's. If it's Electron, we need to file an issue.

### DIFF
--- a/bin/extra-lint.js
+++ b/bin/extra-lint.js
@@ -18,13 +18,6 @@ files.forEach(function (file) {
   lines.forEach(function (line, i) {
     let error
 
-    // Consistent JSX tag closing
-    if (line.match(/' {2}\/> *$/) ||
-      line.match('[^ ]/> *$') ||
-      line.match(' > *$')) {
-      error = 'JSX tag spacing'
-    }
-
     // No lines over 100 characters
     if (line.length > 100) {
       error = 'Line >100 chars'
--- a/package-lock.json
+++ b/package-lock.json
@@ -2614,9 +2614,9 @@
       "dev": true
     },
     "electron": {
-      "version": "4.2.8",
-      "resolved": "https://registry.npmjs.org/electron/-/electron-4.2.8.tgz",
-      "integrity": "sha512-/D9zfs+EWLN4yLV7tu2kWyXUnZQ3CKG1cmWbXeSFXF+0dNXQ8iFpY49dqZRoHGIBImFfp2x4N3Zc5Tu7rw3PJw==",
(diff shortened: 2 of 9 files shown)


hicom150:
I don't know if I have permission to push commits to this branch, so here I will write the changes that work for me to make it work with electron 6.

- Small change in https://github.com/webtorrent/webtorrent-desktop/blob/master/src/renderer/lib/telemetry.js
  Change `electron.screen.getAllDisplays()` with `electron.remote.screen.getAllDisplays()`
```
function getScreenInfo () {
  return electron.remote.screen.getAllDisplays().map((screen) => ({
    width: screen.size.width,
    height: screen.size.height,
    scaleFactor: screen.scaleFactor
  }))
}
```

- Another small change in https://github.com/webtorrent/webtorrent-desktop/blob/master/src/renderer/pages/player-page.js
  It seems that `onCanPlay` is called at least twice, where in each call with different `elem.readyState`
  As we can see in [HTMLMediaElement Documentation](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/readyState) `readyState` should be >= 3 to start playing, so this fixed the problem for me:
```
function onCanPlay (e) {
    const elem = e.target

    if (elem.readyState < 3) return

    if (state.playing.type === 'video' &&
      elem.webkitVideoDecodedByteCount === 0) {
      dispatch('mediaError', 'Video codec unsupported')
    } else if (elem.webkitAudioDecodedByteCount === 0) {
      dispatch('mediaError', 'Audio codec unsupported')
    } else {
      dispatch('mediaSuccess')
      elem.play()
    }
  }
}
```

Hope it helps 😉

feross:
@hicom150 Brilliant! Your fixes indeed solve the issue. I appreciate it very much! I think we can go ahead and merge this PR now.

> I don't know if I have permission to push commits to this branch, so here I will write the changes that work for me to make it work with electron 6.

Little known GitHub trick: You can actually send a pull request for a pull request! The way you do it is just to send a pull request like normal, but instead selecting the `master` branch as the target of the PR, you select the branch for the PR you want to add to, e.g. `electron-6`.

feross:
Thanks for the review, @Borewit.

> It looks good to me, downloading audio files in slow motion works 😉

Hah!

Issue #1637, opened 2019-08-04
Stop shipping 32-bit binaries


feross:
We currently offer 32-bit binaries on Windows and Linux. I'd like to stop shipping these and go 64-bit only. It'll mean less work for us. And the 32-bit builds have always had a higher chance of fatal errors due to running out of the limited memory address space available to 32-bit processes.

The [latest OS stats](https://store.steampowered.com/hwsurvey/Steam-Hardware-Software-Survey-Welcome-to-Steam) indicate that 32-bit OSs are almost completely gone.

<img width="395" alt="Screen Shot 2019-08-04 at 10 30 54 AM" src="https://user-images.githubusercontent.com/121766/62427132-38c6fc00-b6a3-11e9-9f4f-d7a59f63356d.png">

Windows 7 32-bit is 1.06% and Windows 10 32-bit is 0.19% of overall usage. Linux 32-bit doesn't even show up on the list. Users on 32-bit systems can continue to use the current version of WebTorrent Desktop which will continue working just fine. Or they can build WebTorrent Desktop from source.

Feedback welcome.

hicom150:
I think it is a good decision to go forward and more given that electron is [discontinuing support for 32-bit Linux](https://electronjs.org/blog/linux-32bit-support)

feross:
> I think it is a good decision to go forward and more given that electron is [discontinuing support for 32-bit Linux](https://electronjs.org/blog/linux-32bit-support)

Well, looks like our decision has been made for us, at least on Linux. We use Electron 4 at the moment.

Borewit:
I agree with dropping support for 32-bit Linux, but I would give issue #1584 another tought.

Unless you only want support Debian Linux, I don't think a zip is sufficient to satisfy the non-Debian part of Linux community. A "simple user" does not accept a zip file, he wan't something which installs like an app, and he can rid with same speed. As simple as the app store on your iPhone.

I think that is in line with your vision keeping webtorrent-desktop a simple no-nonse torrent client accessible for every user (users which do hang around on GitHub).

feross:
> Unless you only want support Debian Linux, I don't think a zip is sufficient to satisfy the non-Debian part of Linux community. A "simple user" does not accept a zip file, he want something which installs like an app, and he can rid with same speed. As simple as the app store on your iPhone.

[...]

PR #1642, opened 2019-08-07
Stop shipping 32-bit binaries

feross:
For: https://github.com/webtorrent/webtorrent-desktop/issues/1637


hicom150:
LGTM! 👍

All I can add is that maybe we should merge #1660 into this PR to fix packaging process 😉

feross:
Thanks for the review, @hicom150! ❤️ Thanks also for the followup PR!

Issue #1645, opened 2019-08-12
npm run clean fails


aurabindo:
**What version of WebTorrent Desktop?** master (f6bd2cfcf4f5ae48a7)

**What operating system and version?** Archlinux 

**What did you do?** `git clone repo..`, `npm install`, `npm run build`, `npm run clean`

**What did you expect to happen?** complete without errors

**What actually happened?**:
```

> webtorrent-desktop@0.20.0 clean /run/media/aj/Buffer/git-clones/webtorrent-desktop
> node ./bin/clean.js

assert.js:84
  throw new AssertionError(obj);
  ^

AssertionError [ERR_ASSERTION]: rimraf: callback function required
    at rimraf (/run/media/aj/Buffer/git-clones/webtorrent-desktop/node_modules/rimraf/rimraf.js:52:10)
    at uninstallLinux (/run/media/aj/Buffer/git-clones/webtorrent-desktop/src/main/handlers.js:358:3)
    at Object.uninstall (/run/media/aj/Buffer/git-clones/webtorrent-desktop/src/main/handlers.js:26:19)
    at Object.<anonymous> (/run/media/aj/Buffer/git-clones/webtorrent-desktop/bin/clean.js:33:10)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:831:12)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! webtorrent-desktop@0.20.0 clean: `node ./bin/clean.js`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the webtorrent-desktop@0.20.0 clean script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/aj/.npm/_logs/2019-08-12T06_54_32_841Z-debug.log
```
Contents of debug log file: https://paste.ubuntu.com/p/cg2mT26YrW/

hicom150:
Should be fixed with #1646 😉

feross:
This has been fixed!

PR #1646, opened 2019-08-12
Fix uninstall handlers in Linux

hicom150:
As described in #1645 `npm run clean` fail in Linux as `rimraf` expects a callback to be defined unless we use the `.sync` method.

This PR fixes this issue.


feross:
@hicom150 Thank you for the excellent PR!

Issue #1349, opened 2018-04-01
No way to pass title metadata to MPC HC


johnzach:
**What version of WebTorrent Desktop?** (See the 'About WebTorrent' menu)
0.19.0

**What operating system and version?**
Windows 10 

**What did you do?**
Open external player

**What did you expect to happen?**
have the name of the torrented video file as a title

**What actually happened?**
Had '1' as a title or "localhost" which prevents me from autosearching subtitles

Is there a way to pass actual video file title to the player?

feross:
If it's showing `1` then I think that it's fixable. In the WebTorrent server we expose two types of URLs:

```
https://localhost:port/<file-index>
https://localhost:port/<file-index>/<file-name>
```

Which are equivalent, but the latter can be useful when a file name is being inferred by the URL. We just need to switch to the latter URL format to improve the situation it seems.

This should be a relatively easy fix.

hicom150:
Should be fixed in #1648 😉

feross:
Fixed by #1648

PR #1648, opened 2019-08-13
Pass title metadata in media url

hicom150:
This is a small fix to show the file name when streaming content.

As @feross suggested in #1349 changing `mediaUrl` from `https://localhost:port/<file-index>` to `https://localhost:port/<file-index>/<file-name>` fixes the issue.

### DIFF
--- a/src/renderer/controllers/playback-controller.js
+++ b/src/renderer/controllers/playback-controller.js
@@ -269,6 +269,7 @@ module.exports = class PlaybackController {
     // update state
     state.playing.infoHash = infoHash
     state.playing.fileIndex = index
+    state.playing.fileName = fileSummary.name
     state.playing.type = TorrentPlayer.isVideo(fileSummary) ? 'video'
       : TorrentPlayer.isAudio(fileSummary) ? 'audio'
         : 'other'
--- a/src/renderer/lib/playlist.js
+++ b/src/renderer/lib/playlist.js
@@ -38,7 +38,7 @@ function getPreviousIndex (state) {
 
 function getCurrentLocalURL (state) {
   return state.server
-    ? state.server.localURL + '/' + state.playing.fileIndex + '/' + state.window.title
+    ? state.server.localURL + '/' + state.playing.fileIndex + '/' + encodeURIComponent(state.playing.fileName)
     : ''
 }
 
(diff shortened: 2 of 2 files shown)


src/renderer/lib/playlist.js
```diff
@@ -38,7 +38,7 @@ function getPreviousIndex (state) {
 
 function getCurrentLocalURL (state) {
   return state.server
-    ? state.server.localURL + '/' + state.playing.fileIndex
+    ? state.server.localURL + '/' + state.playing.fileIndex + '/' + state.window.title
```
feross: Getting the file name off of the window title is not ideal. What if we change the Window title to not be an exact match of the filename in the future?

We should add the file name to `state.playing` and then reference it here.

hicom150: > We should add the file name to state.playing and then reference it here.

That's right! 😉

> Getting the file name off of the window title is not ideal. What if we change the window title to not be an exact match of the filename in the future?

I don't have any better idea at the moment, so could we leave this for the future?

src/renderer/controllers/playback-controller.js
```diff
@@ -269,6 +269,7 @@ module.exports = class PlaybackController {
     // update state
     state.playing.infoHash = infoHash
     state.playing.fileIndex = index
+    state.playing.fileName = fileSummary.name
```
feross: We should add a default value to the state file so that it acts as documentation. I will add that in a followup commit.

src/renderer/lib/playlist.js
```diff
@@ -38,7 +38,7 @@ function getPreviousIndex (state) {
 
 function getCurrentLocalURL (state) {
   return state.server
-    ? state.server.localURL + '/' + state.playing.fileIndex
+    ? state.server.localURL + '/' + state.playing.fileIndex + '/' + encodeURIComponent(state.playing.fileName)
```
feross: This line is over 100 chars so the linter complains. I will fix.

Borewit:
I would fix it by increasing the maximum line length. Breaking up lines or removing trailing comments or reducing the length of variables does not make code more readable in many occasions.

feross:
@Borewit Yeah, I generally agree with you. `standard` does not enforce line length. That comes from this file: https://github.com/webtorrent/webtorrent-desktop/blob/master/bin/extra-lint.js

hicom150:
Thank you guys! I appreciate your help 💪

feross:
Thank you for the nice PR, @hicom150!

PR #1650, opened 2019-08-15
introduce linux arm64 support

gdamjan:
I've tested building this on an arm64 VM, just in case, only now I realize that it should build on any other linux system too.

Tested running it on an arm64 Chromebook in Crostini support (was a bit slow).


feross:
Thanks for this PR. I'll look at merging it when doing the next release of the app. I think a few other things might need to change as well. Will leave this open as a reminder.

feross:
This needs some changes to actually produce .deb and .zip files which are what we ship to users. I'll followup with a PR to fix this.

feross:
Followup PR: https://github.com/webtorrent/webtorrent-desktop/pull/1696

PR #1653, opened 2019-08-15
Fix path selector value when path is not defined

hicom150:
In PathSelector component value is mapped to electron's `showOpenDialog` `defaultPath` property that should be a valid string.

In `preferences-page.js` in some cases that value was passed as `null` and causes an error in `showOpenDialog` preventing from displaying it.

This PR fixes the issue passing `''` as a default value when no there is no valid path.


feross:
Fixed by https://github.com/webtorrent/webtorrent-desktop/pull/1498

PR #1657, opened 2019-08-18
Fix integration test setup

hicom150:
It seems that with the update to `spectron@8.0.0`
```
app.setPath('userData', config.CONFIG_PATH)
```
fails unless we set
```
chromeDriverArgs: [`--user-data-dir=${userDataDir}`],
```
in integration test setup.


feross:
@hicom150 For some reason, this PR is showing a ton of commits that are not part of the PR. I think you might need to rebase on master. You can do this with `git checkout master`, then `git pull`, then switch back to your feature branch with `git checkout update_spectron`, and then `git rebase master`. Then you can force push your changes to your remote branch with `git push origin update_spectron -f`.

If it's not too much trouble, can you give that a try?

hicom150:
Sorry I made a mess with this PR 😅
I think I have solved now. Thank you for your advises 👍

feross:
Glad that I could be helpful :) This PR looks good!

PR #1419, opened 2018-06-07
Fix startup problems (fix for #1417)

Borewit:
Reference: issue: webtorrent/webtorrent-desktop#1417

A promise resolve result ended up as an err argument in the callback, causing initialization of application data to fail.

PR #1679, opened 2019-09-07
Use promise then() instead of catch() with callbacks

feross:

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

Addresses my review here: https://github.com/webtorrent/webtorrent-desktop/pull/1419#pullrequestreview-285182318

### DIFF
--- a/src/renderer/lib/state.js
+++ b/src/renderer/lib/state.js
@@ -151,8 +151,10 @@ function setupStateSaved (cb) {
   })
 
   Promise.all(tasks)
-    .then(() => cb(null, saved))
-    .catch(err => cb(err))
+    .then(
+      () => cb(null, saved),
+      err => cb(err)
+    )
 
   function createTorrentObject (t) {
     // TODO: Doing several fs.readFileSync calls during first startup is not ideal
--- a/src/renderer/webtorrent.js
+++ b/src/renderer/webtorrent.js
@@ -358,11 +358,15 @@ function getAudioMetadata (infoHash, index) {
     : mm.parseStream(file.createReadStream(), file.name, options)
 
   onMetaData
-    .then(() => {
-      console.log(`metadata for file='${file.name}' completed.`)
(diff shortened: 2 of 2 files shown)


feross:
@Borewit These changes weren't about readability but about correctness. The `.catch` will catch more than just the promise rejections from the `tasks` array. It will also catch any exceptions thrown from the handler in `.then()` which means that if the `cb` function throws an exception, then we'll catch it here. Instead we should use `.then(success, fail)` for correct behavior.

Borewit:
I honestly don't see the difference Ferros. 
I think in either cases it will catch all.

feross:
@Borewit There's a difference. See https://stackoverflow.com/a/33278420/292185 Let me know if it still doesn't make sense and I'll try to explain further.

Borewit:
Ah yes, you are right Feross! I completely missed the point.

```js
catch(errorCallback)` is an alias
```

for

```js
then(null, errorCallback)
```

where the alias is appended and not instead of the error handler of the previous then.

codealchemist:
This has an interesting behavior:
```
const p1 = new Promise((resolve, reject) => {
  throw(new Error('Promise 1 THROWN'))
})

const p2 = new Promise((resolve, reject) => {
  setTimeout(() => {
    throw(new Error('Promise 2 THROWN'))
  })
})

p1
  .then(null, message => console.log('P1 failed:', message))
  .catch(message => console.log('P1 catched:', message))

p2
  .then(null, message => console.log('P2 failed:', message))
  .catch(message => console.log('P2 catched:', message))
```

You can [run it here](https://repl.it/repls/BuzzingFunctionalAdware).
If the code fails synchronously, the failure callback of `then` will be called.
If there's no failure callback it will naturally flow thru `catch`.
If it fails asynchronously, it won't be caught by neither the failure callback nor the catch.

PR #1682, opened 2019-09-09
Replace `nobin-debian-installer`

hicom150:

[ ] Documentation update
[ ] Bug fix
[x] New feature
[ ] Other, please explain:

Replace `nobin-debian-installer` with `electron-installer-debian`.

This change has some implications:
- The code in [handlers.js](https://github.com/webtorrent/webtorrent-desktop/blob/master/src/main/handlers.js) that manages linux `desktop-file` and `icon` installation get simplified as `electron-installer-debian` takes the responsibility to install it correctly. 👍
- [postinst](https://github.com/webtorrent/webtorrent-desktop/blob/master/static/linux/postinst), [prerm](https://github.com/webtorrent/webtorrent-desktop/blob/master/static/linux/prerm) and [webtorrent-desktop.desktop](https://github.com/webtorrent/webtorrent-desktop/blob/master/static/linux/webtorrent-desktop.desktop) get removed as `electron-installer-debian` handles binary installation and permissions correctly. 👍
- Linux installation path gets changed to `/usr/lib` by default as it is explained [here](https://github.com/electron-userland/electron-installer-debian#installed-package)

In addition, along with `electron-installer-debian` there are another installer flavors that could help to reach new distributions:
- [electron-installer-redhat](https://github.com/electron-userland/electron-installer-redhat)
- [electron-installer-snap](https://github.com/electron-userland/electron-installer-snap)
- [electron-installer-windows](https://github.com/electron-userland/electron-installer-windows)
- [electron-installer-dmg](https://github.com/electron-userland/electron-installer-dmg)

and some more...

Maybe it is worth to give it a try 😉

Although I test this changes carefully in Ubuntu 18.04.3 LTS every check or suggestion is more than welcome 😊


bin/package.js
```diff
@@ -501,35 +501,30 @@ function buildLinux (cb) {
...
+      mimeType: ['application/x-bittorrent', 'x-scheme-handler/magnet', 'x-scheme-handler/stream-magnet'],
+      desktopTemplate: path.join(config.STATIC_PATH, 'linux/webtorrent-desktop.ejs')
+    }
+
+    installer(options)
+      .then(function () {
+        console.log('Linux: Created deb.')
+        cb(null)
+      })
+      .catch(function (err) {
```
feross: Using `.catch` means that this will catch any exceptions thrown by `cb(null)` in the `.then` function. We only want to catch exceptions thrown by the call to `installer(options)`.

Can you change this to use the two-argument `.then` function?

static/linux/webtorrent-desktop.ejs
```diff
@@ -0,0 +1,26 @@
...
+Name=Create New Torrent...
+<% if (name) { %>Exec=<%= name %> -n <% } %>
+
+[Desktop Action OpenTorrentFile]
+Name=Open Torrent File...
+<% if (name) { %>Exec=<%= name %> -o <% } %>
+
+[Desktop Action OpenTorrentAddress]
+Name=Open Torrent Address...
+<% if (name) { %>Exec=<%= name %> -u <% } %>
```
feross: These actions don't show up in Ubuntu for me when I right click on the dock icon. Is that the place to look for them? I tried installing the current latest version and I also couldn't find these, so has this just been broken for a while? Also, are you sure that `name` is the right variable? We had a path here before, I think.

src/main/handlers.js
```diff
@@ -269,100 +265,3 @@ function uninstallWin32 () {
...
-
-  const iconFilePath = path.join(
-    os.homedir(),
-    '.local',
-    'share',
-    'icons',
-    'webtorrent-desktop.png'
-  )
-  rimraf.sync(iconFilePath)
-}
```
feross: I like seeing this much deleted code!

bin/package.js
hicom150: You are right! 👍 

> Can you change this to use the two-argument .then function?

Sure!

src/main/handlers.js
hicom150: 😍

static/linux/webtorrent-desktop.ejs
hicom150: > These actions don't show up in Ubuntu for me when I right click on the dock icon. Is that the place to look for them?

I think the actions are correctly defined according to the [official documentation](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html) Furthermore, I can see them correctly in Ubuntu 18.04.3 LTS

screenshot
 
But it would be helpful if anyone else could check this also 😅

> Also, are you sure that name is the right variable? We had a path here before, I think.

Looking in the [official documentation](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html) a `Exec` key is the one that seems we need 😉

feross: Okay sounds good. I was right clicking on it in the Dock, not from the launcher like you show in your screenshot. Sounds good to me.

Issue #1674, opened 2019-09-06
Release WebTorrent Desktop 0.21.0


feross:
I'll share updates about my progress making this release. Haven't done a release in a while, so there's bound to be hiccups. 😉

feross:
Located backup drive with signing keys and certificates. They're in the name "Feross Aboukhadijeh". I already switched the developer account to "WebTorrent, LLC" so could generate new certs but concerned that this cause issues with the autoupdate (would need to do more testing) or issues with new Mac app notarization requirements (see https://github.com/webtorrent/webtorrent-desktop/issues/1675, but old certificates don't have as strict of requirements, for now)

feross:
Able to build the Mac app without issues. Tested it and it's properly signed! Hooray!

```
check-signature dist/WebTorrent.app
(c) 2014 Apple Inc.  All rights reserved.
YES
```

feross:
Downloaded the app from a browser so the Quarantine bit is set and confirmed that it opens without the need to Right Click > Open. Though, there's a scarier first-run dialog than there is for older versions of WebTorrent.

Old:

<img width="532" alt="Screen Shot 2019-09-05 at 11 00 45 PM" src="https://user-images.githubusercontent.com/121766/64404535-1d983500-d031-11e9-9b38-7318c93c22ed.png">

New:

<img width="532" alt="Screen Shot 2019-09-05 at 11 01 29 PM" src="https://user-images.githubusercontent.com/121766/64404540-20932580-d031-11e9-8a9d-700170b5c177.png">

The fact that Apple supposedly analyzed the old version makes me think they're scraping binaries that Safari users download and checking them for malware. And they've added the old WebTorrent binaries to their database as safe even though they are only signed, **not** notarized. Fascinating.

feross:
The `nobin-debian-installer` package doesn't work on Node 10 due to calling a function without a callback. So I forked it, pulled in the PRs from @hicom150, and we're now using it (https://github.com/webtorrent/webtorrent-desktop/issues/1676). 

Tested the .deb file on Ubuntu. Seems to work.

feross:
Booting my Windows 10 VM. Loaded signing keys. Updated to latest Node.js and Git. Ready to try building and signing the app.

feross:
Ran into a snag. Running the app on Windows, I currently get:

```
[...]

PR #1693, opened 2019-09-14
Fix "EPERM: operation not permitted, open 'C:\Users\account\AppData\Local

feross:

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

This is the error on Windows that was preventing release of WebTorrent Desktop 0.21.0. https://github.com/webtorrent/webtorrent-desktop/issues/1674#issuecomment-528746939

Is the Local folder read-only? Possible long-term fix is to remove 'application-config' dependency and use `electron.app.getPath('appData')` directly.


feross:
Seems to still get the same error. So the issue is not `Local` vs `Roaming`. It seems to be inconsistent too...

Trying to swap out `cp-file` for a plain Node.js `copyFileSync` in case there's some bug in `cp-file`.

feross:
Replacing `cp-file` seems to have fixed it. I can't reproduce it anymore. I don't really know what to make of this... but I'm eager to get 0.21.0 out. So let's push ahead.

Issue #1502, opened 2018-10-17
.rpm package for Fedora, etc


adatum:
Are there plans to make a .rpm package available for Fedora and other rpm-based distributions, like the .deb that is already provided? It would be appreciated!

TeknoMobil:
I have a lot of deployment using Fedora and other rpm packages

feross:
Currently, Ubuntu is the only officially-supported version of Linux, though WebTorrent Desktop works great on many other flavors. I'm open to officially supporting additional flavors, but we need a volunteer to send a PR to produce the new build artifact and to agree to maintain it going forward.

So, we need answers to these questions:

- How do we produce the new build artifact? Is there a Node.js package, or does the Electron build tooling support it already?
- Who will send the PR to add it?
- Who will maintain it going forward?

Borewit:
Implemented with PR #1694

PR #1694, opened 2019-09-15
Add rpm package creation support

hicom150:

[ ] Documentation update
[ ] Bug fix
[x] New feature
[ ] Other, please explain:

Given that we have recently added `electron-installer-debian` it is straightforward to add `electron-installer-redhat` and add rpm file creation support.

It would be nice if someone with Fedora could check if the generated package works as expected 😅

Related:
* #1502 _.rpm package for Fedora, etc_

### DIFF
--- a/bin/package.js
+++ b/bin/package.js
@@ -488,6 +488,9 @@ function buildLinux (cb) {
       if (argv.package === 'deb' || argv.package === 'all') {
         tasks.push((cb) => packageDeb(filesPath, cb))
       }
+      if (argv.package === 'rpm' || argv.package === 'all') {
+        tasks.push((cb) => packageRpm(filesPath, cb))
+      }
       if (argv.package === 'zip' || argv.package === 'all') {
         tasks.push((cb) => packageZip(filesPath, cb))
       }
@@ -520,6 +523,35 @@ function buildLinux (cb) {
     installer(options).then(
       () => {
         console.log('Linux: Created deb.')
+        cb(null)
+      },
+      (err) => cb(err)
+    )
+  }
+
+  function packageRpm (filesPath, cb) {
+    // Create .rpm file for RedHat-based platforms
+    console.log('Linux: Creating rpm...')
+
+    const installer = require('electron-installer-redhat')
+
+    const options = {
+      src: filesPath + '/',
+      dest: DIST_PATH,
+      arch: 'amd64',
+      bin: 'WebTorrent',
+      icon: {
+        '48x48': path.join(config.STATIC_PATH, 'linux/share/icons/hicolor/48x48/apps/webtorrent-desktop.png'),
+        '256x256': path.join(config.STATIC_PATH, 'linux/share/icons/hicolor/256x256/apps/webtorrent-desktop.png')
+      },
+      categories: ['Network', 'FileTransfer', 'P2P'],
+      mimeType: ['application/x-bittorrent', 'x-scheme-handler/magnet', 'x-scheme-handler/stream-magnet'],
+      desktopTemplate: path.join(config.STATIC_PATH, 'linux/webtorrent-desktop.ejs')
+    }
+
+    installer(options).then(
+      () => {
+        console.log('Linux: Created rpm.')
         cb(null)
       },
       (err) => cb(err)
--- a/bin/package.js
+++ b/bin/package.js
@@ -485,12 +485,7 @@ function buildLinux (cb) {
 
     const tasks = []
     buildPath.forEach(function (filesPath) {
-      let destArch = filesPath.split('-').pop()
-
-      // Linux convention for 'x64' is 'amd64'
-      if (destArch === 'x64') {
-        destArch = 'amd64'
-      }
+      const destArch = filesPath.split('-').pop()
 
       if (argv.package === 'deb' || argv.package === 'all') {
         tasks.push((cb) => packageDeb(filesPath, destArch, cb))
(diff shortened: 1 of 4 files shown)


Borewit:
I would expect that this command only builds the rpm, but the target platform is still `all`
```sh
npm run package linux ---platform=rpm
```

bin/package.js
```diff
@@ -488,6 +488,9 @@ function buildLinux (cb) {
       if (argv.package === 'deb' || argv.package === 'all') {
         tasks.push((cb) => packageDeb(filesPath, cb))
       }
+      if (argv.package === 'rpm' || argv.package === 'all') {
```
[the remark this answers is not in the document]
hicom150: Ups, I forgot it 😅 I'm adding it right now 👍
Thank you for the review! 😉

feross:
> @Borewit npm run package linux ---platform=rpm

The command to use is `npm run package -- linux --platform=rpm`.

hicom150:
Done! 👍

feross:
I tried building this and ran into these issues:

- The Homebrew system dep is called `rpm`, not `rpmbuild`. Changing this allowed me to build an `rpm`.

- When double-clicking on the .rpm file in CentOS, I get the following error:

Screen Shot 2019-09-17 at 2 57 44 PM

hicom150:
Thank you @feross for the feedback 😉

> The Homebrew system dep is called rpm, not rpmbuild. Changing this allowed me to build successfully.

Nice catch, changed! 👍

> When double-clicking on the .rpm file in CentOS, I get the following error

Could you try with `yum install ` to check what could be the error cause?

In the other hand, I also try to install in a VM with `Fedora 30` and get a "incompatible architecture" error. So I can guess that for `RedHat` based distributions we have to set `destArch='x86_64'` and for `Debian` based ones `destArch='amd64'`.

I commit this last change so we can test that this assumption is right for all the different distributions 😅

bin/package.js
```diff
@@ -508,6 +506,11 @@ function buildLinux (cb) {
     // Create .deb file for Debian-based platforms
     console.log(`Linux: Creating ${destArch} deb...`)
 
+    // Linux convention for Debian based 'x64' is 'amd64'
+    if (destArch === 'x64') {
+      destArch = 'amd64'
+    }
```
feross: Nit: Can you move this line above the `console.log` above so that the `Linux: Creating ${destArch} deb...` and `Linux: Created ${destArch} deb.` print statements will refer to the architecture using the same name? ;)

bin/package.js
```diff
@@ -533,6 +536,40 @@ function buildLinux (cb) {
...
   }
 
+  function packageRpm (filesPath, destArch, cb) {
+    // Create .rpm file for RedHat-based platforms
+    console.log(`Linux: Creating ${destArch} rpm...`)
+
+    // Linux convention for RedHat based 'x64' is 'x86_64'
+    if (destArch === 'x64') {
+      destArch = 'x86_64'
+    }
```
feross: Same, can you move this above the `console.log`?

bin/package.js
hicom150: Sure!

feross:
Ahh, I was getting a "Cannot find a valid baseurl for repo: base/7/x86_64" error which I think means yum couldn't reach the internet. Weird that it needs to do that to install a local `.rpm` file. Anyway, I resolved that. Now I get:

```
oaded plugins: fastestmirror, langpacks
Examining /home/feross/Desktop/webtorrent-desktop-0.21.0-1.x86_64.rpm: webtorrent-desktop-0.21.0-1.x86_64
Marking /home/feross/Desktop/webtorrent-desktop-0.21.0-1.x86_64.rpm to be installed
Resolving Dependencies
--> Running transaction check
---> Package webtorrent-desktop.x86_64 0:0.21.0-1 will be installed
--> Processing Dependency: (kde-cli-tools or kde-runtime or trash-cli or glib2 or gvfs-client) for package: webtorrent-desktop-0.21.0-1.x86_64
Loading mirror speeds from cached hostfile
 * base: mirror.sjc02.svwh.net
 * extras: mirror.sjc02.svwh.net
 * updates: mirror.sjc02.svwh.net
base                                                     | 3.6 kB     00:00     
extras                                                   | 2.9 kB     00:00     
updates                                                  | 2.9 kB     00:00     
updates/7/x86_64/primary_db                                | 1.1 MB   00:00     
--> Processing Dependency: libXScrnSaver for package: webtorrent-desktop-0.21.0-1.x86_64
--> Running transaction check
---> Package libXScrnSaver.x86_64 0:1.2.2-6.1.el7 will be installed
---> Package webtorrent-desktop.x86_64 0:0.21.0-1 will be installed
--> Processing Dependency: (kde-cli-tools or kde-runtime or trash-cli or glib2 or gvfs-client) for package: webtorrent-desktop-0.21.0-1.x86_64
--> Finished Dependency Resolution
Error: Package: webtorrent-desktop-0.21.0-1.x86_64 (/webtorrent-desktop-0.21.0-1.x86_64)
           Requires: (kde-cli-tools or kde-runtime or trash-cli or glib2 or gvfs-client)
 You could try using --skip-broken to work around the problem
 You could try running: rpm -Va --nofiles --nodigest
```

hicom150:
After some investigation, I found that `electron-installer-redhat` has the [requirement](https://github.com/electron-userland/electron-installer-redhat#requirements) of having `rpm > 4.13` and that is why we can install successfully in `Fedora` and fails in `CentOS` 😓

I don't know what next steps should we take... Any suggestions?

mathiasvr:
We could just make a notice that it is a requirement to have `rpm > 4.13` to use the installer?
I don't think it's worth it to try and support older versions unless someone would like to do it.

I don't know much about snap packages, but is there any chance that would work as an alternative?

feross:
> We could just make a notice that it is a requirement to have `rpm > 4.13` to use the installer?

Yeah, this sounds good to me. Right now there's no .rpm support at all, so this is an improvement :)

We should probably make a page with detailed installation options for all the different OSes. Even on Mac, it's possible to install via Homebrew casks, and there's lots of options on Linux. I made an issue about this: https://github.com/webtorrent/webtorrent-desktop/issues/1706

PR #1695, opened 2019-09-15
Add buble compiler target chrome version

hicom150:

[ ] Documentation update
[ ] Bug fix
[ ] New feature
[x] Other, please explain:

Given that we are currently in `electron@6.0.9` and that in that release `chromium` is included at version `76`, we can tell `buble` to target this and get rid of all the unnecessary transformations and created a more optimized build 💪

Note: `buble` highest target version for chrome is `71` so this is the one included in this PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,7 +96,7 @@
     "url": "git://github.com/webtorrent/webtorrent-desktop.git"
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


feross:
Thanks @hicom150!

PR #1698, opened 2019-09-15
Use babel instead of buble

feross:

[ ] Documentation update
[ ] Bug fix
[x] New feature
[ ] Other, please explain:

Inspired by @hicom150's PR https://github.com/webtorrent/webtorrent-desktop/pull/1695, I swapped out `buble` for `babel`.

This way we can target just the JSX transformation. We don't use any other cutting edge JS features that require compilation, nor should we in the future.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,12 +4,794 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@babel/cli": {
+      "version": "7.6.0",
+      "resolved": "https://registry.npmjs.org/@babel/cli/-/cli-7.6.0.tgz",
+      "integrity": "sha512-1CTDyGUjQqW3Mz4gfKZ04KGOckyyaNmKneAMlABPS+ZyuxWv3FrVEVz7Ag08kNIztVx8VaJ8YgvYLSNlMKAT5Q==",
+      "dev": true,
+      "requires": {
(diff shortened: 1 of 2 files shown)


mathiasvr:
Is this faster than buble?

feross:
1.59s for babel vs 0.82s for buble. I think I'd rather be use `babel` just to be consistent with the majority of the ecosystem, even though it takes a bit longer to build.

mathiasvr:
I’m leaning more towards buble because of speed, but I can see it doesn’t make big difference here.

What exactly becomes more consistent?
I mean is there some obvious drawbacks to using buble?

feross:
I don't feel super strongly, but babel has these advantages:

- Most folks in the React ecosystem use it
- Lets us target just the transformations we want to apply. In this case, we transform only the JSX and don't touch any other part of the code.

mathiasvr:
Applying JSX transformations exclusively sounds great, but in this case I prefer buble simply because it's faster.

I'm fine either way.

hicom150:
Given that `buble` seems to change to `maintenance only` mode https://github.com/bublejs/buble/issues/220 I think that it is the correct time to make the transition to `babel` 😅

feross:
Thanks for the review, @hicom150. I just rebased this and merged it!

Issue #1652, opened 2019-08-15
Error in preferences page path-selector when there is no default value


hicom150:
**What version of this package are you using?**
Latest master commit https://github.com/webtorrent/webtorrent-desktop/commit/afd3b84b406921ecee049c8aeb0d7dc55bdaa76d

**What operating system, Node.js, and npm version?**
OSX Yosemite (10.10.5), Node.js v8.16.0, npm 6.10.2

**What happened?**
In preferences page when you try to push `CHANGE` button in `Folder to watch` or you try to push  `CHANGE` button in `External player` an error message is throw in debug console and dialog does not open

```
Error: Could not call remote method 'showOpenDialog'. Check that the method signature is correct. Underlying error: Default path must be a string
```

**What did you expect to happen?**
I expect the dialog to open

**Are you willing to submit a pull request to fix this bug?**
Sure! 😎

hicom150:
Should be fixed in #1653 😉

feross:
Fixed by https://github.com/webtorrent/webtorrent-desktop/pull/1498

PR #1702, opened 2019-09-18
Set default externalPlayerPath

314eter:

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature
- [ ] Other, please explain:

PR #1498 fixed issue #1652 for `downloadPath` and `torrentsFolderPath`, but not for `externalPlayerPath`. I changed the default value to `""`.

I handled existing config files that have `externalPlayerPath = null` by changing this to `""` when loading the config file. It's not a beautiful solution, but otherwise they have to edit it by hand.

### DIFF
--- a/src/renderer/lib/state.js
+++ b/src/renderer/lib/state.js
@@ -119,7 +119,7 @@ function setupStateSaved () {
       downloadPath: config.DEFAULT_DOWNLOAD_PATH,
       isFileHandler: false,
       openExternalPlayer: false,
-      externalPlayerPath: null,
+      externalPlayerPath: '',
       startup: false,
       soundNotifications: true,
       autoAddTorrents: false,
@@ -207,6 +207,8 @@ function load (cb) {
         onSavedState(err)
         return
       }
+    } else if (saved.prefs.externalPlayerPath == null) {
+      saved.prefs.externalPlayerPath = ''
     }
     onSavedState(null, saved)
   })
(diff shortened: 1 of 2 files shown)


mathiasvr:
Hi, thanks for looking into this.

I wonder how it can be partially fixed, maybe it does not work for the other paths as well?

I think we should allow externalPlayerPath to still be null, and handle the cases where it is not allowed.
If this is the same error as #1652, I think we should add an appropriate null check in the path-selector component instead.

314eter:
The other paths are not set to `null` when a new config file is created.

mathiasvr:
I see, so maybe setting it to undefined fixes it as well?
I still think we should handle it in the path-selector though, I can look into it later, unless you’re up for it? 😉

mathiasvr:
Okay, I have opened #1704 to handle the null's in the `path-selector` component.

However, we should still consider changing the default externalPlayerPath to a string type.

src/renderer/lib/state.js
```diff
@@ -207,6 +207,8 @@ function load (cb) {
         onSavedState(err)
         return
       }
+    } else if (saved.prefs.externalPlayerPath == null) {
+      saved.prefs.externalPlayerPath = ''
```
feross: The correct way to handle old configuration values is by adding an entry to the `migrations.js` file to "upgrade" the user's config when they update to a new version of the app. This way, we can handle the migration in one place. Then, throughout the rest of the codebase we can just assume that the configuration file has the new format. So, in this case, we could just assume that `saved.prefs.externalPlayerPath` will always be a string going forward :)

314eter: I added the migration, assuming the next version will be 0.21.1.

PR #1707, opened 2019-09-19
Add --no-sandbox option to desktop files

### LINKED ISSUE #1703: FATAL:setuid_sandbox_host.cc(157)
**What version of this package are you using?**
0.21 deb
**What operating system, Node.js, and npm version?**
kali linux, v12.10.0 , 6.10.3
**What happened?**
imagen
which will start the program normally
**Are you willing to submit a pull request to fix this bug?**
no

hicom150:

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

Add `--no-sandbox` option to launch scripts (Desktop files) to temporary fix #1703 issue.

Check on different platforms. I've checked successfully in Ubuntu 18.04.3 LTS and Debian 8 Jessie

### DIFF
--- a/static/linux/webtorrent-desktop.ejs
+++ b/static/linux/webtorrent-desktop.ejs
@@ -5,7 +5,7 @@ Name=<%= productName %>
 <% if (genericName) { %>GenericName=<%= genericName %><% } %>
 <% if (description) { %>Comment=<%= description %><% } %>
 Icon=<%= name %>
-<% if (name) { %>Exec=<%= name %> %U<% } %>
+<% if (name) { %>Exec=<%= name %> --no-sandbox %U<% } %><%/*HACK: --no-sandbox fixes an Electron 6 bug. See: #1703*/%>
 Terminal=false
 Actions=CreateNewTorrent;OpenTorrentFile;OpenTorrentAddress;
 <% if (mimeType && mimeType.length) { %>MimeType=<%= mimeType.join(';') %>;<% } %>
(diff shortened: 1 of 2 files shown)


feross:
Did you confirm that it didn't work before this change and then it worked with the change?

static/linux/webtorrent-desktop.ejs
```diff
@@ -5,7 +5,7 @@ Name=<%= productName %>
 <% if (genericName) { %>GenericName=<%= genericName %><% } %>
 <% if (description) { %>Comment=<%= description %><% } %>
 Icon=<%= name %>
-<% if (name) { %>Exec=<%= name %> %U<% } %>
+<% if (name) { %>Exec=<%= name %> --no-sandbox %U<% } %>
```
feross: Can you add a comment after each line with `--no-sandbox` to note that this is a temporary hack? Something like:

```
<!-- HACK: --no-sandbox fixes an Electron 6 bug. See: #1703 -->
```

hicom150:
> Did you confirm that it didn't work before this change and then it worked with the change?

Yes, I've been able to reproduce the whole issue and the fix following the steps described [here](https://github.com/webtorrent/webtorrent-desktop/issues/1703#issuecomment-533295632).

static/linux/webtorrent-desktop.ejs
hicom150: Sure! 👍

PR #1709, opened 2019-09-23
Implement clear completed torrents

lewisl9029:

[ ] Documentation update
[ ] Bug fix
[X] New feature
[ ] Other, please explain:

Adds the following new button to the header: 

image

When clicked, it will pop up the following confirmation dialog: 

image

Once confirmed, it will remove all torrents from the list where progress is 1 (indicating torrent has finished downloading).

I've found this to be an indispensable quality of life feature for keeping the torrent list tidy, much more convenient than hunting down the X button for every torrent on the list. 

Happy to make any changes necessary to get this feature into core in some shape or form.

- UX feedback
- Testing suggestions

### DIFF
--- a/src/renderer/components/clear-completed-modal.js
+++ b/src/renderer/components/clear-completed-modal.js
@@ -1,0 +1,28 @@
+const React = require('react')
+
+const ModalOKCancel = require('./modal-ok-cancel')
+const { dispatch, dispatcher } = require('../lib/dispatcher')
+
+module.exports = class ClearTorrentModal extends React.Component {
+  render () {
+    const message = 'Are you sure you want to clear all completed torrents from the list?'
+    const buttonText = 'CLEAR'
+
+    return (
+      <div>
+        <p><strong>{message}</strong></p>
+        <ModalOKCancel
+          cancelText='CANCEL'
+          onCancel={dispatcher('exitModal')}
+          okText={buttonText}
+          onOK={handleRemove}
+        />
+      </div>
+    )
+
+    function handleRemove () {
+      dispatch('clearCompletedTorrents')
+      dispatch('exitModal')
+    }
+  }
+}
--- a/src/renderer/components/header.js
+++ b/src/renderer/components/header.js
@@ -32,6 +32,9 @@ class Header extends React.Component {
         <div className='nav right float-right'>
           {this.getAddButton()}
         </div>
+        <div className='nav right float-right'>
+          {this.getClearCompletedButton()}
+        </div>
       </div>
     )
   }
@@ -52,6 +55,20 @@ class Header extends React.Component {
         onClick={dispatcher('openFiles')}
       >
         add
+      </i>
+    )
+  }
+
+  getClearCompletedButton () {
+    const state = this.props.state
+    if (state.location.url() !== 'home') return null
+    return (
+      <i
+        className='icon add'
+        title='Clear completed torrents'
+        onClick={dispatcher('confirmClearCompletedTorrents')}
+      >
+        clear_all
       </i>
     )
   }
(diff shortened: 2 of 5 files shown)


dsernst:
Thoughts:

### 1) I'm personally unclear if I'd use this (despite regularly clearing out my list). 

That hesitation makes me reluctant to want more UI controls. I *love* how clean the current UI is.

You said:

> I've found this to be an indispensable quality of life feature for keeping the torrent list tidy,

Do you mean in other programs? In your own build? Just from a hypothetical point of view? I love your enthusiasm, just trying to understand better 🙂

Curious— how many torrents are you typically clearing out with this? If it was like 10+ that had built up, then I can see how this all-at-once mechanisms would be a lot easier. But in my personal experience (using this program regularly for the last few years), I rarely have more than a single or maybe couple of items active at once, so I'm usually tidying down from 1 to 0, thus negating any benefit from a "Clear All" function. I don't mean to come across as argumentative, just sharing my experience, and I'd love to understand your context better. 🙂



### 2) I don't understand this initial icon choice. 



It looks like a hamburger menu icon in italics mode? 😄 

Can you explain the thinking behind that choice? Are there other programs that use that icon as a "Clear All" button?

The immediate reference that comes to mind for me is Safari's written out "Clear" button in its download menu:




### 3) Could this button be hidden by default?

And then only become visible when there are completed torrents that can be cleared? That would help me feel better about it. 🙂


----

Awesome that you already submitted a working implementation. 🙌 Hope my feedback/questions are helpful.

lewisl9029:
@dsernst Thank you for the detailed feedback! And apologies for the _really_ late follow up. I completely forgot about this PR until earlier today when the 0.23 release showed up on HN. 🙂 

Let me respond to each part separately:

1) I have been running my own fork of WebTorrent Desktop with this feature for literally years now, and use this feature on a daily basis to clear out a handful of torrents at a time (usually 3-5).

2) The icon choice didn't have too much thought behind it, other than the fact that it's literally named `clear_all` in the [Material icon set](https://material.io/resources/icons/?search=n&icon=clear_all&style=baseline) 😅, and I vaguely recall seeing it in some old version of Android for clearing notifications at some point (they've since replaced it with a "Clear all" text button though, probably for good reason). Definitely not married to the choice, and happy to take suggestions.

3) Yes, happy to implement that if people feel it's the best way forward. One small concern I have is it might be a bit jarring for buttons to just randomly appear and disappear from the top bar (none of the other buttons in the bar currently do this). A middle ground could be to disable it when no torrents can be cleared instead, making it less visible? This would be similar to the back and forward buttons on the left side being disabled when not available. Let me know what you think of this.

dsernst:
1. Nice!

2. I see. Maybe that icon is like a stack of items, as they are being removed? Like a pile of magazines, being pushed over? 😅 I guess I get it now, but yeah not sure that’d be super intuitive for people that didn’t help build the feature. It is nice that it’s the same size as other icons though, unlike writing out a whole word like “Clear” or “Clear All”.

3. Yeah. Interestingly the Safari downloads button only appears when you have any items to clear out, and I’ve personally never found it jarring. But on the other hand (agreeing with you), how often are people going to be looking at the list of torrents if there’s nothing on it, and is that even really a state we care to really add design edge cases for? Maybe not worth it.

DiegoRBaquero:
The conflict shows there's already delete all torrents, do we still want this PR?

lewisl9029:
It's not quite the same thing since this only removes completed torrents as opposed to all torrents regardless of progress, but since the other was merged first, I don't think we need both.

Unfortunately I've since moved on to using a seedbox + remote client, so don't have a personal stake in this PR anymore. Going to close it out.

Issue #1668, opened 2019-09-05
Got an "Audio codec unsupported" from time to time


Borewit:
**What version of this package are you using?**
PR: #1666, but I have seen it before.

**What operating system, Node.js, and npm version?**
OS: Windows 10
Node: v10.13.0
NPM: 6.9.0

**What happened?**
I cannot play a FLAC file from time to time. Got an error an error: 'Audio codec unsupported'
![image](https://user-images.githubusercontent.com/23255260/64317582-75468a00-cfb8-11e9-8ded-8841b06b52ad.png)

**What did you expect to happen?**
My computer to explode, thank goodness that was not the case.

**Are you willing to submit a pull request to fix this bug?**
I will have a look

**Reproduce**
Torrent: [Various Artists - 2008 - netBloc Vol 13 (color in a world of monochrome) {BSCOMP0013} [FLAC].zip](https://github.com/webtorrent/webtorrent-desktop/files/3577804/Various.Artists.-.2008.-.netBloc.Vol.13.color.in.a.world.of.monochrome.BSCOMP0013.FLAC.zip)

Click on playing tracks nervously.

feross:
The way that we detect unplayable media seems error prone, but it's unclear how to improve it. This seems like a false positive maybe caused by timing?

Borewit:
> This seems like a false positive maybe caused by timing?

It does seems to be related to timing indeed.

hicom150:
I haven't been able to reproduce this issue 😓

> The way that we detect unplayable media seems error prone, but it's unclear how to improve it.

Maybe we can do some experiment with `HTMLMediaElement.play() Promise` based on what it is explained in [this](https://developers.google.com/web/updates/2016/03/play-returns-promise) article.

hicom150:
I add #1711 to try to fix this issue 😉

Issue #759, opened 2016-08-04
WebTorrent Desktop can't play media files that don't contain audio


FluorescentHallucinogen:
**What version of WebTorrent Desktop?** (See the 'About WebTorrent' menu)

0.9.0 (0.95.2).

**What operating system and version?**

Microsoft Windows 7.

**What did you do?**

Play media files that contain only video stream and do not contain audio stream.

**What did you expect to happen?**

WebTorrent Desktop successfully plays the media files (video without audio).

**What actually happened?**

WebTorrent Desktop shows "Audio codec unsupported" message.

feross:
Ahh, yes. This is an artifact of the way that we detect that audio is not working. If we see that 0 bytes of audio are decoded, then we assume it's an error. 

This is not ideal. cc @dcposch

mathiasvr:
It would be nice if we could use the [HTMLMediaElement.audioTracks](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/audioTracks) property, but it isn't widely supported yet. Here is an issue tracking it for chromium https://bugs.chromium.org/p/chromium/issues/detail?id=249427.

dcposch:
@feross ya, that sucks. I see two ways to fix it:
1. Wait for Chromium to add support for `audioTracks`
2. Parse the video stream in JS, like https://github.com/jhiesey/videostream .

Problem w/ #2 is we have to write code for each container type (MP4, MKV, etc), as far as I can tell.

scanlime:
I ran into this bug today. Here's an example video, a short CC-licensed 4K animation with no audio track:

magnet:?xt=urn:btih:ebd69f7d7a9b7bd61c7f3afa49f7a8c5fefb2620&dn=Zen+Photon+Garden%2C+Expanding+Particles+%2312+(4K)+2304p.mp4&tr=https%3A%2F%2Fdiode.zone%2Ftracker%2Fannounce&tr=wss%3A%2F%2Fdiode.zone%3A443%2Ftracker%2Fsocket&tr=wss%3A%2F%2Ftracker.btorrent.xyz&tr=wss%3A%2F%2Ftracker.fastcast.nz&tr=wss%3A%2F%2Ftracker.openwebtorrent.com&ws=https%3A%2F%2Fdiode.zone%2Fstatic%2Fwebseed%2F6f111ebd-d595-4b01-97b1-903d9a694665-2304.mp4

hicom150:
[...]

PR #1711, opened 2019-09-25
Improve codec unsupported detection

hicom150:

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

It seems that from time to time we get some "codec unsupported" false positive. This PR changes unsupported codec detection to use  `HTMLMediaElement.play() Promise` based on what it is explained in [this](https://developers.google.com/web/updates/2016/03/play-returns-promise) article.

Check that we don't get false positives caused by timing issues.


feross:
Wow, can it really be this simple? If this actually works better, let's merge it. Haven't had time to test it.

src/renderer/pages/player-page.js
```diff
@@ -56,6 +54,14 @@ function renderMedia (state) {
       mediaElement.pause()
     } else if (!state.playing.isPaused && mediaElement.paused) {
       mediaElement.play()
+        .then(() => {
+          dispatch('mediaSuccess')
+        })
+        .catch((err) => {
```
Borewit: Maybe better to comply with: _[use promise then() instead of catch() with callbacks](https://github.com/webtorrent/webtorrent-desktop/pull/1679)_, to prevent we handle exception of the `dispatch` as pointed out by master @feross 😉.

Borewit:
Great work @hicom150.

src/renderer/pages/player-page.js
```diff
@@ -56,6 +54,14 @@ function renderMedia (state) {
...
+          if (err.name === 'NotSupportedError') {
+            dispatch('mediaError', 'Codec unsupported')
+          }
```
Borewit: Maybe remove the condition and display any error?

src/renderer/pages/player-page.js
hicom150: Sure! 👍

src/renderer/pages/player-page.js
hicom150: Yes! In our case all the road lead to `Codec unsupported` 😅

src/renderer/pages/player-page.js
```diff
@@ -54,14 +54,10 @@ function renderMedia (state) {
...
-          dispatch('mediaSuccess')
-        })
-        .catch((err) => {
-          if (err.name === 'NotSupportedError') {
-            dispatch('mediaError', 'Codec unsupported')
-          }
-        })
+        .then(
+          () => dispatch('mediaSuccess'),
+          () => dispatch('mediaError', 'Codec unsupported')
```
Borewit: Maybe?
```js
        .then(
          () => dispatch('mediaSuccess'),
          err => dispatch('mediaError', err.name === 'NotSupportedError' ? 'Codec unsupported' : `${err.name}: ${err.message}`)
        )
```
I put it on a branch: https://github.com/webtorrent/webtorrent-desktop/tree/issue-1711-improve-error-handling

For some reason I would send PR to your clone, I think select your fork as the base. Maybe a bug in GitHub, to many forks of this repo.

hicom150: I think that we should dispatch `mediaError` when error type is `NotSupportedError` (original code) or when any error case (like the latest code change).

Given that the possible [exceptions](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/play#Exceptions) are `NotSupportedError` or `NotAllowedError` I wouldn't show this last one to the final user 😅

Borewit: > I wouldn't show this last one to the final user

why not?

Borewit: It also says _other exceptions may be reported,_. I suppose there are not very likely, But it may be good user gets an error if something odd happens.

Borewit: With the current code, in the unlikely event of an error named `NotAllowedError`, or any other exception, it will prompt the user with 'Codec unsupported' right?

hicom150: > With the current code, in the unlikely event of an error named NotAllowedError, or any other exception, it will prompt the user with 'Codec unsupported' right?

Yes that is right 😅 Although it is not 100% correct for the final user, the action should be the same: **open an external player to be able to play the file**, otherwise it would see the message `The request is not allowed by the user agent or the platform in the current context, possibly because the user denied permission.` which I found a bit confusing for the user as it doesn't give any "call to action" to solve the problem.

On the other hand, it is true that we should log the proper error code, so we can debug future problems easily 😉

What do you think @Borewit ?

RecoX:
I believe the current implementation using the audio/video api for manage this flows is ok, I tried in the past use this approach but I ended up leaving as is because it seems wiser anyways this pr looks simple.

I would like to add some details I found.

I put a console.log, to see that element as a lot of times is throwing me a false positive
```
  function onCanPlay (e) {
    const elem = e.target
    console.log(666, elem.webkitAudioDecodedByteCount)
    
    if (elem.readyState < HTMLMediaElement.HAVE_FUTURE_DATA) return
    if (state.playing.type === 'video' &&
      elem.webkitVideoDecodedByteCount === 0) {
      dispatch('mediaError', 'Video codec unsupported')
    } else if (elem.webkitAudioDecodedByteCount === 0) {
      dispatch('mediaError', 'Audio codec unsupported')
    } else {
      dispatch('mediaSuccess')
      elem.play()
    }
  }

```
Screen Shot 2019-10-08 at 2 37 04 PM


Both are mp3's so im not sure if this `webkitAudioDecodedByteCount` is really trustable

Borewit:
@RecoX 
> I believe the current implementation using the audio/video api for manage this flow is ok, 

With the current implementation we get false positives as described in #1668 

> I tried in the past use this approach but I ended up leaving

Why? If you had an issue with PR approach, is that still valid? Can you test the PR @RecoX ? Or do you advice to merge, it is not clear to me.

> anyways this pr looks simple

Looks very clean and simple indeed.

Borewit:
As there is some discussion, I suggest to keep the PR open for 2 day. Please use the review process fellow contributors.

Sorry for the delay @hicom150.

hicom150:
> Sorry for the delay @hicom150.

Don't worry @Borewit 😉

I find it super helpful to get feedback from other developers, I think that the code gets better when we discus and combine different point of views.

BTW, thank you for your review and great feedback @Borewit 💪

RecoX:
> @RecoX
> 
> > I believe the current implementation using the audio/video api for manage this flow is ok,
> 
> With the current implementation we get false positives as described in #1668
> 
> > I tried in the past use this approach but I ended up leaving
> 
> Why? If you had an issue with PR approach, is that still valid? Can you test the PR @RecoX ? Or do you advice to merge, it is not clear to me.
> 
> > anyways this pr looks simple
> 
> Looks very clean and simple indeed.

I tested it and it works ok, but I found a problem.

When I tried to reproduce a `wma` it throws the modal with the unsupported codec and that is ok but as you see in the console is doing that in an infinite loop so the program freezes.

https://github.com/refreex/refreex-webtorrent-desktop/blob/master/src/renderer/pages/audio-player.js

here the link of the file where i implement it just in case you want to take a look.

Screen Shot 2019-10-09 at 9 33 43 AM

hicom150:
@RecoX I can't reproduce the error case 😓 but I noticed that you didn't delete [onCanPlay](https://github.com/refreex/refreex-webtorrent-desktop/blob/master/src/renderer/pages/audio-player.js#L116) and [onError](https://github.com/refreex/refreex-webtorrent-desktop/blob/master/src/renderer/pages/audio-player.js#L113) as it is in this PR code 😅

RecoX:
> @RecoX I can't reproduce your error case 😓 but I noticed that you didn't delete [onCanPlay](https://github.com/refreex/refreex-webtorrent-desktop/blob/master/src/renderer/pages/audio-player.js#L116) and [onError](https://github.com/refreex/refreex-webtorrent-desktop/blob/master/src/renderer/pages/audio-player.js#L113) as it is in this PR code 😅

You don't see them because i didn't push the changes, probably is because the player in my fork is in the main window instead of the player window. Anyways is strange that that occurs to me with the new code and with the older no.

mathiasvr:
Video with AC3 audio plays silently (like Chrome), and does not give the user the option to open in an external player.
I also had a false positive (saw the `Codec unsupported` error on a H264 file), which worked fine all other times I tried.

hicom150:
> Video with AC3 audio plays silently (like Chrome), and does not give the user the option to open in an external player.
I also had a false positive (saw the Codec unsupported error on a H264 file), which worked fine all other times I tried.

Ups! I didn't expect that result 😓

In order to help the debugging process can we share the test files where we get the unexpected behaviors? @mathiasvr @RecoX 

Here I share the some files a what should be the expected result:

- [Video: mpeg2video. Audio: ac3](https://send.firefox.com/download/6ade81cef88e9397/#sTLqQZMGQZ5qFaQ612rYrA)
**expected result**: `Codec Unsupported` dialog.
screenshot2

- [Video: h264 Audio: none](https://send.firefox.com/download/c629bf2287e1d275/#uj_xQmUe2CF15elVZR6UAQ)
**expected result**: Plays video normally.
screenshot4

mathiasvr:
@hicom150 I just went ahead and generated a bunch of different samples with various video/audio codecs, because why not? 😃

https://drive.google.com/open?id=1ACw6hOooQNfc-oniyL0Gp-z4OSq-5Mc7

Before this PR, supported a video file (such as h264) with AC3 audio codec displays 'Audio codec not supported', and after this PR it plays the video with audio muted. It would be nice if we could still detect unsupported audio.

hicom150:
After some further testing with the awesome @mathiasvr video smaples 😉 it seems that `HTMLMediaElement.play() Promise` is not as fine grained as the previous implementation 😭

So unless someone comes with a better solution, I think that the only proper option is to try to optimize the previous `onCanPlay` function 😅

RecoX:
> After some further testing with the awesome @mathiasvr video samples 😉 it seems that `HTMLMediaElement.play() Promise` is not as fine grained as the previous implementation 😭
> 
> So unless someone comes with a better solution, I think that the only proper option is to try to optimize the previous `onCanPlay` function 😅

Agreed :) +1

mathiasvr:
@hicom150 Did you see other problems than muted audio? Otherwise I think the `play() Promise` could still be a good solution, if we can detect the audio problem (maybe using canPlay) as well.

hicom150:
I changed the approach and instead of using  `HTMLMediaElement.play() Promise` I try with experimental `videoTracks` and `audioTracks` which describe the number of available (correctly decoded) tracks.

Having this, I see two options:

- If we use the same approach we use before, we can say that for a `video` element we expect at least one video track and one audio track and for an `audio` element we expect at least one audio track. (this has the caveat that videos with no audio track are treated like error #759)

- The other option could be to allow to play videos without any audio track either because it has not any audio stream or because the audio stream codec is not supported. (this has the caveat that we directly play videos without supported audio without giving the option to open in VLC)

To sum up, the advantage of this method is that we can check available tracks early in `onLoadedMetadata` event and that we don't rely on `webkitVideoDecodedByteCount` and `webkitAudioDecodedByteCount` that sometimes have timing issues.

mathiasvr:
Nice work @hicom150!
I haven’t tested it yet, but definitely sounds more stable than the current approach.
I think not supporting video with no audio is fine for now, we could consider adding a “play without audio” button on the “Audio codec not supported” error dialog? But maybe it’s a little too hacky 😉

RecoX:
> Nice work @hicom150!
> I haven’t tested it yet, but definitely sounds more stable than the current approach.
> I think not supporting video with no audio is fine for now, we could consider adding a “play without audio” button on the “Audio codec not supported” error dialog? But maybe it’s a little too hacky 😉

I feel its too hacky too. If we can't play the video as we should we simply can offer to open it with a an external player and that's it. Keep it simple.

hicom150:
After some back and forth I think that finally I'm going to merge this PR a go forward 😉
Thank you for your awesome help in this issue guys! 💪

PR #1720, opened 2019-10-07
Fix a few type errors

mathiasvr:

[ ] Documentation update
[ ] Bug fix
[ ] New feature
[x] Other, please explain:

Fixes type errors, mainly according to new Electron version.

### DIFF
--- a/src/main/windows/main.js
+++ b/src/main/windows/main.js
@@ -32,7 +32,6 @@ function init (state, options) {
 
   const win = main.win = new electron.BrowserWindow({
     backgroundColor: '#282828',
-    backgroundThrottling: false, // do not throttle animations/timers when page is background
     darkTheme: true, // Forces dark theme (GTK+3)
     height: initialBounds.height,
     icon: getIconPath(), // Window icon (Windows, Linux)
--- a/src/main/windows/webtorrent.js
+++ b/src/main/windows/webtorrent.js
@@ -13,7 +13,6 @@ const config = require('../../config')
 function init () {
   const win = webtorrent.win = new electron.BrowserWindow({
     backgroundColor: '#1E1E1E',
-    backgroundThrottling: false, // do not throttle animations/timers when page is background
     center: true,
     fullscreen: false,
     fullscreenable: false,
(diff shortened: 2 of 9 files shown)


src/main/windows/webtorrent.js
```diff
@@ -26,7 +25,8 @@ function init () {
     title: 'webtorrent-hidden-window',
     useContentSize: true,
     webPreferences: {
-      nodeIntegration: true
+      nodeIntegration: true,
+      backgroundThrottling: false // do not throttle animations/timers when page is background
```
mathiasvr: I'm not sure this was in effect before, do we actually want to disable background throttling?

hicom150: Given that it was not in inside `webPreferences` I assume it was in it's default state: `true`. So I wouldn't change it here 😅

src/main/windows/main.js
```diff
@@ -44,7 +43,8 @@ function init (state, options) {
     useContentSize: true, // Specify web page size without OS chrome
     width: initialBounds.width,
     webPreferences: {
-      nodeIntegration: true
+      nodeIntegration: true,
+      backgroundThrottling: false // do not throttle animations/timers when page is background
```
hicom150: Given that it was not in inside `webPreferences` I assume it was in it's default state: `true`. So I wouldn't change it here 😅

src/renderer/pages/player-page.js
```diff
@@ -497,7 +497,7 @@ function renderPlayerControls (state) {
       <div
         key='scrub-bar'
         className='scrub-bar'
-        draggable='true'
+        draggable
```
hicom150: I not sure but I think that we should leave it as it was according to [draggable documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/draggable)

hicom150:
Great work @mathiasvr 👍
I point just a pair of things to check 😉

src/renderer/pages/player-page.js
mathiasvr: The jsx draggable type is boolean (not string), and this is shorthand for `draggable={true}`. I believe that should be correct according to the docs.

mathiasvr:
@hicom150 Thanks for the review, I have updated the PR to remove the `backgroundThrottling: false` setting.

feross:
I actually think we want `backgroundThrottling: false` so our timers and intervals aren't throttled when the app is backgrounded.

PR #1731, opened 2019-10-22
Add experimental video stream info parser

hicom150:

[ ] Documentation update
[ ] Bug fix
[x] New feature
[ ] Other, please explain:

Add a experimental video stream container parser for `MP4` and `WebM` containers.

This would allow to get detailed stream count (audio and video) regardless if the browser is able to decode them or not. This would help to better identify not supported codecs and play video streams with no audio.

### DIFF
--- a/src/renderer/lib/stream-info.js
+++ b/src/renderer/lib/stream-info.js
@@ -1,0 +1,219 @@
+
+module.exports = {
+  getStreamInfo
+}
+
+function getStreamInfo (src) {
+  if (/\.(mp4|m4v|mov)$/.test(src)) {
+    return getMp4StreamInfo(src)
+  } else if (/\.(mkv|webm)$/.test(src)) {
+    return getWebmStreamInfo(src)
+  } else {
+    return getDefaultStreamInfo(src)
+  }
+}
+
+function getMp4StreamInfo (src) {
+  const BUFFER_SIZE = 16384
+  const streamInfo = { video: 0, audio: 0 }
+  const offset = { start: 0, end: 0, pointer: 0, done: Infinity }
+
+  return new Promise(function (resolve, reject) {
+    consume()
+
+    function consume () {
+      window
+        .fetch(src, {
+          headers: {
+            range: `bytes=${offset.pointer}-${offset.pointer + BUFFER_SIZE}`
+          }
+        })
+        .then(response => response.arrayBuffer())
+        .then(chunk => {
+          offset.start = offset.pointer
+          offset.end = offset.pointer + BUFFER_SIZE
+
+          parseAtom(offset, chunk)
+        })
+    }
+
+    function parseAtom (offset, chunk) {
+      // exit when we finish parsing moov atom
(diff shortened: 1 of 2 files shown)


Borewit:
Oh wow, looks very familiar, EBLM and MPEG-4 parsing, well done.

Although [music-metadata](https://github.com/Borewit/music-metadata) is intended for audio files, it does work with principle parse MP4 video files. It may be interesting to put the required functionality in there. 

If it not suitable to implement that functionality in [music-metadata](https://github.com/Borewit/music-metadata), it may be better to move it in a dedicated parser module, with it's own unit tests. Using that option, I would suggest to have a look to [strtok3](https://github.com/Borewit/strtok3) which is very handy for parsing. Maybe isolate the MPEG-4 parser?

[Here you can find a similar simplistic EBML micro parser](https://github.com/Borewit/file-type/blob/2264fb042a6658b07dd7d5e0277ed1559800b800/index.js#L649-L708) using [strtok3](https://github.com/Borewit/strtok3). 

These are my initial thoughts, please allow me to have closer look another time.

hicom150:
@Borewit thank you very much for your feedback! 👍

My idea was to write minimum code possible to read the number of video and audio tracks from the file and compare with the audio and video tracks available, so we can decide if any tracks is not playable by our player 😅

I don't know if this functionality could fit in `music-metadata` in that case it would be awesome as it would be already included in the dependencies 😉
I fit does not fit, I could try to move to a dedicated module and of course I will give a look [strtok3](https://github.com/Borewit/strtok3) 👍

Thank you again for giving your time to review this 💪

Borewit:
Thank you for your feedback as well Julen.

I will get music-metadata up to speed by adding [EBML](https://en.wikipedia.org/wiki/Extensible_Binary_Meta_Language) (getting an initial implementation out supporting [Matroska](https://en.wikipedia.org/wiki/Matroska) audio & [WebM](https://en.wikipedia.org/wiki/WebM) audio files): [music-metadata#356](https://github.com/Borewit/music-metadata/issues/356).

The current music-metadata API does not take the concept of multiple streams into account (internally it does). The most common use case of using music-metadata is, I toss my audio file in, and I want to get the title, artist track-number, duration back. I need a find a way expose specific information, without causing to much harm to the simple use cases and backwards compatibility.

I captured the requirement: [music-metadata#357](https://github.com/Borewit/music-metadata/issues/357)

Borewit:
@hicom150  Please have a look to [this unit test](https://github.com/Borewit/music-metadata/blob/feature/stream-info/test/test-streaminfo.ts). 
It is testing the new `format.streamInfo` API. Is that the kind of input you need?

Borewit:
@hicom150, can you check if [`metadata.trackInfo`](https://github.com/Borewit/music-metadata#metadatatrackinfo) in [music-metadata version  6.3.0](https://github.com/Borewit/music-metadata/releases/tag/v6.3.0) gives you the input you need. It's an initial attempt, I'd like to here whatever your are missing.

Issue #1736, opened 2019-10-30
About WebTorrent shows half window


akaustav:
**What version of WebTorrent Desktop are you using?**
0.21.0

**What operating system and version?**
Windows 10

**What happened?**
When I click on Help > About WebTorrent, I see a window which seems to display only the top half of the content. So, I can only see the application icon, but not the text (name, version number, etc) - like this:
![image](https://user-images.githubusercontent.com/11030959/67826959-83ca9280-fa8b-11e9-9bd0-84d244233a96.png)

**What did you expect to happen?**
To view the full information about the application. Similar to what was showing until the previous version [`v0.20.0`](https://github.com/webtorrent/webtorrent-desktop/releases/tag/v0.20.0).
![image](https://user-images.githubusercontent.com/11030959/67827963-e5403080-fa8e-11e9-9394-43f4e832cef0.png)

**Are you willing to submit a pull request to fix this bug?**
Not sure. I am a novice in Electron. However, it was fine in the previous version [`v0.20.0`](https://github.com/webtorrent/webtorrent-desktop/releases/tag/v0.20.0). I can try tracing the change log to find which commit(s) might have caused this.

akaustav:
**Here is a log of what I have found so far:**

- Compared commits between `0.20.0` and `0.21.0` tags.
- From [electron documentation](https://electronjs.org/docs/api/browser-window#class-browserwindow), it looks like a window's height is controlled by the optional `height` property.
- The `height` property of the 'About' window is set at `170`.
https://github.com/webtorrent/webtorrent-desktop/blob/d7783749ca155e196cd5edbb5fae9f8c83414945/src/main/windows/about.js#L18
- That property has not changed since 2016-05-27
https://github.com/webtorrent/webtorrent-desktop/blame/d7783749ca155e196cd5edbb5fae9f8c83414945/src/main/windows/about.js#L18
- Noticed that `webtorrent-desktop@0.20.0` was using `electron@1.6.16`:
https://github.com/webtorrent/webtorrent-desktop/blob/fbb3fa6a270c4f41042ad8ce2f8c28a15d7e2007/package.json#L57
- And `webtorrent-desktop@0.21.0` is using `electron@~6.0.9`:
https://github.com/webtorrent/webtorrent-desktop/blob/d7783749ca155e196cd5edbb5fae9f8c83414945/package.json#L56
[...]

PR #1737, opened 2019-10-30
Fix: Increase height of 'About' window

akaustav:

- [x] Bug fix

Increased the height of the **About WebTorrent** window which shows when you click **Help** > **About WebTorrent**.

#1736

- Tested in Windows 10 (64 bit) only.
- Need help testing in other supported platforms.
- Need instructions/help to update unit tests, and/or end-to-end tests.

**Screenshot**
image


mathiasvr:
This about window is only used on Windows and Linux, on macOS the native `about` [role](https://electronjs.org/docs/api/menu-item#roles) is used instead, and looks like this:

Screenshot 2019-10-30 at 12 34 30

I see the Copyright notice is missing for the other platforms, it would be great if you could add it as well @akaustav.

akaustav:
Cool, thanks @mathiasvr. I verify that the copyright information is available in the window now. However, `220` height was not sufficient to display the copyright line. So, I had to increase the height even further. After changing the height to `250`, it looks like this in Windows 10:

image

akaustav:
Also, I see that the lines related to Help menu in Windows and Linux:
https://github.com/webtorrent/webtorrent-desktop/blob/0198d7493d1191be7ee8617ef71f2bc12037028e/src/main/menu.js#L441-L450

Also, like @mathiasvr said, here is the code snippet for Mac (Darwin):
https://github.com/webtorrent/webtorrent-desktop/blob/0198d7493d1191be7ee8617ef71f2bc12037028e/src/main/menu.js#L340-L347

However, I fail to understand what controls the height of the about dialog in **macOS**. Does **macOS** automatically adjust the window width and height to accommodate the contents? Or is there a setting in a configuration file somewhere?

Also, would you think it is a good idea to keep the dimensions of the About window/dialog box consistent across multiple platforms?

mathiasvr:
@akaustav On mac the about window is populated with info from the packaging process, and I believe you are correct that the dimensions are automatically adjusted.

I think dimensions should be the same across platforms, but it's more important to have a consistent look.
I tried displaying the Windows/Linux about window on mac and I get this:

Screenshot 2019-10-31 at 17 55 05

It's weird that there is such a big difference between macOS and Windows? However, as long as it looks good in Windows and Linux it's fine since we don't use this window on mac.

hicom150:
This fix looks great, but I found two small issues 😅

- Copyright path should point to `build` instead of `src` otherwise it will not be correctly loaded after packing
- After upgrading `master` electron to `electron@7` there is a small issue to remove the window menu https://github.com/electron/electron/issues/21088 so a I used a small _hack_ 😉

I will add this corrections in the following commit 👍

hicom150:
Good work @akaustav and @mathiasvr 🎉

mathiasvr:
Thanks @hicom150, nice catches 👍

PR #1751, opened 2019-11-22
Fix deprecated dialog callback functions

hicom150:

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

This PR updates dialog functions to use up to date *Sync versions. 

When updating to `electron@7.0.0` dialog with callback functions have been deprecated, so I've updated with the up to date ones.

### DIFF
--- a/src/main/dialog.js
+++ b/src/main/dialog.js
@@ -61,11 +61,10 @@ function openFiles () {
       properties: ['openFile']
     }
   setTitle(opts.title)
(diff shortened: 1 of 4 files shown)


hicom150:
@alxhotel thank you for your review! 👍

I totally agree that we should define if we want to go full "promise style" or "sync style".
Maybe I could open an issue to talk about this 😅

In the meantime, I will merge this just to fix the bug 😉

PR #1785, opened 2020-02-01
Greenkeeper/music metadata 6.3.2

Borewit:
Prevent a (rare) infinite loop caused by an invalid ADTS-frame-header in `.mp3` or `.aac` files.

Adds experimental Matroska / mp4 stream-info which could be used in PR #1731.


Borewit:
Superseded by #1795

Issue #970, opened 2016-09-26
Remove torrent immediately, but allow Undo


DiegoRBaquero:
I'd like to make a feature request.

If you are removing many torrents from the client you have to click on the remove button in the confirmation dialog each time. It would be great if I could check a checkbox that said "Do not show this again" and just click on the X that removes the torrents.

What do you guys think?

mathiasvr:
I actually included such a checkbox in my initial PR for the confirmation modals, but it seemed more clean not to have it. Another way to solve this is by having a preference as suggested in #853.

feross:
The cleanest way to solve this, IMO, is to just remove it immediately, but to show a little "Undo" hover card at the bottom of the view for 5 seconds after deleting a torrent.

The thinking is: Do what the user wants immediately. Treat the user like an adult who knows what they're doing, so don't ask for confirmation constantly. Yet, provide a means for them to recover from mistakes.

Here's an example of how Inbox (by Google) implements this, in the bottom-left corner of the app.

<img width="505" alt="screen shot 2016-09-26 at 11 19 26 pm" src="https://cloud.githubusercontent.com/assets/121766/18862253/bb96dc84-843f-11e6-8432-7aaafc055d10.png">

DiegoRBaquero:
@feross Damn, I love that idea. Would definitely leave preferences to be important stuff and never bother the user.

feross:
@DiegoRBaquero Yeah, some things need preferences, but I think the right instinct is to push back and consider other solutions first. Once you add a preference it's really hard to remove. Let's try really hard to do the right thing, and use preferences as a last resort.

mathiasvr:
@feross I like this idea, but I wonder if it would be enough to just use undo/cmd+z then?

feross:
Hm, I think that a keyboard shortcut alone works as a first attempt. But I would want to play around with the little notification message at the bottom and see if that's helpful to the user. Maybe it's too much for a desktop app, but we should try it IMO.

Btw, for the "delete file data" option in the right click menu, we probably still want the confirmation dialog since that's not possible to reverse, and it's a less commonly used than just clicking the 'X'.

PR #1788, opened 2020-02-06
Remove torrent immediately, but allow Undo

dsernst:

[ ] Documentation update
[ ] Bug fix
[x] New feature
[ ] Other, please explain:

This is a continuation of the work started in PR https://github.com/webtorrent/webtorrent-desktop/pull/1087 to replace the `Remove a Torrent` UI modal with a Snackbar style alert + option to Undo.

Kapture 2020-02-05 at 17 26 10

The 3 other methods of removing a torrent:

  - Right click in list > `Remove Data File`
  - `Transfers` menu > `Remove All From List`
  - `Transfers` menu > `Remove All Data Files`

all continue to use the same modal as before (per discussion in https://github.com/webtorrent/webtorrent-desktop/issues/970).

https://github.com/webtorrent/webtorrent-desktop/issues/970, https://github.com/webtorrent/webtorrent-desktop/pull/1087 (abandoned PR)

👍 This is tested locally and works on my machine running OS X 10.14.6.


dsernst:
From https://github.com/webtorrent/webtorrent-desktop/issues/970#issuecomment-249908009:

> image


This PR doesn't add support for a `Ctrl-Z` style Undo hotkey. 

I wanted to get something working out the door first, & don't think the hotkey ought to be a blocker.

dsernst:
### One important thing to note:

The Undo feature currently only works for magnet links, which doesn't include the Featured torrents on the WebTorrent Desktop homepage, so I don't know how to demo that properly.

✅ I confirmed that the Undo feature does work on a tracker that supports magnet links. 

   - [ ] If someone can point me to a good public tracker w/ magnet support, I can make a quick gif demo of that.



### Even better would be to fix this up so the Undo works for torrents added via a .torrent file. 

I can work on that, but would appreciate a little advice on how best to approach. 

Currently, the locally stored `${appConfig}/Torrents/` files are still being deleted immediately.

- So one solution is to simply delay deleting the .torrent files until after the Snackbar autohides itself after 5 seconds. That ought to work, but if the Application quits before the 5 second timer, the file may never get cleaned up, which doesn't seem great.

- A different solution is to remember how the torrent was added in first place, such as via pasting in a link like https://archive.org/download/art_of_war_librivox/art_of_war_librivox_archive.torrent. Then the Undo call could go try to add that way again. But as far as I could tell, these links aren't being stored within `torrentSummary` right now, just the path to the new local .torrent file.

subins2000:
I tested this locally and the Undo option also worked for the Featured torrents in WebTorrent.

dsernst:
Fixed merge conflict

dsernst:
Best path forward, per discussion in today's Webtorrent Maintainers call:

> So one solution is to simply delay deleting the .torrent files until after the Snackbar autohides itself after 5 seconds. That ought to work, but if the Application quits before the 5 second timer, the file may never get cleaned up, which doesn't seem great.

We can add a flag like `pendingDelete` to the torrent. The UI will hide these items, so from the user's perspective it's been deleted. Then have a function like `removePendingDeletes` that goes through and actually deletes it from the state. That function can be called after the Undo's timer ends, as well as on app shutdown and app start (in case the app had crashed).

dsernst:
Yes it’s still relevant, if we want to simplify the UI (replace two click modal w/ one click undoable).

Nothing’s changed since my last comment.

- [x] This works great for .torrent files.
- [ ] Undo doesn’t work for magnet links yet. There is a relatively clear path forward.

I just haven’t taken the time to work on this lately, as I’ve been focused on other projects 😄

DiegoRBaquero:
@dsernst What's the path forward of with magnets, I'll take it on myself

dsernst:
> @dsernst What's the path forward of with magnets, I'll take it on myself

Awesome to hear. 🎉

Last I spoke about this w/ @feross, he liked [this approach](https://github.com/webtorrent/webtorrent-desktop/pull/1788#issuecomment-738954690): 

> Best path forward, per discussion in today's Webtorrent Maintainers call:
> 
> > So one solution is to simply delay deleting the .torrent files until after the Snackbar autohides itself after 5 seconds. That ought to work, but if the Application quits before the 5 second timer, the file may never get cleaned up, which doesn't seem great.
> 
> We can add a flag like `pendingDelete` to the torrent. The UI will hide these items, so from the user's perspective it's been deleted. Then have a function like `removePendingDeletes` that goes through and actually deletes it from the state. That function can be called after the Undo's timer ends, as well as on app shutdown and app start (in case the app had crashed).

Issue #1024, opened 2016-10-11
Cannot create a torrent without WebTorrent trackers


dcposch:
**What version of WebTorrent Desktop?** 0.17.2

**What operating system and version?** all

**What did you do?**

Drag any file onto the app. In the Create Torrent screen, delete all trackers from the announce list except the first one (`udp://tracker.openbittorrent.com:80`). Then, right-click the newly created torrent and select Save Torrent File As... Finally, use `parse-torrent` to inspect the resulting `.torrent` file.

**What did you expect to happen?**

You should see only one tracker in the announce list.

**What actually happened?**

You see four trackers in the announce list:

```
    "udp://tracker.openbittorrent.com:80",
    "wss://tracker.btorrent.xyz",
    "wss://tracker.fastcast.nz",
    "wss://tracker.openwebtorrent.com"
```

feross:
This seems related to https://github.com/webtorrent/webtorrent-desktop/issues/881

PR #1789, opened 2020-02-10
Remove forced webtorrent trackers.

KJ2ME:

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain: 

Remove forced webtorrent trackers.

#1024

### DIFF
--- a/src/renderer/webtorrent.js
+++ b/src/renderer/webtorrent.js
@@ -4,7 +4,6 @@ console.time('init')
 
 const crypto = require('crypto')
 const util = require('util')
-const defaultAnnounceList = require('create-torrent').announceList
 const electron = require('electron')
 const fs = require('fs')
 const mkdirp = require('mkdirp')
@@ -23,11 +22,6 @@ crashReporter.init()
 
 // Send & receive messages from the main window
 const ipc = electron.ipcRenderer
-
-// Force use of webtorrent trackers on all torrents
-global.WEBTORRENT_ANNOUNCE = defaultAnnounceList
-  .map((arr) => arr[0])
-  .filter((url) => url.indexOf('wss://') === 0 || url.indexOf('ws://') === 0)
 
 /**
  * WebTorrent version.


mathiasvr:
We need webtorrent trackers for “normal” torrents to make it into the WebRTC peer network.
I think issue #1024 is about creating a torrent file without the trackers.

KJ2ME:
Yes and the client suggest some default trackers when a new torrent is created, but is different if that is forced.

This not remove the suggested trackers, only the forced trackers when the .torrent file is created.

"global.WEBTORRENT_ANNOUNCE" only is used on the creation of the torrent file and generate the #1024 bug.

mathiasvr:
But doesn’t this also remove the trackers for torrents/magnet that are added?

KJ2ME:
No, only remove the create torrent.

I checked where is used before try make the change: https://i.imgur.com/OYC66cT.png

And just in case, just now I tried build and run the app and works fine. This is a magnet link created:

`magnet:?xt=urn:btih:6c2ea2e4d96c039de0e6c319662ccc3e586fd1c3&dn=a95jCIZOQ3s1XXJY.mp4&tr=udp%3A%2F%2Fexplodie.org%3A6969&tr=udp%3A%2F%2Ftracker.coppersurfer.tk%3A6969&tr=udp%3A%2F%2Ftracker.empire-js.us%3A1337&tr=udp%3A%2F%2Ftracker.leechers-paradise.org%3A6969&tr=udp%3A%2F%2Ftracker.opentrackr.org%3A1337&tr=wss%3A%2F%2Ftracker.btorrent.xyz&tr=wss%3A%2F%2Ftracker.fastcast.nz&tr=wss%3A%2F%2Ftracker.openwebtorrent.com`

It is using the default torrent trackers.

mathiasvr:
Right, I haven’t checked it, but if you add an existing magnet or torrent not created by Webtorrent Desktop, that does not already have the wss trackers, I think they won’t be added automatically after this change. This is important to make this app a “hybrid client”.
The ‘WEBTORRENT_ANNOUNCE‘ global is used by webtorrent (not this app).

KJ2ME:
Why it by intrusive only for be a hybrid client?

For be hybrid only need works with the 2 ways, but only when the user need, not forcing it.

mathiasvr:
Yes, I agree it makes sense to not force it, but because most torrents don't use WebRTC, adding the extra wss trackers by defaults helps bring content to WebRTC peers. It is only hybrid clients that can bridge the two peer networks, and because webtorrent is still relatively new, it makes sense to "force" the wss trackers. In the future, this may not be necessary.

However, it is a problem that the wss trackers are forced into a saved torrent file, even when the user explicitly removed them.

KJ2ME:
Ok, it is understood.

subins2000:
#1836 addresses this

PR #1795, opened 2020-03-14
Greenkeeper/music metadata 6.3.6

Borewit:
Update [v6.3.6](https://github.com/Borewit/music-metadata/releases/tag/v6.3.6).
Minor fixes, low risk.

Can we switch to [Dependabot](https://github.com/dependabot) @feross?
Would be convenient if pending PR's are automatically updated to the latest dependency.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -306,9 +306,14 @@
       }
     },
     "@tokenizer/token": {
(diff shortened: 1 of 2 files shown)


Borewit:
Thx @alxhotel

Issue #881, opened 2016-09-03
Add custom trackers to an existing torrent


fenilgandhi:
WebTorrent Version : Version 0.8.1 (0.95.2)

Operating System(s) : Windows 7 (x64) , Windows 8 (x32) , Windows 10 Insider (x64)

Well , this is a feature request . 

I have webtorrent installed on some of my computers and it works perfectly on all of them . 
But some torrents are really bad at finding trackers while other pc's on same network work properly . 
So can you add a feature of adding custom trackers , so we can add appropriate trackers to get proper peers for the streaming. 

Nice work guys ,
Thank You

dcposch:
You can already add your own trackers when creating a new torrent.

If I understand it right, you want to add trackers to existing torrents?

Do other torrent clients, like Transmission and uTorrent, support this?

Do you add custom trackers to each torrent individually, or is there some preference to announce *all* your torrents to a specific list of extra trackers?

fanilgandhi:
Transmission and uTorrent both allow private trackers to be added during the start and even later on . And in my case i only want to add private trackers when the torrent does not have proper peers so that can be found out at a later stage only. 
Also trackers are indiviual for torrents. but having them announce for all your torrents would be a new and    useful feature.

feross:
@fanilgandhi Thanks for the feature request.

Honestly, this seems like a bit of an edge case to me. I know that most torrent apps do support this, but I can't imagine that more than a few percent of our users would ever use a feature that let them modify the trackers used by an individual torrent.

That said, we're currently adding four default webtorrent (ws:// and wss://) trackers to all torrents without making this very clear to users. There is an argument to be made that we should allow users to turn this off (just from a privacy perspective), so we could add a preference called "Global Trackers" that defaults to the four webtorrent trackers, and would also incidentally satisfy this feature request as well.

Thoughts?

fanilgandhi:
Thank You , for providing time for my request. Yes, it definitely it a bit of an edge case but also very useful one . And sometimes we have torrents going on for weeks and you better know how dynamic trackers are . So it could be a nice feature for any user. Also some private high speed torrent sites change their trackers every 48 hours in which case it can be truly great.

Also love your work man , it's very nice .

mfurlend:
Personally, I think this is an important feature to have. Whenever I am downloading a torrent which has very few seeds/peers I will google the hash and add any trackers I find to it.

Great job on the app, btw.

jxoesneon:
[...]

PR #1836, opened 2020-07-16
feat: add preference to customize global trackers

subins2000:

[ ] Documentation update
[ ] Bug fix
[x] New feature
[ ] Other, please explain:

Added a new preference item, "Global Trackers". This list will be used as the default announceList instead of `create-torrent` module's list. This is done by setting the global variable `WEBTORRENT_ANNOUNCE`

image

~~Currently it requires a restart for the changes to apply. But I wonder if changes to `WEBTORRENT_ANNOUNCE` will reflect in app runtime after init ? :thinking:~~

Tracker changes will apply on pause & resume of a torrent.

#881

~~Does changes to `WEBTORRENT_ANNOUNCE` reflect in app runtime after init ? :thinking: My guess is that torrent needs to be paused and resumed for the new trackers to get applied ?~~

Changes to `WEBTORRENT_ANNOUNCE` will reflect to future webtorrent `client.add` calls. So pausing and resuming a torrent is enough.


feross:
Rather than setting `global.WEBTORRENT_ANNOUNCE`, you can just pass in the `announce` option to the `client.add` or `client.seed` methods. That will add the specified trackers to the list of used trackers.

See: https://github.com/webtorrent/webtorrent/blob/2b8f65ab59422a70fe23d529bb9380d71e6cc2d8/lib/torrent.js#L262-L265

subins2000:
In #881, I read

> we're currently adding four default webtorrent (ws:// and wss://) trackers to all torrents without making this very clear to users. There is an argument to be made that we should allow users to turn this off (just from a privacy perspective)`

So change needs to be made to the global variable too, right ? Should I also pass it to `announce` option in `.add` or just applying to the global variable would apply it to there too ?

I'm now trying to avoid loading state in `webtorrent.js` and instead use IPC to pass the global trackers list.

subins2000:
Yes, this is still relevant. Ability to change the default trackers used is important, from a privacy perspective and user experience. Users have submitted duplicate issues since the creation of this PR asking this (#881).

DiegoRBaquero:
@subins2000 is this good to go?

DiegoRBaquero (approved):

PR #1847, opened 2020-07-22
Update music metadata

Borewit:
Update [music-metadata](https://github.com/Borewit/music-metadata) to [version 7.0.0.](https://github.com/Borewit/music-metadata/releases/tag/v7.0.0)

Includes PR #1846 as well, which fixes metadata displayed for this and earlier versions.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5751,15 +5751,15 @@
       }
     },
     "music-metadata": {
-      "version": "6.3.6",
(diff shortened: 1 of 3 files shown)


hicom150:
I've tested this fix and works great! 👍

Borewit:
> But I don't know if #1846 does what is says, as I see just the opposite. All the dependencies are switched to `http` instead of `https` 🤔

Ah, yeah. 
 
4323893b73864a1f92ed06bd1cd48c6d

Fixed it.

PR #1850, opened 2020-07-25
Fix #1807 (VLC does not start)

### LINKED ISSUE #1807: VLC playing not working
**What version of WebTorrent Desktop are you using?**
Version 0.21.0-3

Manjaro 20.0

**What happened?**
Torrent doesn't play in VLC player.  I got a error 
```
TypeError [ERR_INVALID_ARG_VALUE]: The argument 'file' cannot be empty. Received ''
    at normalizeSpawnArguments (child_process.js:414:11)
    at Object.spawn (child_process.js:547:16)
    at spawnExternal (/usr/lib/webtorrent-desktop/build/main/external-player.js:55:13)
    at Object.spawn (/usr/lib/webtorrent-desktop/build/main/external-player.js:25:36)
    at IpcMainImpl.<anonymous> (/usr/lib/webtorrent-desktop/build/main/ipc.js:264:26)
    at IpcMainImpl.emit (events.js:210:5)
    at IpcMainImpl.ipc.emit (/usr/lib/webtorrent-desktop/build/main/ipc.js:304:18)
    at WebContents.<anonymous> (electron/js2c/browser_init.js:3851:15)
    at WebContents.emit (events.js:210:5) {
  code: 'ERR_INVALID_ARG_VALUE'
}

```
Torrent video must open in vlc.

Hinara:
https://github.com/webtorrent/webtorrent-desktop/blob/eb5a291c8b9cbfae96f306e153387e720a7baa72/src/renderer/lib/migrations.js#L219 cause external player to be a empty string instead of a null one

[ ] Documentation update
[X] Bug fix
[ ] New feature
[ ] Other, please explain:

VLC does not start

Fix #1807

### DIFF
--- a/src/main/external-player.js
+++ b/src/main/external-player.js
@@ -17,12 +17,12 @@ let proc = null
 function checkInstall (playerPath, cb) {
   // check for VLC if external player has not been specified by the user
   // otherwise assume the player is installed
-  if (playerPath == null) return vlcCommand(cb)
+  if (playerPath.length === 0) return vlcCommand(cb)
   process.nextTick(() => cb(null))
 }
 
 function spawn (playerPath, url, title) {
-  if (playerPath != null) return spawnExternal(playerPath, [url])
+  if (playerPath.length !== 0) return spawnExternal(playerPath, [url])
 
   // Try to find and use VLC if external player is not specified
   vlcCommand((err, vlcPath) => {
--- a/src/main/external-player.js
+++ b/src/main/external-player.js
@@ -17,12 +17,12 @@ let proc = null
 function checkInstall (playerPath, cb) {
   // check for VLC if external player has not been specified by the user
   // otherwise assume the player is installed
-  if (playerPath.length === 0) return vlcCommand(cb)
+  if (!playerPath) return vlcCommand(cb)
   process.nextTick(() => cb(null))
 }
 
 function spawn (playerPath, url, title) {
-  if (playerPath.length !== 0) return spawnExternal(playerPath, [url])
+  if (playerPath) return spawnExternal(playerPath, [url])
 
   // Try to find and use VLC if external player is not specified
   vlcCommand((err, vlcPath) => {


src/main/external-player.js
```diff
@@ -17,12 +17,12 @@ let proc = null
 function checkInstall (playerPath, cb) {
   // check for VLC if external player has not been specified by the user
   // otherwise assume the player is installed
-  if (playerPath == null) return vlcCommand(cb)
+  if (playerPath.length === 0) return vlcCommand(cb)
```
shahejz, line 20: Won't this throw a TypeError when `playerPath` _is_ `null` tho?

Hinara: That's why I included a reference to the migration file replacing null by an empty string.
However if it still possible to receive null I will modify my PR accordingly.

mathiasvr, line 20: Can we simplify like this to cover both `null` and `empty string` just in case:
```suggestion
  if (!playerPath) return vlcCommand(cb)
```

src/main/external-player.js
```diff
@@ -17,12 +17,12 @@ let proc = null
...
   process.nextTick(() => cb(null))
 }
 
 function spawn (playerPath, url, title) {
-  if (playerPath != null) return spawnExternal(playerPath, [url])
+  if (playerPath.length !== 0) return spawnExternal(playerPath, [url])
```
mathiasvr, line 25: Likewise here:
```suggestion
  if (playerPath) return spawnExternal(playerPath, [url])
```

Hinara:
I considered that null should be also accepted without throwing based on @mathiasvr review and change the PR content accordingly.

hicom150:
@Hinara as you correctly referenced, in the past we have made some changes #1704 #1702 so that all paths in the application should be default to `''`, so you shouldn't find any `null` in a path 😅

feross:
Thanks for the quick fix, folks! I'm going to put out a new WebTorrent Desktop release to get this into the hands of users.

PR #1878, opened 2020-10-18
Play/Pause video on click

agyanttiwari:
**Toggle Play/Pause on click on the player**

[x] Documentation update
[ ] Bug fix
[x] New feature
[ ] Other, please explain:

### DIFF
--- a/src/renderer/pages/player-page.js
+++ b/src/renderer/pages/player-page.js
@@ -129,6 +129,7 @@ function renderMedia (state) {
     <MediaTagName
       src={Playlist.getCurrentLocalURL(state)}
       onDoubleClick={dispatcher('toggleFullScreen')}
+      onClick={() => {state.playing.isPaused = !state.playing.isPaused}}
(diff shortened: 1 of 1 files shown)


src/renderer/pages/player-page.js
```diff
@@ -129,6 +129,7 @@ function renderMedia (state) {
     <MediaTagName
       src={Playlist.getCurrentLocalURL(state)}
       onDoubleClick={dispatcher('toggleFullScreen')}
+      onClick={() => { state.playing.isPaused = !state.playing.isPaused }}
```
feross, line 132: This should use `dispatch('playPause')`. We shouldn't modify the state from the render function, though I see that this rule is actually not being followed consistently in this file at the moment.

feross (changes requested):
LGTM

feross (approved):

alxhotel (approved):

PR #1880, opened 2020-10-23
Update BUG_REPORT.md

I-Cat:
+ up date bug report


feross:
Thanks for the PR, but I don't think this improves things.

PR #1886, opened 2020-11-04
Disable background throttling

feross:
Reference: https://github.com/webtorrent/webtorrent-desktop/pull/1720#issuecomment-722011487

I actually think we want backgroundThrottling: false so our timers and intervals aren't throttled when the app is backgrounded.

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

Disable background throttling. See https://www.electronjs.org/docs/api/browser-window

None

### DIFF
--- a/src/main/windows/about.js
+++ b/src/main/windows/about.js
@@ -27,7 +27,8 @@ function init () {
     webPreferences: {
       nodeIntegration: true,
       enableBlinkFeatures: 'AudioVideoTracks',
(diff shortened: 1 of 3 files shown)


hicom150 (approved):
LGTM 👍

alxhotel (approved):

feross:
Thanks for the reviews, @hicom150 and @alxhotel!

Issue #1894, opened 2020-11-12
support of Mac OS Big Sur?


saurabh02468:
Will the Web torrent app run on Big Sur?

justchen1369:
> **Mac build notes**
  The Mac app can only be packaged from macOS.

justchen1369:
https://webtorrent.io/desktop/

feross:
It should work on macOS Big Sur – however I'm updating right now to test it out.

feross:
It works perfectly for me, but I've heard a report from someone else that the app doesn't start up. We can probably fix this by updating Electron.

feross:
The person who said it doesn't work on macOS Big Sur ended up having issues with other apps too that seem to caused by a macOS bug. So not WebTorrent specific.

I think it's safe to say that WebTorrent Desktop works on macOS Big Sur without issues.

PR #1895, opened 2020-11-13
Upgrade to Electron 11

feross:

[ ] Documentation update
[ ] Bug fix
[ ] New feature
[x] Other, please explain: Update dep

Upgrade to Electron 11

Possibly addresses #1894

I'd like to start testing out Electron 11 on `master` so we can see if there's any deprecated APIs that we need to upgrade, etc. Then when Electron 11 is stable, we can do another release of the app :)

UPDATE: Electron 11 stable was just released.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2771,9 +2771,9 @@
       "dev": true
     },
     "electron": {
-      "version": "10.1.5",
-      "resolved": "https://registry.npmjs.org/electron/-/electron-10.1.5.tgz",
-      "integrity": "sha512-fys/KnEfJq05TtMij+lFvLuKkuVH030CHYx03iZrW5DNNLwjE6cW3pysJ420lB0FRSfPjTHBMu2eVCf5TG71zQ==",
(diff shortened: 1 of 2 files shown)


feross:
Electron 11 stable was just released. https://www.electronjs.org/releases/stable#release-notes-for-v1100

No breaking changes.

feross:
Self-merging without review -- sorry. I want to start testing Electron 11 so we're certain it's stable before the next release.

hicom150:
Sorry for the late review 😓

I've noticed that with `electron@11` this error is thrown 
```
electron.remote.screen.getAllDisplays is not a function (telemetry.js:77)
```

I've tested and this error does not appear with `electron@10`, so I suspect that could be something related with the warning that refers to https://github.com/electron/electron/issues/23506

I'm trying to find a fix, but I've not find a solution yet 😓

feross:
I opened an issue to track this error, @hicom150. See #1896

PR #1920, opened 2021-01-31
Improve libappindicator checking

### LINKED ISSUE #1399: Recover PR "only checking for the libappindicator1 package on debian systems"
We should take a look at this PR and decide if there's anything to do: https://github.com/webtorrent/webtorrent-desktop/pull/1216

(Context: The repository that submitted the PR was deleted, so I can't reopen it.)

### LINKED ISSUE #1851: No tray icon in non-Debian based distributions
https://github.com/webtorrent/webtorrent-desktop/blob/3a92f596ffab65a13386eb3cddf8e1e362b7aca9/src/main/tray.js#L57

This check should be rewritten to avoid using `dpkg` but instead look into `/usr/lib` or `/usr/lib64` for `libappindicator*` files.

I use Gentoo which doesn't use `dpkg`. I have [libappindicator](https://packages.gentoo.org/packages/dev-libs/libappindicator) package installed.

```
ls -al /usr/lib64/libappindicator*   
lrwxrwxrwx 1 root root    25  Oct  1  2019 /usr/lib/libappindicator3.so -> libappindicator3.so.1.0.0
lrwxrwxrwx 1 root root    25  Oct  1  2019 /usr/lib/libappindicator3.so.1 -> libappindicator3.so.1.0.0
-rwxr-xr-x 1 root root 56104  Oct  1  2019 /usr/lib/libappindicator3.so.1.0.0
```

Related: https://github.com/webtorrent/webtorrent-desktop/issues/383, https://github.com/webtorrent/webtorrent-desktop/issues/1215, https://github.com/webtorrent/webtorrent-desktop/pull/1216, https://github.com/webtorrent/webtorrent-desktop/issues/1399.

BlueManCZ:
Improved functionality

I rewrote check for `libappindicator` to work on all system, not only Debian based.

Closes #1399, closes #1851 

I tested this only on my Gentoo system. It should work the same way everywhere else but test it, please.

A simple test without uninstalling `libappindicator` is to change `ls /usr/lib*/libappindicator*` command to something nonexistent, like `ls /usr/lib*/nonexistent*`. In the second case, the tray icon should not show.


alxhotel (changes requested):
Hi @BlueManCZ,

Thanks for the PR!

I just tested this in my Ubuntu machine and it does not find the library. In addition, this needs permissions to access the `/usr` directory.

Please change the command to this one:

```sh
ldconfig -p | grep libappindicator
```

Check if the above command works in your Gentoo machine. If it works, then please update the PR with the change.

BlueManCZ:
Yes, this command works for me as well. Changed.

Just curious, where are those libraries located in Ubuntu?

And `/usr` directory is readable for everyone. I can't think of anyone not having access there.

alxhotel:
My libappindicator library in Ubuntu is located at `/usr/lib/x86_64-linux-gnu/libappindicator3.so.1`. That's probably why the other command doesn't work.

alxhotel (approved):

Issue #1315, opened 2018-01-29
TypeError: Cannot set property 'enabled' of undefined


feross:
**What version of WebTorrent Desktop?** (See the 'About WebTorrent' menu)

0.19.0

**What operating system and version?**

win32

**What did you do?**

```
Processes: main, platforms: win32, versions: 0.18.0 0.19.0
TypeError: Cannot set property 'enabled' of undefined
    at Object.setWindowFocus (...\build\main\menu.js:50:39)
    at BrowserWindow.onWindowFocus (...\build\main\windows\main.js:258:8)
    at emitOne (events.js:101:20)
    at BrowserWindow.emit (events.js:188:7)
```

This is the second most common error in the telemetry since the release of 0.19. cc @codealchemist - thoughts?

codealchemist:
I wasn't able to reproduce, but found that I was missing an initial state for `highestPlaybackPriority`, which is used to set a checkbox on Preferences page. It this doesn't fix the issue we should continue looking for a way to reproduce it and I'll work on a fix.

Thanks!

codealchemist:
Interesting. Let's keep an eye on this one.
If it continues to happen I'd like to spend more time on it.

Thanks!

feross:
This is our fifth-most common WebTorrent Desktop error. Adding it to the next sprint.

feross:
Should be fixed by #1938. We'll find out after the next WebTorrent Desktop release. If not, we can re-open this issue.

PR #1938, opened 2021-05-15
Fix cannot set property 'enabled' of undefined

codealchemist:

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

- Init menu before main window.
- Always return an object from `getMenuItem`.

#1315

I wasn't able to reproduce the bug and I'm currently working on OSX.
I might be able to get a Windows setup later to try to reproduce and verify the fix.
It doesn't seem possible for the main window events to trigger callbacks before `onReady()` finishes executing right?
In case I'm missing something about Electron inner workings `menu.init()` is called first now.
Then we ensure `getMenuItem()` always returns an object, which will fix the current error, but we need to see if it makes menu options unusable, being that `Full Screen, Float on Top`.
The problem is that for some reason the main window fires `onWindowBlur` or `onWindowFocus`, which calls `menu.setWindowFocus()` before the menu is built or the labels doesn't match.
Maybe on some Windows versions `menu.items[i].submenu.items` is constructed differently?
If someone who was able to consistently reproduce the issue can give more details about how to reproduce or can test this change that would be awesome :)

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -81,9 +81,9 @@ function init () {
     isReady = true
     const state = results.state
 
+    menu.init()
     windows.main.init(state, { hidden })
     windows.webtorrent.init()
-    menu.init()
 
     // To keep app startup fast, some code is delayed.
     setTimeout(() => {
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -74,7 +74,6 @@ function getMenuItem (label) {
     })
     if (menuItem) return menuItem
   }
-  
   return {}
 }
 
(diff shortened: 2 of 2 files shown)


alxhotel (approved):

feross:
> The problem is that for some reason the main window fires onWindowBlur or onWindowFocus, which calls menu.setWindowFocus() before the menu is built or the labels doesn't match.

I think you've identified it. Maybe Windows fires the event earlier? Let's try this PR and see if the issue improves.

Issue #1908, opened 2020-12-26
webtorrent download virus with vlc


2020karaulov2020:
After press download vlc, webtorrent download https://dnvnt0lrsqk55.cloudfront.net/vlc-3.0.5-streamer.exe file, but it blocked as Malware. Please remove this file from webtorrent !

alxhotel:
Hi @2020karaulov2020,

That link doesn't seem to be used in the Webtorrent Desktop app.

If you want to download VLC, please do it from the official source: https://www.videolan.org/vlc/

feross:
I noticed that we're linking to VLC over HTTP. We should use an HTTPS link in case this user was man-in-the-middled and redirected to a malicious website.

PR here: https://github.com/webtorrent/webtorrent-desktop/pull/1987

PR #1987, opened 2021-06-02
fix: Link to VLC using HTTPS

feross:
May help with users who are man-in-the-middled and redirected to malicious websites.

See: https://github.com/webtorrent/webtorrent-desktop/issues/1908


alxhotel (approved):
LGTM

feross:
Thanks for the review, @alxhotel!

PR #2143, opened 2022-02-22
fix(accessibility): Fix roles and add correct label for navigation controls

ctoth:
This makes screen readers properly indicate the control name, type, and state for the Back, Forward, and Add buttons.


src/renderer/components/header.js
```diff
@@ -11,20 +11,27 @@ class Header extends React.Component {
...
       >
         {this.getTitle()}
         <div className='nav left float-left'>
           <i
             className={'icon back ' + (loc.hasBack() ? '' : 'disabled')}
             title='Back'
             onClick={dispatcher('back')}
+            role='button'
+            aria-disabled={!loc.hasBack()}
+            aria-label  ='Back'
```
alxhotel, line 24:
```suggestion
            aria-label ='Back'
```

alxhotel (approved):

PR #2170, opened 2022-03-19
feat(Player Page): Add support for mediaSession

lavalleeale:

- [ ] Documentation update
- [ ] Bug fix
- [X] New feature
- [ ] Other, please explain:

Added support for the chrome `mediaSession` API so that hardware play/pause/next/back buttons will work to control the player
N/A

### DIFF
--- a/src/renderer/pages/player-page.js
+++ b/src/renderer/pages/player-page.js
@@ -1,8 +1,6 @@
 const React = require('react')
 const BitField = require('bitfield').default
 const prettyBytes = require('prettier-bytes')
-const path = require('path')
-const fs = require('fs')
 
 const TorrentSummary = require('../lib/torrent-summary')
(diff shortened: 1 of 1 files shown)


src/renderer/pages/player-page.js
```diff
@@ -1,6 +1,8 @@
 const React = require('react')
 const BitField = require('bitfield').default
 const prettyBytes = require('prettier-bytes')
+const path = require('path')
+const fs = require('fs')
```
DiegoRBaquero, line 5: why are these required?

lavalleeale: Sorry about that, I was reviewing the possibility of using the torrents image in the `mediaSession`, but removed it after I encountered issues

DiegoRBaquero (approved):

Issue #2462, opened 2025-03-25
Add Select All Files, Deselect All files Buttons to make the UX better 😎


kingomarnajjar:
**What version of WebTorrent Desktop are you using?**
Latest webtorrent-desktop@0.24.0'

**What operating system and version?**
OS Name	Microsoft Windows 11 Pro
Version	10.0.22631 Build 22631

**What problem do you want to solve?**
If user is downloading 500 individual files in 1 torrent for example a tv show, I want to be able to deselect ALL files so that I can Only easily Select the individual files I want to download at one time
**What do you think is the correct solution to this problem?**
I've already made it in a PR. Just right click and two buttons. 
"Select All Files"
"Deselect All Files"
**Are you willing to submit a pull request to implement this change?**
Yes. Already Done ✅

PR #2463, opened 2025-03-25
Added new feature to select and deselect torrent files

kingomarnajjar:

[ ] Documentation update
[ ] Bug fix
[X] New feature
[ ] Other, please explain:

Use case:
If downloading a tv show with 10 seasons and 500 files, I want to only download say 3 of those files, so that
 they can download quicker and I can start watching faster.
This allows customers to have a better UX experience by ADDING two buttons:
"Select All Files"
"Deselect All files"

Add Select All Files, Deselect All files Buttons to make the UX better 😎 #2462

its straight forward. Zero bugs. It's in  line with existing code so should be easy to add to master with only looking at the changed code for 10 seconds.

### DIFF
--- a/src/renderer/controllers/torrent-list-controller.js
+++ b/src/renderer/controllers/torrent-list-controller.js
@@ -253,6 +253,33 @@ module.exports = class TorrentListController {
     }
   }
 
+  selectAllFiles (infoHash) {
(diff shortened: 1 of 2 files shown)


kingomarnajjar:
New GUI changes
https://github.com/user-attachments/assets/ceabe29a-2378-4c51-8474-0ddc326f3a42

